PR #1, opened 2022-08-26
New rails application

starswan:
Create rails 7.0 application with Design System support with import maps and propshaft.

This is a simple attempt to start the EFE front-end rails application using rails new. It doesn't attempt to cover everything, so there may well be omissions.

Screenshot 2022-08-26 at 08 49 01
The main decisions are:

- Use rails 7.0 and propshaft as the asset pipeline
- setup rspec and capybara as testing tools
- integrate GOV.UK design system using formbuilder gem, and set it to the default.

This PR does not attempt to cover:

- Access to the CFE API using Faraday, and associated test tools (VCR?)
- Controller state handling (wizard gem?) aasm gem?
- Security checks via brakeman or snyk
- Content translation and editing using i18n.t and yml files
- Circle CI 
- automatic deployment to cloud platform
- ephemeral deploys of branches using Helm
- axe-core for accessibility testing
- propshaft asset pipeline in test/development mode (noticed that it's not obvious how it's meant it work)


config/initializers/content_security_policy.rb
```diff
@@ -0,0 +1,25 @@
+# Be sure to restart your server when you modify this file.
+
+# Define an application-wide content security policy.
+# See the Securing Rails Applications Guide for more information:
+# https://guides.rubyonrails.org/security.html#content-security-policy-header
+
+# Rails.application.configure do
```
willc-work, line 7: Should some/any/all of these be uncommented?

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,44 @@
+<!DOCTYPE html>
+<html lang="en" class="govuk-template">
+  <head>
+    <title><%= [yield(:page_title).presence, t('service.name')].compact.join(' - ') %></title>
+
+    <%= csrf_meta_tags %>
+    <%= csp_meta_tag %>
+
+    <%= tag :meta, name: 'viewport', content: 'width=device-width, initial-scale=1' %>
```
willc-work, line 9: there is no images directory containing these assets, so the page load fails.

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,44 @@
...
+    <%= tag :meta, property: 'og:image', content: asset_path('images/govuk-opengraph-image.png') %>
+    <%= tag :meta, name: 'theme-color', content: '#0b0c0c' %>
+    <%= favicon_link_tag asset_path('images/favicon.ico') %>
+    <%= favicon_link_tag asset_path('images/govuk-mask-icon.svg'), rel: 'mask-icon', type: 'image/svg', color: "#0b0c0c" %>
+    <%= favicon_link_tag asset_path('images/govuk-apple-touch-icon.png'), rel: 'apple-touch-icon', type: 'image/png' %>
+    <%= favicon_link_tag asset_path('images/govuk-apple-touch-icon-152x152.png'), rel: 'apple-touch-icon', type: 'image/png', size: '152x152' %>
+    <%= favicon_link_tag asset_path('images/govuk-apple-touch-icon-167x167.png'), rel: 'apple-touch-icon', type: 'image/png', size: '167x167' %>
+    <%= favicon_link_tag asset_path('images/govuk-apple-touch-icon-180x180.png'), rel: 'apple-touch-icon', type: 'image/png', size: '180x180' %>
+
+    <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
```
willc-work, line 19: No stylesheet or application.js either. Probably easier to just comment/delete these lines out for now.

config/initializers/content_security_policy.rb
starswan: don't think so - early days

app/views/layouts/application.html.erb
starswan: it doesn't? There's a test that shows the page load working

app/views/layouts/application.html.erb
starswan: This does work. I'm confused as to why you think it doesn't?

willc-work (approved):

Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+RUN apk add --update --no-cache tzdata && \
+    cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
+    echo "Europe/London" > /etc/timezone
+
+# libpq: required to run postgres
+RUN apk add --no-cache libpq
+
+# Copy files generated in the builder image
+COPY --from=builder /app /app
+COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
```
exonian, line 71: I've not worked with this pattern before. This is better than simply using layer caching, yeah?

exonian (approved):

Dockerfile
starswan: It just came for free in the Rails 7 template I used

PR #2, opened 2022-09-01
Try out usage of Wicked::Wizard gem

starswan:
The aim of this change is to demonstrate that our basic functionality is likely to be implementable using the patterns from Wicked::Wizard (which are essentially a flip between show and update actions, with a 'step' variable allowing different steps to do different things, and rendering to use the 'step' as the template name.

There is similar code in laa-apply (which doesn't use the gem) however their abstraction looks very similar so this seems like a reasonable approach.

There were no other contender libraries, so this decision could be reviewed in the future. The actual surface used is fairly small (as described above) - just the flow between show and update, and the concept of rendering a step. The 'save' functionality of the Wizard is not used here because it doesn't work well in conjunction with the usage of session storage, which belongs to the controller.

The session storage only uses attributes as this can cause issues with different session stores (if this is longer-term solution, we may well have to include Redis-backed sessions)


willc-work (approved):
Looks good to me, also tested in browser in an effort to UAT, all looks good there too.

PR #14, opened 2022-09-08
Change dependabot schedule to weekly

starswan:
Dependabot will always warn about security updates, but having it set to the default 'daily' generates lots of noise in the repository as packages are always having minor upgrades.


exonian:
Is weekly an option? That feels like a natural cadence to me

starswan:
> Is weekly an option? That feels like a natural cadence to me

Yes - done it. Previous project did monthly, but happy to try this

willc-work (approved):
Weekly is my preference too

PR #15, opened 2022-09-08
Create pull_request_template.md

willc-work:
This is a template for PRs created for the repo. It standardises the format and makes linking the JIRA ticket easier. Other than the link the rest is just a starter we may want to amend/add to these after we have discussed our ways of working.

it currently resides in a hidden .github folder in the repo but ti can be saved elsewhere if anyone prefers it to be more visible e.g in a `/docs` folder for example

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,14 @@
 [Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-XXX)
 
-What I have done on this ticket:
+Describe what you did and why.
 
+Checklist
 
-Checklist:
- * all tests pass
- * changes have full test coverage/repo has 100% coverage
- * recently rebased on main
+Before you ask people to review this PR:
+
+    Tests and rubocop should be passing
+    Github should not be reporting conflicts; you should have recently run git rebase main.
+    There should be no unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
+    The PR description should say what you changed and why, with a link to the JIRA story.
+    You should have looked at the diff against main and ensured that nothing unexpected is included in your changes.
+    You should have checked that the commit messages say why the change was made.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,9 @@
+[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-XXX)
+
+What I have done on this ticket:
+
+
+Checklist:
+ * all tests pass
+ * changes have full test coverage/repo has 100% coverage
+ * recently rebased on main
```
exonian, line 9: How recent? Do we expect one- or two-day tickets to do this, or do we just trust that if git thinks it can merge it it will be ok? For longer integrations, agree, but I'm not certain I like it in the general checklist

starswan (commented):
Might we just steal https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/docs/pull_request_template.md ? 

It says a bit more, but I like the coverage note (CI now checks it)

exonian (commented):
Fine by me as a starting point, happy to just get it going and then iterate on it readily.

One thing I'm not sure about, but not wedded to

.github/pull_request_template.md
willc-work: is there a reason not to rebase on main before merging? I have always felt it was good practice to do so.

starswan (approved):

.github/pull_request_template.md
exonian: I think I've run into some frustrations with it mucking with ongoing conversations on the PR (because they relate to commits, and rebasing and force-pushing replaces the commits) but I guess there's a difference between rebasing before creating your PR and rebasing after making changes based on feedback. I have no reason not to do the former, now I think of it. And I guess as this is the PR _creation_ template I've probably just got the wrong end of the stick.

Also I just wanted to push back slightly and challenge what we think the benefit is, and to clarify what we mean by "recently" so it can be enforced

exonian (approved):

.github/pull_request_template.md
willc-work: I am going to leave it in for now, perhaps at the next ways of working meeting we can discuss it further, see if we can quantify what 'recently' means so we can enforce it, otherwise we can remove the requirement.

PR #16, opened 2022-09-08
Investigate rails conventions around :js tag

starswan:
[EL-261](https://dsdmoj.atlassian.net/browse/EL-261)

Unknown to the author (and not documented anywhere I can find) there is an interesting 
distinction between system specs and feature specs.

System specs are expected to run using a browser (although strangely the rspec:system generator sets driven_by :rack_test) whereas feature specs are what the author is more used to, where rack_test is the default and is over-ridden on a case-by-case basis using the js: true annotation to the spec (or group). This moves the welcome_spec into the feature tests directory to generate discussion about what type the tests should be on this project.


exonian (approved):
From what I understand of this from the description, this solves an immediate problem as well as potentially being philosophically more correct, yes? I found it a little hard to identify exactly what the benefit was, but hopefully I've got there.

As long as either of those things is true, I'm absolutely fine with it.

PR #18, opened 2022-09-09
Add coverage tools to Circle:CI

starswan:
Add simplecov coverage tool with threshold set to 100% coverage. This will help us make sure that we don't forget to test any scenarios


spec/spec_helper.rb
```diff
@@ -11,7 +11,23 @@
 # a separate helper file that requires the additional dependencies and performs
 # the additional setup, and require it from the spec files that actually need
 # it.
-#
+
+if ENV.key? "COVERAGE"
```
exonian, line 15: Does this just check if the key is present? Or is it also checking the truthiness of the value? (this is my lack of familiarity with Ruby showing through...)

I ask because it looks like the former and I'd be more comfortable with the latter.

.circleci/config.yml
```diff
@@ -12,6 +12,7 @@ executors:
         environment:
           PGHOST: localhost
           PGUSER: ubuntu
+          COVERAGE: 1
```
exonian, line 15: If this is just for boolean true, is there a more explicit literal we can use to be more, well, explicit about intent here? I know 0/1 as booleans _is_ well established convention, but where true and false literals are available I think it's clearer ie I'd never wonder if I could turn true into doubletrue ;)

starswan: not really no. The 2 items you have commented on kind of go together - checking the presence means that the value here doesn't need to be truthy...

exonian: Yeah it would, which is not a good thing in my opinion. You need more familiarity than I think you should in order to know that changing this to `"false"` or `false` or `0` would do nothing at all, which is my issue with the other bit.

And my issue with this bit is that having it as a 1 looks like it's numerical information, which is again needlessly opaque

starswan: There are several ways of dealing with this, not sure which way is best? I've used several patterns for this in the past, none of them completely satisfactory. 

- Treat as an integer, with > 0 being enabled (so 0 turns it off) and calling to_i on the string (so nil becomes 0)
- This method with presence implying on
- check for literal string 'true'

WDYT? Is there a 4th option you think is better?

starswan: We don't have the YAML translation layer here sadly - because yes you're right YAML false is false, but environment variable 'false' would come out true. So I'm not sure whether there is a 'best' option. Ruby doesn't have a 'to_boolean' method that I know of. 

There's ActiveRecord::Type::Boolean.new.cast(x) which converts lots of values into 'false', but we don't have rails loaded (present? doesn't work here) so it probably doesn't work either

exonian: Ahhhh of course, yes that's what I was missing: that we were going from yaml to ruby, via env vars! 🤦 This is a good compromise solution!

exonian (approved):

PR #19, opened 2022-09-09
Remove RSpec github action

starswan:
RSpec tests are now being run from Circle CI with mandatory pass, so this action is no longer required


willc-work (approved):
LGTM

PR #20, opened 2022-09-09
Deploy with helm and circle

willc-work:


Gemfile
```diff
@@ -68,3 +68,8 @@ end
 gem "govuk-components"
 gem "govuk_design_system_formbuilder"
 gem "wicked"
+
+group :test do
+  gem "rspec"
+  gem "rspec-rails"
+end
```
starswan, line 75: Think this needs reverting

helm_deploy/laa-estimate-eligibility/templates/deployment.yaml
```diff
@@ -0,0 +1,42 @@
...
+          livenessProbe:
+            httpGet:
+              path: /estimates/1/build_estimates/new
+              port: http
+            initialDelaySeconds: 30
+          readinessProbe:
+            httpGet:
+              path: /estimates/1/build_estimates/new
+              port: http
+            initialDelaySeconds: 30
```
starswan, line 40: This is going to break with the next PR merge. Probably better to hit the root path and make sure that the root is mapped ?

willc-work: I think we need another ticket for this. Essentially I feel this ticket is big enough as it currently is. Looking at CFE/Apply they have implemented a ping page for the K8s readiness/liveness probe. Its probably worth discussing what approach we want to take for EFE. Unless the next PR merge removes that route it should work fine until we decide on a more permanent solution.

exonian: Can we throw the most basicest of routes in to get a bit of a best of both? I got the impression from @starswan's comment that that's exactly what the next PR will do.
If we went with /ping/ just returning the most absolute bare mininum 200 response for now that'd not be too much work I think and would mean other bits of work don't have to worry aobut compaitbility for now (which they shouldn't have to), without committing us to a solution for what the real probe endpoints will look like

exonian: But yes agree we should have a ticket for deciding on and creating the live and ready endpoints

willc-work: basic status page added and readiness/liveness endpoint updated

Dockerfile
```diff
@@ -3,14 +3,17 @@
...
 
 # RUN apk -U upgrade && \
 #     apk add --update --no-cache gcc git libc6-compat libc-dev make nodejs \
 #     postgresql13-dev yarn
 
 WORKDIR /app
 
+RUN addgroup -g 1000 -S appgroup \
+  && adduser -u 1000 -S appuser -G appgroup
+
```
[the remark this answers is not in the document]
willc-work: these user group changes are required due to the runAsNonRoot requirement. However I have removed them from the production image section as it was just duplicating the same task.

starswan (approved):

PR #22, opened 2022-09-12
Spike for slim templates

starswan:
This is an example of how the trivial home page looks like re-written using Slim. This was driven partly by the 'erb-lint' story which this format makes pretty redundant

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,13 +74,9 @@ jobs:
       - attach_workspace:
           at: ./
       - run:
-          name: Rubocop checks
+          name: slim-lint checks
           command: |
-            bundle exec rubocop
-      - run:
-          name: erb-lint checks
-          command: |
-            bundle exec erblint --lint-all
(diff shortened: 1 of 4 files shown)


exonian:
Is there a ticket for this? I know I saw some discussion about templating language and would like to look at this in the context of that

starswan:
> Is there a ticket for this? I know I saw some discussion about templating language and would like to look at this in the context of that

It came out of the erb-lint story (EL-264) where I noticed that if we stopped writing ERB code then erb-lint disappears. The only discussion has been https://mojdt.slack.com/archives/C014YJ90G48/p1663668474488949?thread_ts=1662022208.568359&cid=C014YJ90G48

exonian:
OK. I don't think I understand why erb-lint disappearing is particularly valuable: did we have a particular issue with it?

As for this example in slim, I know I'm unfamiliar but I find it almost incredibly hard to read 😅

PR #23, opened 2022-09-12
Rubocop-govuk and ERB lint checks

starswan:
Add rubocop-govuk and erb-lint tools to Circle:CI toolchain to assist with code consistency and avoiding known defect patterns


willc-work (commented):
I prefer to use https://www.rubydoc.info/gems/rubocop/0.11.1/Rubocop/Cop/Style/StringLiterals in place of the govuk suggestion that all strings should be in double quotes. This way the intent and ability to interpolate is clear. 

Not a hill I want to die on but wanted to raise it for discussion at least.

willc-work (approved):
LGTM

PR #24, opened 2022-09-12
Run bundler-audit to check for (new) security issues in codebase

starswan:
This replaces the (opaque) dependency-check github action with a more definitive (and transparent) run of bundle-audit


willc-work (dismissed):
LGTM but there is a snyk license violation that probably should be resolved before merging

willc-work (changes requested):
At present we cannot merge this as the gem bundler-audit is released under the GNU 'General Public License v3.0 or later' and we release our work as standard under the MIT license. according to this: https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l
we would need to change from MIT to a GPL v3+ license for the entire project.

exonian:
From @starswan on Slack:
> Think I might just have to close [this] as it doesn't give us any benefit and snyk doesn't seem to understand that bundler-audit is not part of the product. WDYT @exonian?

My response:
This seems to be a limitation of gems, that even if you group it for development only it still installs everywhere. I wonder if there's a Snyk setting for that, but I could well imagine not because this lib does get installed everywhere, so even if it doesn't run it's kinda hard to argue it's not part of our software.
One to keep in mind for if we ever run into this type of problem again when we really do need the thing, but yes if we can do the same thing without this let's just do that for now.

PR #26, opened 2022-09-13
Add Brakeman checks

starswan:
This adds brakeman via a GitHub action for ease of implementation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,6 @@ references:
         echo "${GIT_CRYPT_KEY}" | base64 -d > git-crypt.key
         git-crypt unlock git-crypt.key
 
-
 jobs:
   install_dependencies:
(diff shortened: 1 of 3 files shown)


willc-work (dismissed):

willc-work (commented):
At present we cannot merge this as the gem bundler-audit is released under the GNU 'General Public License v3.0 or later' and we release our work as standard under the MIT license. according to this: https://opensource.stackexchange.com/questions/1640/if-im-using-a-gpl-3-library-in-my-project-can-i-license-my-project-under-mit-l
we would need to change from MIT to a GPL v3+ license for the entire project.
Is the gem included here just because it was based on the branch that added that gem, or is the gem required for this branch? (I dont think it is but I may be wrong)

willc-work (approved):

PR #28, opened 2022-09-14
Remove database dependency

starswan:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-282)

For now it looks quite likely that the calculator application will not have a local database attached, so remove all references to ActiveRecord to simplify application setup.


willc-work (approved):
agree removing this makes sense for now, we can add a db later if the service needs one.

PR #30, opened 2022-09-16
EL-321 Ephemeral deployments to uat

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-321)

Update circleci config to deploy to UAT for new branches and deploy to staging on merge to main
Add a UAT deploy script that creates a kubernetes deployment using the git branch name
Update values files to remove hardcoded values for ingress and deploy host

Checklist


.circleci/config.yml
```diff
@@ -175,6 +194,37 @@ workflows:
...
+  merge_pr_to_main:
+    jobs:
+      - install_dependencies:
+          filters:
+            branches:
+              only: main
+      - linters:
+          filters:
+            branches:
+              only: main
```
exonian, line 211: Is this redundant given the `requires` having that branch condition too?

.circleci/config.yml
```diff
@@ -175,6 +194,37 @@ workflows:
...
+          requires:
+            - install_dependencies
+      - run_specs:
+          filters:
+            branches:
+              only: main
```
exonian, line 217: As in `linters` job. Keep conversation there, but if that conversation results in a change it will likely be needed here too.

.circleci/config.yml
```diff
@@ -162,7 +181,7 @@ jobs:
             
 
 workflows:
-  open_pr:
+  open_new_pr:
```
exonian, line 184: Is this name special, ie used by either github or circleci to run it on the PR creation event?
Or is it just a name, and it's run on all commits or something?

.circleci/config.yml
```diff
@@ -162,7 +181,7 @@ jobs:
             
 
 workflows:
```
exonian, line 183: There's a lot of repetition between these workflows. I wonder whether that can be optimised easily

.circleci/config.yml
willc-work: Possibly. I will test it, I was going with belt and braces approach.

.circleci/config.yml
willc-work: we could possibly pull the jobs such as `linters`, `install_dependencies`, `run_specs` into a job of their own and call them all at once, however I think this would not really be an optimisation rather just DRYing the code. Which I dont think adds value here as when we deploy to main we want to be sure that all the individual sections run and pass before we deploy. It also allows for monitoring the progress on CircleCI as here: https://app.circleci.com/pipelines/github/ministryofjustice/laa-estimate-financial-eligibility-for-legal-aid/418/workflows/f7f12e13-d136-4ded-a877-128e098483e4
I think we could (if you wanted to) remove some of these checks from a deployment to main, after all the PR would have to pass all these before it could be merged. it adds duplication but also a level of assurance. Similarly we will repeat these steps for a deployment to staging and again for production, so we may want a wider discussion on what exactly our pipeline should look like.

.circleci/config.yml
willc-work: Appears I can remove all filters after the first one.

.circleci/config.yml
exonian: To clarify, I like the lots of small jobs. What I'm unsure about is having two workflows that mostly call all the same jobs, and whether we could have one workflow for both but with conditional steps (eg by branch filters etc) to handle where they differ.

It was the DRYing that I thought would be valuable. Just feels very vulnerable to going out of sync with each other, if something gets changed in one place but not the other.

.circleci/config.yml
exonian: OK so workflows run only on PR, not on all commits, so I'm wondering what will cause the `merge_pr_to_main` workflow to run. Do we also have a webhook to run workflows on PR merges not just commits to open PRs? If so, is there anything to stop _this_ workflow from also running when we merge a PR? I can see that the other one is only on main, but this one doesn't exclude main

.circleci/config.yml
```diff
@@ -137,6 +137,25 @@ jobs:
...
+      - *authenticate_k8s
+      - deploy:
+          name: Deploy LAA Estimate Eligibility Service to UAT
+          command: |
+            ./bin/uat_deployment 
+
+# this job does the same as deploy_uat however it will be
+# deploying to the main uat branch as it is only called
+# in the merge_pr_to_main workflow
+  deploy_main_uat:
```
exonian, line 156: Also I don't understand why this is separate. Is it going to get its own steps? Its own parameters (not passed from the workflow)?

willc-work: the reason for this is really a circleci display thing. On the workflow pages e.g https://app.circleci.com/pipelines/github/ministryofjustice/laa-estimate-financial-eligibility-for-legal-aid/500/workflows/e72cf1a3-5e56-4630-a966-6777bcec3af3
the name given here is displayed in one of the boxes. This helps to make sense of the pipeline when viewing the visual workflow.

.circleci/config.yml
exonian: OK we'll see how it goes. I'm struggling to follow that logic entirely, but if you're confident I'm confident (and the negative impact of being wrong is minimal)

.circleci/config.yml
exonian: Super explanation. In that case, I think [reusing the job directly but passing a `name` parameter](https://github.com/ministryofjustice/cla_backend/blob/master/.circleci/config.yml#L331-L332) may be best of both. What do you reckon?

exonian (approved):
Good work

PR #33, opened 2022-09-20
Call CFE API via Faraday

starswan:
This adds the start of calling the CFE v5 API via Faraday. It includes the tooling for tests around VCR and webmock, so that replies have actually been received from the API. The reduces the risk of a 'mocking error' where the dependency is mocked incorrectly and then the code fails when attached to the actual dependency

(https://dsdmoj.atlassian.net/browse/EL-253)


app/lib/cfe_connection.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  class << self
+    def connection
+      CfeConnection.new
+    end
+  end
+
+  def create_assessment_id
+    create_request = {
+      client_reference_id: "EFE-REFERENCE-ID",
```
willc-work, line 12: should this ID increment with each call? or is this more like a placeholder until we decide on what we send to CFE?

starswan: we don't have a client reference - and it usage is unclear (e.g. the API doesn't seem to allow retrieval of old applications by client reference) and I think its mandatory

exonian (commented):
I'm afraid I don't have much pertinent on this. Nothing beyond the one query stood out but definitely put more stock by Will's review on this as I suspect quite a lot could get past me on Rails stuff still 😅

app/controllers/build_estimates_controller.rb
```diff
@@ -1,61 +1,70 @@
...
-    when :monthly_income
-      @estimate = MonthlyIncomeForm.new session[session_key].slice(*MONTHLY_INCOME_ATTRIBUTES)
-    when :property
-      @intro_form = IntroForm.new session[session_key]
-    end
+    handler = HANDLER_CLASSES[step]
+    @form = if handler
+              handler.model(session_data)
+            else
+              cfe_connection.api_result(estimate_id)
```
patrick-laa, line 23: It might be that we want to display a page with CFE's calculation results more than once (e.g. if the user views the results page and then the print-friendly result page). Given that, rather than call CFE at the point where we first want to display the results, would it make more sense to call it when we submit the penultimate page, and stash the result in the session? Then we could add a Flow::CalculationResultHandler and do away with this if/else statement entirely.

If that's beyond the scope of this ticket then ignore me entirely!

app/helpers/steps_helper.rb
```diff
@@ -0,0 +1,42 @@
+module StepsHelper
+  TAIL_STEPS = %i[vehicle assets summary results].freeze
+  ONLY_PROPERTY_STEPS = %i[property property_entry].freeze
+
+  STEPS_NO_PROPERTY = (%i[applicant monthly_income outgoings property] + TAIL_STEPS).freeze
+  STEPS_WITH_PROPERTY = (%i[applicant monthly_income outgoings] + ONLY_PROPERTY_STEPS + TAIL_STEPS).freeze
+  PASSPORTED_STEPS = (%i[applicant property] + TAIL_STEPS).freeze
+  PASSPORTED_STEPS_WITH_PROPERTY = ([:applicant] + ONLY_PROPERTY_STEPS + TAIL_STEPS).freeze
+
+  # codify steps into a readable rule-set table rather than in code
```
patrick-laa, line 10: :+1: to this approach!

config/initializers/govuk_formbuilder.rb
```diff
@@ -8,7 +8,7 @@
   # config.default_legend_size: 'm'
   # config.default_legend_tag: nil
   # config.default_caption_size: 'm'
-  # config.default_submit_button_text: 'Continue'
+  config.default_submit_button_text = "Save and continue"
```
patrick-laa, line 11: Should this be an `I18n.t`?

spec/features/applicant_page_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+      allow(CfeConnection).to receive(:connection).and_return(mock_connection)
+      visit "/estimates/new"
+
+      %i[over_60 dependants partner passporting employed].reject { |f| f == field }.each do |f|
+        select_applicant_boolean(f, true)
+      end
+      click_on "Save and continue"
+    end
+
+    context "with over_60" do
```
patrick-laa, line 21: I'm a little slow in grokking specs, and I didn't understand at first why the expectation was that there would be an error in these specs. To make it easier for me to tell what's going on, could we make the context string something like `"when the 'over 60' question is omitted"`?

spec/features/assets_page_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+
+RSpec.describe "Assets Page", :vcr do
+  let(:arbitrary_fixed_time) { Time.zone.local(2022, 9, 5, 9, 0, 0) }
+  let(:assets_header) { "Which assets does your client have?" }
+
+  before do
+    travel_to arbitrary_fixed_time
```
patrick-laa, line 8: I think `arbitrary_fixed_time` is one of the best names I've encountered in a while. It captures the what and the why really nicely!

app/lib/cfe_connection.rb
```diff
@@ -0,0 +1,60 @@
+class CfeConnection
+  CFE_HOST = "https://check-financial-eligibility-staging.cloud-platform.service.justice.gov.uk/".freeze
```
patrick-laa, line 2: When we eventually want to point our production at production CFE is there a smart system for service discoverability or is it just a case of different URLs in env vars?

patrick-laa (commented):
I hope I'm not treading on any toes here, but for the sake of spending a little bit of time  today _not_ reading documentation, I read this PR. I really like the overall approach, and just have a couple of comments and questions.

app/controllers/build_estimates_controller.rb
```diff
@@ -1,61 +1,77 @@
...
-      if @estimate.valid?
-        session[session_key].merge(@estimate.attributes)
+    handler = HANDLER_CLASSES[step]
+    @form = handler.form(params)
+
+    if @form.valid?
+      if step == :summary
+        estimate = load_estimate
+        cfe_connection.create_applicant estimate_id,
+                                        date_of_birth: estimate.over_60 ? 61.years.ago.to_date : 59.years.ago.to_date,
```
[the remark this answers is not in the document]
starswan: I changed it to 50 and 70 to remove that distinction. However if this ever happens then we might have a bigger problem if the applicant in 59 years 11 months old...

app/controllers/build_estimates_controller.rb
```diff
@@ -1,61 +1,70 @@
...
-      if @estimate.valid?
-        session[session_key].merge(@estimate.attributes)
+    handler = HANDLER_CLASSES.fetch(step)
+    @form = handler.form(params)
+
+    if @form.valid?
+      if step == :summary
+        estimate = load_estimate
+        cfe_connection.create_applicant estimate_id,
+                                        date_of_birth: estimate.over_60 ? 61.years.ago.to_date : 59.years.ago.to_date,
```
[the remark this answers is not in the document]
starswan: thanks, I have been trying to remove all this logic out of the controller so that it just looks after flow (we will probably have sub-flows soon or something) so I've managed to move this into SummaryHandler#save_data where it seems to belong

config/initializers/govuk_formbuilder.rb
starswan: I was going to say no - then the word 'Welsh Translation' appeared...

app/controllers/build_estimates_controller.rb
starswan: That would solve the problem. I'm not sure we want to cache the API result in the session though - there might be too many places to forget to update it. I wanted to push everything to the API, but it doesn't allow that (specifically w/o an applicant the 'GET' fails, and applicants can't be overwritten). Maybe we should consider 'fixing' that in our dev meeting - it is weird that we collect the applicant data first but submit it last

patrick-laa (approved):
LGTM

PR #34, opened 2022-09-20
More CFE API calls

starswan:
(https://dsdmoj.atlassian.net/browse/EL-253)

This is the second half of the CFE API work. It includes most of the (existing) API calls that the application will need to make in order to create an estimate, although a lot of the data capture screens still need to be filled in.


patrick-laa (approved):
LGTM

patrick-laa:
Changes since review LGTM 👍

PR #38, opened 2022-09-23
Restrict ECR and deploy tasks to main branch

starswan:
Currently we are deploying all builds to ECR and staging - this is slow and is resulting in resource clashes for the staging environment


willc-work (approved):
This is covered in my PR for UAT deployments. I dont mind merging this as a temp fix, however please note that the build and push to ecr will need to happen for every branch for UAT

PR #45, opened 2022-09-26
Remove force-ssl config as we are behind an SSL reverse proxy

starswan:
One of the security checks added the force_ssl flag to our production config, but this is incorrect as we deploy behind an SSL-enabled reverse proxy


exonian (approved):
Good. Nice PR description. If I weren't familiar already, this would have given me everything I needed to feel confident in what could otherwise look like a scary / ill-advised PR :)

PR #47, opened 2022-09-26
Bug: assets form can't ask second home question if no first home owned

starswan:
https://dsdmoj.atlassian.net/browse/EL-253

After merging the second part of EL-253, I remembered that the API prevents submission of a second home if there is no first home - so we must not ask the question in that scenario.


patrick-laa (approved):
LGTM 👍

PR #48, opened 2022-09-26
Add test for results page defect

starswan:
The original source of the defect on number formatting was the lack of a test.

This adds the test to the PR which fixes the defect


patrick-laa (approved):
Thank you!

PR #49, opened 2022-09-27
Bugfix: Remove duplication from CircleCI workflow and fix staging deploy issue

willc-work:
Original ticket: https://dsdmoj.atlassian.net/browse/EL-321

bugfix - the pipeline workflow was being duplicated after a merge to main. This adds filters for the open_new_pr job so that it doesn't run the workflow on main.

Removed --debug statements from the deploy commands

Updated the ingress naming for staging to match the required naming convention

Checklist


.circleci/config.yml
```diff
@@ -180,19 +179,34 @@ jobs:
...
     jobs:
-      - install_dependencies
+      - install_dependencies:
+          filters:
+            branches:
+              ignore: main
       - linters:
+          filters:
+            branches:
+              ignore: main
```
exonian, line 189: Can we use the same logic we used on the PR one to only specify the branch filter once and then rely on the `requires` for it to block everything else?

exonian (commented):
I think it would be good to link this to the same ticket as the PR we just merged, and also I think a slightly more descriptive title might be helpful

.circleci/config.yml
exonian: I'm pretty confident it'll work. To be brutally honest, I'm more confident it'll work than I was that the previous one would work 😉 and we went ahead with that for science...

exonian (approved):
Appreciate it. If it goes wrong, there's a paper trail right back to me 😁 :shipit:

PR #50, opened 2022-09-27
EL-334 add service start page in slim.html

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-334)

Added a home page as root so we have a starting point for the service.
Removed pages#home as we dont need that in the service
update the footer with some placeholder links - there are further tickets for content for these e.g. accessibility_statement/privacy policy
CTA "Start Now" button links to the status page for now

Add slim gem and convert existing html.erb files to slim

Add slim-lint gem for linting locally

What this PR does not do: Add slim-lint checks on CircleCI or remove the erb-lint checks as we still have a lot of erb pages.

Checklist


app/views/layouts/_footer.html.slim
```diff
@@ -0,0 +1,36 @@
+footer class="govuk-footer no-print" role="contentinfo"
```
patrick-laa, line 1: The shorthand for classes in slim, which I think is more idiomatic, is to use `.` syntax. So this would be: `footer.govuk-footer.no-print role="contentinfo"`

app/views/layouts/_footer.html.slim
```diff
@@ -0,0 +1,36 @@
...
+  div class="govuk-width-container"
```
patrick-laa, line 2: Slim also assumes all elements are divs unless otherwise specified, so you can simply put `.govuk-width-container` for this line

app/views/layouts/application.html.slim
```diff
@@ -0,0 +1,32 @@
...
+
+  body class="govuk-template__body"
+    javascript: document.body.className = ((document.body.className) ? document.body.className + " js-enabled" : "js-enabled");
+    
+    = govuk_skip_link 
+    
+    = govuk_header(service_name: t("service.name")) 
+
+    div class="govuk-width-container"
+      main class="govuk-main-wrapper" id="main-content" role="main"
```
patrick-laa, line 29: Also the slim id shorthand which, again, I think is more idiomatic, is `#`, so in this case `main#main-content.govuk-main-wrapper role="main"`

patrick-laa (commented):
A few slim comments based on what I believe to be the most "normal" way of using slim. But if we deliberately want to avoid shorthand syntax that's OK too.

app/views/layouts/_footer.html.slim
willc-work: thanks, this is due to my unfamiliarity so all pointers are welcome

app/views/start/index.html.slim
```diff
@@ -0,0 +1,31 @@
...
+    p class="govuk-body"= t(".required_information")
+
+    .govuk-list
+      ul class="govuk-list govuk-list--bullet"
+        li = t(".item_1")
+        li = t(".item_2")
+        li = t(".item_3")
+
+      = link_to start_button_label(:start_button),
+        "/status",
```
patrick-laa, line 20: One other thing: it would be really cool if this linked to `new_estimate_path`, as that is now able to call CFE, generate an assessment ID and move the user into the flow.

app/views/layouts/application.html.slim
```diff
@@ -0,0 +1,32 @@
...
+    = favicon_link_tag asset_path("images/govuk-apple-touch-icon.png"), rel: "apple-touch-icon", type: "image/png" 
+    = favicon_link_tag asset_path("images/govuk-apple-touch-icon-152x152.png"), rel: "apple-touch-icon", type: "image/png", size: "152x152" 
+    = favicon_link_tag asset_path("images/govuk-apple-touch-icon-167x167.png"), rel: "apple-touch-icon", type: "image/png", size: "167x167" 
+    = favicon_link_tag asset_path("images/govuk-apple-touch-icon-180x180.png"), rel: "apple-touch-icon", type: "image/png", size: "180x180" 
+
+    = stylesheet_link_tag "application", "data-turbo-track": "reload" 
+    = javascript_include_tag "application", "data-turbo-track": "reload", defer: true
+
+  body class="govuk-template__body"
+    javascript: document.body.className = ((document.body.className) ? document.body.className + " js-enabled" : "js-enabled");
```
patrick-laa, line 22: One other other thing, sorry: Slim doesn't parse this correctly, so JS doesn't run correctly. To make it work you need a line break after `javascript:` and indent the next line 2 spaces further in than the `javascript:` indentation

willc-work: thanks, this was one I was concerned about

patrick-laa (approved):

PR #51, opened 2022-09-28
EL-358: Update income page

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-358)

This ticket:
* Converts the monthly income page to Slim
* Changes the list of categories of non-employment, non-benefits income recordable to match the [designs](https://miro.com/app/board/uXjVPXRjhas=/?moveToWidget=3458764534164899406&cot=14)
* Sends the additional information to CFE*
* Adds appropriate validation error wording
* Fleshes out the feature spec to confirm the right information is passed to the CfeConnection.

(* Note that as I've raised in Slack and Confluence, CFE currently ignores the 'other' income value.)

Checklist


starswan (approved):

app/views/layouts/application.html.slim
```diff
@@ -27,6 +27,10 @@ html lang="en" class="govuk-template"
     = govuk_header(service_name: t("service.name")) 
 
     div class="govuk-width-container"
+
+      - if content_for?(:back)
+        = yield :back
+
```
starswan, line 33: Not a big deal, but it appears that we don't need the check: https://apidock.com/rails/ActionView/Helpers/CaptureHelper/content_for%3F

starswan (approved):
I've added one comment, but it's **_really_** tiny

patrick-laa:
This screen has the following known limitations:

* Any 'other income' entered will be ignored for the purposes of eligibility calculations
* If too many applications are created then eventually (and quite possibly on this screen), the app will fail with a cookie overflow error
* Going back to this screen and re-submitting it, with or without changes will cause inaccuracies in the eligibility calculation
* Going back to this screen and re-submitting it will cause an error if a 'student finance' figure was present both times it was submitted

None of these issues can be fixed at the moment, as they all depend on unresolved architectural decisions. So I've documented them in comments the code (as well as Jira, Slack etc) to try to ensure we don't forget about them.

starswan (approved):

PR #52, opened 2022-09-29
EL-361 Complete Vehicle Screens

starswan:
This implements the extra vehicle question logic, and the subsequent submission of the vehicle data to CFE

(https://dsdmoj.atlassian.net/browse/EL-361)


app/controllers/flow/vehicle/owned_handler.rb
```diff
@@ -0,0 +1,18 @@
+module Flow
+  module Vehicle
+    class OwnedHandler
+      class << self
+        def model(session_data)
```
patrick-laa, line 5: Is there a benefit in DRYing up all these `.model` and `.form` methods? I'm asking as a genuine question.

app/helpers/steps_helper.rb
```diff
@@ -19,21 +25,45 @@ module StepsHelper
...
+    vehicle_owned: FIRST_VEHICLE_STEPS,
+    vehicle_regular: ALL_VEHICLE_STEPS,
+  }.freeze
+
   def next_step_for(intro, step)
-    next_estimate_step(steps_list_for(intro), step)
+    next_estimate_step(steps_list_for(intro).flatten, step)
+  end
+
+  def last_step?(model, step)
```
patrick-laa, line 38: I don't like this method name, because it made me assume this was referring to the last step overall, not the last step in the current subgroup. `last_step_in_group?` ? `last_step_in_current_list?`

config/locales/en.yml
```diff
@@ -170,6 +186,29 @@ en:
       percentage_owned:
         input: What percentage share of their home does your client legally own?
         input_hint: Your client's name must be on the property deeds, lease or freehold.
+    vehicle_value:
+      caption: Yout clients finances
+      heading: Add details about your clients vehicle
```
patrick-laa, line 191: `client's` - apostrophe missing

config/locales/en.yml
```diff
@@ -170,6 +186,29 @@ en:
       percentage_owned:
         input: What percentage share of their home does your client legally own?
         input_hint: Your client's name must be on the property deeds, lease or freehold.
+    vehicle_value:
+      caption: Yout clients finances
```
patrick-laa, line 190: `Your` - typo
`client's` - apostrophe missing

config/locales/en.yml
```diff
@@ -170,6 +186,29 @@ en:
...
+      heading: Add details about your clients vehicle
+      vehicle_value:
+        input: What is the estimated value of the vehicle?
+        input_hint: You can use car valuation websites to find this out
+      vehicle_in_regular_use:
+        legend: Is the vehicle in regular use?
+        hint: This means the vehicle is used at least once every 2 weeks for things like work, shopping or driving children to school.
+    vehicle_age:
+      caption: Your clients finances
```
patrick-laa, line 199: Apostrophe

config/locales/en.yml
```diff
@@ -170,6 +186,29 @@ en:
...
+      heading: ''
+      vehicle_three_years_old:
+        legend: Did your client buy the vehicle over 3 years ago?
+        hint: ""
+    vehicle_finance:
+      caption: Your clients finances
```
patrick-laa, line 205: Apostrophe

config/locales/en.yml
```diff
@@ -170,6 +186,29 @@ en:
...
+      vehicle_pcp:
+        legend: Are there any payments left on the vehicle?
+        hint: Include any outstanding payments on loans, hire purchase(HP) or personal contract purchase (PCP) agreements
```
patrick-laa, line 208: `hire purchase (HP)` - space missing

app/views/build_estimates/vehicle_value.html.slim
```diff
@@ -0,0 +1,26 @@
...
+    h1.govuk-heading-l
+      = t(".heading")
+
+    = form_for(@form,
+                 url: wizard_path,
+                 method: :put) do |form|
+
+      = form.govuk_error_summary t("generic.error_summary_title")
+
+      = render "money_input",
```
patrick-laa, line 15: You're going to hate this one... I'm getting flaky behaviour on this input. _every so often_ when I update the value in the field and click 'Save and continue' then click back, the updated value is lost and the previous value is retained. See screen capture:
https://user-images.githubusercontent.com/113888736/193307375-5b986188-0693-4996-948f-c9b46c4c6b0a.mov

patrick-laa (commented):
Mostly small things, but I'm slightly worried about the screencap I shared, cos it really ought to just reliably work.

app/controllers/flow/vehicle/owned_handler.rb
starswan: I'm not sure. I hope there isn't much duplication between them, but I have noticed that there's a (non-obvious) pattern around having to call map(&:to_s) when loading the form in from the session (as the session appears to not use HashWithIndifferentAccess)

config/locales/en.yml
starswan: Actually I have no idea. We're currently inconsistent - I was about to say this way was right, and that we should ask James' opinion

starswan: I've fixed up all the inconstistencies

app/views/build_estimates/vehicle_value.html.slim
starswan: > You're going to hate this one... I'm getting flaky behaviour on this input. _every so often_ when I update the value in the field and click 'Save and continue' then click back, the updated value is lost and the previous value is retained. See screen capture: https://user-images.githubusercontent.com/113888736/193307375-5b986188-0693-4996-948f-c9b46c4c6b0a.mov

I've converted the input into a 'pure numeric' as per the style guide. It's not quite clear to me what we should be using https://govuk-form-builder.netlify.app/form-elements/text-input/ - but it seems to hint that numeric_input shouldn't be used

config/locales/en.yml
```diff
@@ -170,6 +186,29 @@ en:
       percentage_owned:
         input: What percentage share of their home does your client legally own?
         input_hint: Your client's name must be on the property deeds, lease or freehold.
+    vehicle_value:
+      caption: Yout client's finances
```
patrick-laa, line 190: Still got a 'Yout' here

app/views/build_estimates/_money_input.html.slim
```diff
@@ -0,0 +1,16 @@
+- if hint_text.present?
+  = form.govuk_text_field field,
+          width:,
+          hint: { text: hint_text },
+          label: { text: label_text, size: "m" },
+          prefix_text: BuildEstimatesHelper::POUND,
+          pattern: "[0-9]*",
+          inputmode: "numeric"
```
patrick-laa, line 8: This may be _way_ beyond the scope of this ticket, but since we're changing this field: I think it's weird that if I enter `500` here then click back, what I see is `500.0`. (Because `#to_s` on an integer value stored as a BigDecimal always appends `.0` )

I think a monetary field should either have zero decimal places or two decimal places, but not one. I can't see an _easy_ fix for this, the best I can come up with is to insert a `value: decimal_as_money_string(form, field)` here, and then adding a helper method along the lines of:

```ruby
  def decimal_as_money_string(form, field)
    current_value = form.object.attributes[field.to_s]
    return if current_value.blank?
    return current_value.to_i if current_value.round == current_value

    sprintf("%.2f", form.object.attributes[field.to_s])
  end
```

Massive overkill?

patrick-laa (commented):
Sorry, possibly opening up a can of worms about number formatting. Happy to accept "this is very clearly out of scope for this ticket" as a response here!

app/views/build_estimates/_money_input.html.slim
starswan: I've decided that the best thing to do is to stop digging here. I've created a new ticket EL-377 to look at this whole area

patrick-laa (approved):

app/controllers/flow/vehicle/finance_handler.rb
```diff
@@ -0,0 +1,25 @@
...
+          VehicleFinanceForm.new(params.require(:vehicle_finance_form).permit(*VehicleFinanceForm::VEHICLE_FINANCE_ATTRIBUTES))
+        end
+
+        def save_data(cfe_connection, estimate_id, form, session_data)
+          age_form = Vehicle::AgeHandler.model(session_data)
+          date_of_purchase = age_form.vehicle_over_3_years_ago ? 4.years.ago.to_date : 2.years.ago.to_date
+          value_form = Vehicle::ValueHandler.model(session_data)
+          cfe_connection.create_vehicle estimate_id, date_of_purchase:,
+                                                     value: value_form.vehicle_value,
+                                                     loan_amount_outstanding: form.vehicle_finance.presence || 0,
```
patrick-laa, line 19: I've just realised there's an edge case bug here: If a user clicks 'Yes' to 'Are there any payments left on the vehicle?', and fills in an amount, then changes their answer to 'No', then submits, loan_amount_outstanding here will be whatever amount they put in, but it _should_ be zero.

starswan: Ooh yes thanks for that

app/views/layouts/application.html.slim
```diff
@@ -28,9 +28,10 @@ html lang="en" class="govuk-template"
 
     div class="govuk-width-container"
 
-      = yield :back
-
       main#main-content.govuk-main-wrapper role="main"
-        = yield
+        = yield :back
```
patrick-laa, line 32: As there's already been UAT back-and-forth over the location of the back link relative to the rest of the page, I think if we're changing the HTML structure we need to do it without changing the appearance of the link.

This shouldn't be too hard to do - I think it's a case of using a govuk utility class to reduce the margin-top of main, and then another one to add the same amount of margin-top to to the `.govuk-grid-row`

patrick-laa (approved):

PR #53, opened 2022-09-30
EL-363 initial branching question

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-363)
Added a branching question page with radio buttons after the start page
Added a referral page for non-legal aid provider users to refer them to alternative sources
Reduce the number of characters used from the branch name to create the release name as it was causing silent deployment failures

Checklist

### DIFF
--- a/app/assets/stylesheets/application.sass.scss
+++ b/app/assets/stylesheets/application.sass.scss
@@ -3,3 +3,4 @@ $govuk-new-link-styles: true;
 $govuk-assets-path: "/";
 
 @import "govuk-frontend/govuk/all";
+@import "panel-grey.scss";
--- a/app/views/referrals/show.html.slim
+++ b/app/views/referrals/show.html.slim
@@ -1,20 +1,17 @@
 .govuk-grid-row
   .govuk-grid-column-two-thirds
-    h1 class="govuk-heading-xl" = t(".referral_heading")
+    .govuk-panel.govuk-panel--confirmation.panel-grey
 
-    p class="govuk-body" = t(".legal_aid_provider")
+      h1.govuk-panel__title.govuk--font-size-36
+        = t(".page_heading")
 
-    h2 class="govuk-heading-m"= t(".sub_heading")
+    h2.govuk-heading-m = t(".sub_heading")
 
     .govuk-list
       ul class="govuk-list govuk-list--bullet"
-        li = link_to t(".item_1"), "https://www.gov.uk/civil-legal-advice", class: "govuk-link govuk-body govuk-!-font-size-16"
-        li = link_to t(".item_2"), "https://www.gov.uk/check-legal-aid", class: "govuk-link govuk-body govuk-!-font-size-16"
+        li = link_to t(".item_1"), "https://www.gov.uk/civil-legal-advice"
+        li = link_to t(".item_2"), "https://www.gov.uk/check-legal-aid"
+        li = link_to t(".item_3"), "https://www.gov.uk/legal-aid"
+        li = link_to t(".item_4"), "https://www.gov.uk/guidance/civil-legal-aid-means-testing"
 
-  .govuk-grid-column-one-third
-    .govuk-header__container.govuk-width-container
-    div class="govuk-!-padding-bottom-4"
-    h3.govuk-heading-m = t(".related_content_title")
-    ul.govuk-list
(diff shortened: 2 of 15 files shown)


config/locales/en.yml
```diff
@@ -25,7 +25,10 @@ en:
         property_form:
           attributes:
             property_owned:
-              inclusion: Please select the option that best decribes your clients property ownership
+              inclusion: Please select the option that best describes your clients property ownership
```
patrick-laa, line 28: As you're fixing typos in this line, could you possibly add the missing apostrophe to "client's"?

spec/requests/provider_users_spec.rb
```diff
@@ -0,0 +1,46 @@
+require "rails_helper"
```
patrick-laa, line 1: I'm definitely open to a different opinion here, but to me it feels like feature specs would be a more appropriate way of testing these new pages than request specs. We're more interested in "when the user clicks certain things in the browser where do they end up?" than "if an HTTP client makes a particular request, what is in the HTTP response?"

app/views/referrals/show.html.slim
```diff
@@ -0,0 +1,17 @@
+.govuk-grid-row
+  .govuk-grid-column-two-thirds
+    .govuk-panel.govuk-panel--confirmation.panel-grey
```
patrick-laa, line 3: So in the designs, this box is grey, but for us it's green (I think Lel [wrote some custom CSS](https://github.com/ministryofjustice/eligibility-estimate/blob/main/app/assets/sass/main.scss#L71) to define "panel-grey").

I'm not sure if there is a pre-build grey-background panel component. If not, are we allowed/supposed to write some custom CSS ourselves? Or are we strictly sticking to components that are available to us?

patrick-laa (commented):
A couple of discussion points, otherwise looks good.

app/views/referrals/show.html.slim
willc-work: I had assumed this was a standard class but I cant find any reference to it. The page this is based on (https://apply-provisional-driving-licence.service.gov.uk/eligibility/dropout_provisional_licence_before?locale=en) uses the 'panel-grey' class but perhaps it is also uses custom CSS. I couldnt find anything in the GDS docs about when or why you should change the panel colour though.

spec/requests/provider_users_spec.rb
willc-work: to be honest I would appreciate a chat about this next week. I am not sure what really defines whether a test should be in features vs requests.

app/views/referrals/show.html.slim
willc-work: added custom css to match the prototype

app/assets/stylesheets/panel-grey.css
```diff
@@ -0,0 +1,4 @@
+.panel-grey{
+    color: #0b0c0c;
+    background-color: #f3f2f1;
```
patrick-laa, line 3: Should we be using the [colour palette](https://design-system.service.gov.uk/styles/colour/#colour-palette) here, so that we say ` background-color: govuk-colour("light-grey");` instead of `background-color: #f3f2f1;` etc? Or is that overkill?

app/assets/stylesheets/application.sass.scss
```diff
@@ -3,3 +3,4 @@ $govuk-new-link-styles: true;
 $govuk-assets-path: "/";
 
 @import "govuk-frontend/govuk/all";
+@import "panel-grey.scss";
```
patrick-laa, line 6: Ah, hang on. If we import `panel-grey` into the main stylesheet here, then we actually end up putting the new CSS rule onto the page _twice_ - once in the application.css which gets included in the `head` and once in the panel-grey.css that gets included in the body of `referrals/show.html.slim`.

So I think either you remove this import statement or you remove the `stylesheet_link_tag` in the view. As I understand it, `<link rel="stylesheet">` elements in the HTML body are valid, but not considered best practice, but equally, having this CSS rule in the main stylesheet means it's a tiny bit of extra redundant CSS getting loaded on every page, so I'm not sure which is the best option - I reckon either would be fine.

willc-work: I ended up leaving the import statement in and removing the stylesheet from the view.

patrick-laa (approved):

PR #54, opened 2022-09-30
Dry up applicant data handling

starswan:
Remove code duplication around applicant handling


patrick-laa (approved):
I like this. `EstimateData` is a slightly weird class, because it sort of represents a subset of data entered by the user that can be used to calculate next and previous steps. Stopping using it here is a good first step, and I suggest than in a future PR we look to either rename it to make clearer what it's there for, or find a way to remove it entirely.

PR #56, opened 2022-09-30
EL-351: Employment questions page

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-351)

This PR does 2 things:

- First, it presents a UI to collect employment income and tax information from the user, as per the prototype designs
- Second, it converts this information to a format understandable by CFE and saves it (note that, as with other CFE calls, this will not produce expected results if the user fills out this screen more than once by use of the back button)

Note that this PR builds on #51 to take advantage of the correction to how we place 'back' buttons on the page
Checklist

### DIFF
--- a/app/controllers/build_estimates_controller.rb
+++ b/app/controllers/build_estimates_controller.rb
@@ -2,10 +2,11 @@ class BuildEstimatesController < ApplicationController
   include Wicked::Wizard
   include StepsHelper
 
-  steps(*STEPS_WITH_PROPERTY)
+  steps(*ALL_STEPS)
 
   HANDLER_CLASSES = {
     applicant: Flow::ApplicantHandler,
+    employment: Flow::EmploymentHandler,
     monthly_income: Flow::MonthlyIncomeHandler,
     property: Flow::PropertyHandler,
     vehicle: Flow::VehicleHandler,
--- a/app/views/build_estimates/_money_input.html.slim
+++ b/app/views/build_estimates/_money_input.html.slim
@@ -1,13 +1,6 @@
 - hint_text ||= nil
-- if hint_text.present?
-  = form.govuk_number_field field,
-                              width:,
-                              hint: { text: hint_text },
-                              label: { text: label_text, size: "m" },
-                              prefix_text: BuildEstimatesHelper::POUND
-- else
-  = form.govuk_number_field field,
-                              width:,
-                              hint: { text: hint_text },
-                              label: { text: label_text },
-                              prefix_text: BuildEstimatesHelper::POUND
+= form.govuk_number_field field,
+                          width:,
+                          hint: { text: hint_text },
+                          label: { text: label_text, size: "m" },
+                          prefix_text: BuildEstimatesHelper::POUND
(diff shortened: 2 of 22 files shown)


app/controllers/build_estimates_controller.rb
```diff
@@ -2,10 +2,11 @@ class BuildEstimatesController < ApplicationController
   include Wicked::Wizard
   include StepsHelper
 
-  steps(*STEPS_WITH_PROPERTY)
+  steps(*StepListerService.call.map(&:name))
```
starswan, line 5: Would it be possible to separate out the feature from the steps refactor? I'm worried that this is 'too much too soon' (although I agree with the basic premise, that the step switching logic is slowly getting out of hand) but think we may want a 'more radical' solution that can't be defined by YAML

patrick-laa: You're right, it's too much of a refactor to bung in here. I'll remove it for now.

app/views/build_estimates/employment.html.slim
```diff
@@ -0,0 +1,24 @@
...
+      = form.govuk_error_summary t("generic.error_summary_title")
+
+      - EmploymentForm::DECIMAL_ATTRIBUTES.each do |field|
+        = render "money_input",
+                  form:,
+                  field:,
+                  width: 5,
+                  label_text: t(".#{field}.input")
+      = form.govuk_collection_select :frequency,
+                                     @form.frequency_options,
```
starswan, line 19: Is there any specific reason for adding the blank into the options, rather than passing blank: true to govuk_collection_select (it forwards to collection_select)?

patrick-laa: Just my own ignorance! Fixing.

app/views/build_estimates/_money_input.html.slim
```diff
@@ -0,0 +1,13 @@
...
+- if hint_text.present?
+  = form.govuk_number_field field,
+                              width:,
+                              hint: { text: hint_text },
+                              label: { text: label_text, size: "m" },
+                              prefix_text: BuildEstimatesHelper::POUND
+- else
+  = form.govuk_number_field field,
+                              width:,
+                              hint: { text: hint_text },
```
starswan, line 11: I know this is a conversion, but this doesn't look quite right - if hint_text is not present we still provide it?

app/views/build_estimates/_money_input.html.slim
```diff
@@ -0,0 +1,6 @@
+- hint_text ||= nil
+= form.govuk_number_field field,
+                          width:,
+                          hint: { text: hint_text },
+                          label: { text: label_text, size: "m" },
+                          prefix_text: BuildEstimatesHelper::POUND
```
starswan, line 6: Think you mis-interpreted the previous review. Think we need a mini-PR to port this to SLIM or something. The 'if' condition is/was quite a bit unclear

app/views/build_estimates/employment.html.slim
```diff
@@ -0,0 +1,25 @@
...
+                                     :value,
+                                     :label,
+                                     label: { text: t('.frequency.input') },
+                                     options: { include_blank: t("build_estimates.employment.frequency.blank") }
+      = form.govuk_submit t("generic.save_and_continue")
+  = render 'sidebar_with_date'
```
starswan, line 25: As previously noted, each new screen needs an accessibility test

willc-work (approved):
LGTM

PR #67, opened 2022-10-05
EL-323: Cookie banner

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-323)

This PR:
- Prompts the user to accept or reject non-essential cookies on every screen until the user accepts or rejects non-essential cookies
- Their choice is stored for a year
- If they accept non-essential cookies, we switch on Google Analytics
- They can edit their choice at any time on the new `/cookies` page

Checklist

### DIFF
--- a/app/views/cookies/show.html.slim
+++ b/app/views/cookies/show.html.slim
@@ -1,0 +1,96 @@
+h1.govuk-heading-xl
+  | Cookies
+p.govuk-body
+  == t('.paragraph_1', service_name: link_to(t('.service_name'), root_path, class: 'govuk-link'))
+p.govuk-body
+  = t('.paragraph_2', service_name: t('.service_name'))
+p.govuk-body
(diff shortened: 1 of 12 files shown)


app/views/cookies/show.html.slim
```diff
@@ -0,0 +1,96 @@
...
+    tr.govuk-table__row
+      th.govuk-table__header[scope="row"]
+        | optional_cookie_choice
+      td.govuk-table__cell
+        = t('.saves_settings')
+      td.govuk-table__cell
+        = t('.one_year')
+    tr.govuk-table__row
+      th.govuk-table__header[scope="row"]
+        = request.env['rack.session.options'].instance_variable_get('@by').key
```
willc-work, line 36: Can we use something with a more descriptive name here? also in lines 78 and 85 the names are not very clear. Perhaps we change the _ga to GoogleAnalytics and rename this one to session_cookie?

PR #69, opened 2022-10-06
EL-382: Graceful error handling

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-382)

* The default error page is now styled as per the gov.uk design system guidance
* The error message returned by CFE 422 errors will now be included in the payload sent to Sentry

Checklist


app/lib/cfe_connection.rb
```diff
@@ -125,6 +125,8 @@ def api_result(assessment_id)
 
   def create_record(assessment_id, record_type, record_data)
     cfe_connection.post "/assessments/#{assessment_id}/#{record_type}", record_data
+  rescue Faraday::UnprocessableEntityError => e
+    raise e, "CFE returned the following message: #{e.response_body}", e.backtrace
```
starswan, line 129: I wondering if it might be clearer just to turn off the raise_error Faraday middleware, and then check the result code for success here and then raise the custom exception. We could also add the URL of the request (or maybe just the record_type so show which call failed)

app/lib/cfe_connection.rb
```diff
@@ -124,13 +124,16 @@ def api_result(assessment_id)
 private
 
   def create_record(assessment_id, record_type, record_data)
-    cfe_connection.post "/assessments/#{assessment_id}/#{record_type}", record_data
+    url = "/assessments/#{assessment_id}/#{record_type}"
+    response = cfe_connection.post url, record_data
+    return if response.status == 200
+
+    raise "Call to CFE url #{url} returned status #{response.status} and message:\n#{response.body}"
```
starswan, line 131: Sorry I just realised that api_result can fail too (it fails with a 422) when something goes wrong.
Also IMHO it might be better to check result for success (result.success? I think) as POST requests are often *meant* to return 201 rather than 200 (although I think CFE doesn't do this currently)

starswan (approved):

PR #75, opened 2022-10-07
EL-362 assets screens

starswan:
https://dsdmoj.atlassian.net/browse/EL-362

Complete asset screen by adding valuables over £500


app/controllers/flow/asset_handler.rb
```diff
@@ -12,23 +12,40 @@ def form(params)
...
+        }.select { |asset, _value| form.assets.include? asset.to_s }.map(&:last)
+        cfe_connection.create_capitals estimate_id, liquid_assets, illiquid_assets if liquid_assets.any? || illiquid_assets.any?
 
         if form.assets.include?("property")
+          second_property = {
+            value: form.property_value,
+            outstanding_mortgage: form.property_mortgage,
+            percentage_owned: form.property_percentage_owned,
+          }
+          cfe_connection.create_properties(estimate_id, main_home(session_data), second_property)
```
patrick-laa, line 28: Discussion point: As we've currently architected it, this may cause CFE to return incorrect assessments, because if the client owns their home, we'll already have called `create_properties` on submission of the property screen, so calling it again now and passing in the main home details again means that CFE will have _two_ main homes recorded (because CFE doesn't overwrite itself).

For this to work, we need to _not_ call `create_properties` on submission of the property_entry page, and only call it here, with or without a second property.

Now, that might seem kinda redundant because we know that we're going to move all the CFE logic around anyway. But when we do, the same rule applies: We can't call the `create_properties` endpoint twice, so we need to either call it from the asset_handler logic or the property_entry_handler logic, but not both. So I reckon we might as well get the behaviour correct now, and reduce the rewriting we have to do later.

patrick-laa (approved):

PR #76, opened 2022-10-07
Refactor CFE_HOST to make it more flexible

willc-work:
Whilst trying to learn how EFE interacts with CFE I set this up to be able to watch what happens in the console for both services. This allows us to set a local ENV_VAR to CFE UAT, CFE STAGING, CFE PROD or localhost.

You can point the env file to `http://localhost:3002`
and then run `rails s -p 3002 `to start CFE locally

This also adds the URL as a secret to allow us to connect UAT to UAT. Staging to Staging and eventually Prod to Prod

Checklist


.env.sample
```diff
@@ -1,3 +1,4 @@
 # These environment variables are required to run the application locally
 
-SENTRY_DSN:this-is-where-you-enter-the-sentry-dsn
+SENTRY_DSN=this-is-where-you-enter-the-sentry-dsn
+CFE_HOST=enter-a-vaid-url-here-or-http://localhost:3000
```
starswan, line 4: IMHO this is a poor default - CFE won't be running locally, and I don't think we want to always run EFE + CFE + (the other API) on our dev machines.

willc-work: I have the following in my local .env file:
CFE_HOST=https://check-financial-eligibility-staging.cloud-platform.service.justice.gov.uk
#CFE_HOST=http://localhost:3002
So I can easily switch between staging and local. When developing on both of these services we may want to have both running locally. Similarly this is where you would enter the uat branch for CFE e.g ap-3415-update-uat-postgres-check-financial-uat.cloud-platform.service.justice.gov.uk if you want to test local EFE against a CFE UAT branch.

starswan: I agree. I just don't think it's a very good default. I don't have the configuration that you have, and hopefully we won't need this for local (as we have the Legal Framework API to deal with as well). IMHO the 'out of the box' setup should be simple and 'just work' i.e. 'rake spec' runs without having to know the address of the staging CFE instance)

patrick-laa (approved):

PR #81, opened 2022-10-10
EL-375: Spike results screen

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-375)

- Added a view model to provide easy access to key parts of the raw CFE response
- Built the basic page structures as per the designs
- Displayed the client income content

Checklist


app/controllers/build_estimates_controller.rb
```diff
@@ -20,7 +20,7 @@ def show
     handler = HANDLER_CLASSES[step]
     @form = if handler
               handler.model(session_data)
-            else
+            elsif step == :results
```
starswan, line 23: I have been thinking that this last step doesn't belong here - we've stopped the 'building' phase so maybe this should be a 'create estimate' (in estimates controller). WDYT? Should we have another PR for it?

patrick-laa: I think that's a very sensible idea, and also that it should be in scope for this ticket, so I'll sort it.

patrick-laa:
@willc-work drawing your attention to the fact that this PR, at @starswan's suggestion, moves the results page from the BuildEstimatesController to the EstimatesController. So if we do it this way, the single call to CFE that you're working on would now be trigger in EstimatesController#create. Obviously very happy to be told if this is a problem.

starswan (approved):

PR #82, opened 2022-10-10
EL-377: Use a forgiving, nicely formatted money input

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-377)

- We allow users to enter "," in their financial amounts, and discard them when we cast them into a model
- We also discard anything beyond 2 decimal places
- When rendering a form input containing a monetary value a user has previously inputted:
  - If the user had previously entered no decimal places, or the value is already rounded to the nearest pound, we don't display pence
  - Otherwise we display pence to 2 decimal places
  - In all cases we show "," thousand separators to improve readability 
- (Also fix a minor bug where we were re-sending the student loan value to CFE even after the student loan checkbox had been unchecked)

### DIFF
--- a/app/views/shared/_money_input.html.slim
+++ b/app/views/shared/_money_input.html.slim
@@ -1,3 +1,7 @@
-= form.govuk_number_field field, width:,
-        label: { text: label_text },
-        prefix_text: BuildEstimatesHelper::POUND
+= form.govuk_text_field field,
+                        width:,
+                        label: { text: label_text },
+                        prefix_text: BuildEstimatesHelper::POUND,
+                        value: decimal_as_money_string(form, field),
+                        pattern: "[0-9]*",
+                        inputmode: "numeric"
--- a/app/forms/assets_form.rb
+++ b/app/forms/assets_form.rb
@@ -12,12 +12,12 @@ class AssetsForm
   ASSETS_ATTRIBUTES = (ASSETS_DECIMAL_ATTRIBUTES.keys + ASSETS_PROPERTY_ATTRIBUTES + [:property_percentage_owned]).freeze
 
   ASSETS_DECIMAL_ATTRIBUTES.each do |asset_type, threshold|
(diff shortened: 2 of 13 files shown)


app/controllers/flow/monthly_income_handler.rb
```diff
@@ -16,7 +16,9 @@ def save_data(cfe_connection, estimate_id, income_form, _session_data)
         # TODO: CFE will raise an error the second time `create_student_loan` is called
         # for a given estimate_id, meaning that submitting the monthly income page
         # twice by using the back button can raise an error
-        cfe_connection.create_student_loan estimate_id, income_form.student_finance
+        if income_form.monthly_incomes.include?("student_finance")
+          cfe_connection.create_student_loan(estimate_id, income_form.student_finance)
+        end
```
starswan, line 21: does this belong in this changeset?

starswan (dismissed):

app/views/shared/_money_input.html.slim
```diff
@@ -1,3 +1,7 @@
-= form.govuk_number_field field, width:,
-        label: { text: label_text },
-        prefix_text: BuildEstimatesHelper::POUND
+= form.govuk_text_field field,
+                        width:,
+                        label: { text: label_text },
+                        prefix_text: BuildEstimatesHelper::POUND,
+                        value: decimal_as_money_string(form, field),
+                        pattern: "[0-9]*",
+                        inputmode: "numeric"
```
starswan, line 7: Having approved this, I just wondered if we should have had a test that demonstrates adding commas during the input of a mortgage (say). WDYT?

patrick-laa: Good idea, will add!

patrick-laa: Thank goodness you suggested this, because there was a bug that the test demonstrated!

starswan (approved):

PR #84, opened 2022-10-10
EL-350 Add simple case details page to flow with just 2 proceeding types

starswan:
https://dsdmoj.atlassian.net/browse/EL-350

Simple case details (with 2 options only) to generate a single proceeding type in CFE


app/forms/proceeding_type_form.rb
```diff
@@ -0,0 +1,9 @@
+class ProceedingTypeForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  ATTRIBUTES = [:proceeding_type].freeze
+
+  attribute :proceeding_type
+  validates :proceeding_type, presence: true
```
patrick-laa, line 8: I think for completeness we should also validate `inclusion: { in: Flow::CaseDetailsHandler::PROCEEDING_TYPES }`, as we do that for other constrained options

patrick-laa (approved):
One teeny-teeny tiny suggestion, but I really don't think it's worth going through another review loop for it.

PR #88, opened 2022-10-12
EL-365: Check answers

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-365)

- Use the existing StepsHelper logic to decide what sections to show on the 'check answers' page
- Use a new presenter, backed by a config file, to populate the contents of each section.
- Make the property and property_entry screens into a single mini-loop like vehicles

Checklist


app/presenters/check_answers_presenter.rb
```diff
@@ -0,0 +1,62 @@
...
+    if field_data[:requires_inclusion_in]
+      key = field_data[:requires_inclusion_of] || field_data[:attribute]
+      return unless @session_data[field_data[:requires_inclusion_in]]&.include?(key)
+    end
+    @session_data[field_data[:attribute]]
+  end
+
+  def valid_step?(step_name)
+    steps_list_for(@model).flatten.include?(step_name&.to_sym)
+  end
```
starswan, line 61: I think the only comment I have is that I'm worried about steps_list_for(x0 - other than that, looks really neat.

app/presenters/check_answers_presenter.rb
```diff
@@ -0,0 +1,62 @@
...
+
+    section_data[:subsections].map { build_subsection(_1, section_data) }.select { _1.fields.any? }
+  end
+
+  def build_subsection(subsection_data, section_data)
+    OpenStruct.new(label: subsection_data[:label],
+                   screen: subsection_data[:screen],
+                   fields: build_fields(subsection_data[:fields],
+                                        subsection_data[:screen] || section_data[:screen],
+                                        subsection_data[:label] || section_data[:label]))
```
starswan, line 31: We could maybe make this a bit tighter by having two Structs (with keyword_init: true) rather than OpenStruct. OS upsets me as it responds to all methods with nil.

patrick-laa:
Adjusted as per discussion with @starswan to avoid calling `steps_list_for`, which is really an implementation internal of the current StepsHelper logic.

starswan (approved):

patrick-laa:
Following UAT we:

- Added the missing date box
- Used HTML-escaped % signs
- Allow the property mortgage field to link back to the property screen if the user has no mortgage

Also rebased to main which means:

- Changes to ensure we passed the now-mandatory slim-lint
- Change to ensure we passed the now-mandatory missing-l10n-key check

I'm also NOT addressing a couple of UAT comments, because the steer from Lel was that the current behaviour is as-intended, namely that on the assets screen, if a user selects that they have no assets, the fields in that section for the possible assets should all display as 'Not applicable' rather than not being displayed at all.

config/locales/en.yml
```diff
@@ -331,7 +329,7 @@ en:
       other_fields:
         property_value: "Second property or holiday home: value"
         property_mortgage: "Second property or holiday home: outstanding mortgage"
-        property_percentage_owned: "Second property or holiday home: % owned"
+        property_percentage_owned_html: "Second property or holiday home: &#37; owned"
```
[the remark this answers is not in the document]
patrick-laa: I did a bit of research and [what little I could dig up](https://stackoverflow.com/questions/20290130/how-to-deal-with-semantic-symbols-for-accessibility-in-html) suggested that in general screen readers, if faced with a symbol, will read out the name for that symbol. Which in this case would be the desired behaviour. (It might say "percent owned" or even "percent symbol owned" instead of "percentage owned", but both are accurate representations of the message on-screen.)

I'm happy to be corrected, but in the absence of obvious docs on this case, my instinct would be to wait for a broader content/accessibility review by a subject matter expert. 

In the meantime, I thought the least I could do was use a proper HTML representation of the character in question.

PR #90, opened 2022-10-13
EL-378: Refactor screen sequencing

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-378)

- Removed the current logic to mix-and-match partial arrays with a short config file that lists each screen and the conditions in which each is skipped.

Checklist


willc-work (commented):
Looks good and makes sense to me. Just commenting as I am not sure if this was intended to be merged or if it is an outline for discussion.

One note, in app/lib/screens.yml where you use children: to denote sub-sections of the page, could we use some alternative wording?  Child/children are likely to appear in the service with their non technical meaning and just for the sake of clarity. Not a major issue though.

patrick-laa:
@starswan I'd love your take on this, because I know you've given it some thought. My reasoning is this:

I want the following to be easily human-readable:
* What the screens are
* What order they go in
* In what circumstances things are skipped

The current StepsHelper isn't providing that for me - trying to bring the property and property_entry screens into a single save group earlier this week cost me half an hour of cursing with nested arrays.

_To me_, this yml file is really easily readable, but I appreciate readability is very subjective.

There's definitely a risk around defining branching logic as config. On balance I'm happy with it, because I _believe_ that all the logic around whether to show/hide certain screens can be made to hinge on a series of boolean attributes of the Estimate model, in which case all we need are these fairly intuitive `skip_if` and `skip_unless` lines. If I'm wrong, we may end up getting kludgy.

patrick-laa:
@exonian it would be great to get your thoughts on this too. There's a piece of broader context, which is that knowledge of `what screens to show in what circumstances` is essentially knowledge of the eligibility ruleset, so the sort of thing that we want to keep reasonably separate from most of the rest of the codebase and potentially reasonably close to other bits of the codebase that contain other sorts of knowledge about the eligibility ruleset, but we've not as yet made a hard decision about where eligibility-ish knowledge sits in the architecture.

patrick-laa:
@starswan that's a really good point. If the question is "given the current state, what's the next screen?", it feels indirect to ask "given the current state, what are _all_ valid screens?", and then work out next screen from that.

The only counter-argument I can think of is this: In #88 it's really useful to access the list of valid screens for a given state in order to populate the right sections of the check answers screen. Since it's useful to have a method to work out all valid screens, we might as well use the work done in that method to work out the next/previous screen. But maybe  _all_ that work should be within the service layer, rather than half being in a service and half being in a helper.

I'll try shifting some stuff around and see if that makes things nicer.

app/controllers/build_estimates_controller.rb
```diff
@@ -8,9 +8,13 @@ def update
...
       estimate = load_estimate
-      handler.save_data(cfe_connection, estimate_id, @form, session_data) if last_step_in_group?(estimate, step)
 
-      redirect_to wizard_path next_step_for(estimate, step)
+      # TODO: Stop saving data here
+      if Screens::TemporarySaveDeterminationService.call(estimate, step)
+        handler.save_data(cfe_connection, estimate_id, @form, session_data)
+      end
+
+      redirect_to wizard_path Screens::NextScreenNameService.call(estimate, step)
```
starswan, line 17: I think this is the crux of the issue we are trying to solve. I think we should be thinking about data rather than screens, and also trying to find some cohesive objects somewhere inside the problem. The CFE spike raises the same time of question (how do we cope with 'save random blob of data from session into CFE?' in a way that it comprehensible)

starswan: possibly the wizard gem has led us slightly down a blind alley here - it's thinking about knowing 'flow' between screens, and we can't know the answer in any reliable fashion

patrick-laa:
Closing this. We're moving in a slightly different direction - see #116 .

PR #91, opened 2022-10-14
EL-402: Bug in communicating weekly income to CFE

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-402)

* Adjust number of "payslips" sent to CFE, which has hidden opinions about the significance of that number


willc-work (approved):
I have also applied this change to my branch (el-383) as it involves moving the save_data method.

PR #94, opened 2022-10-14
El 383 call cfe once

willc-work:

Move the calls to CFE out of the individual handlers and into the `/estimates_controller`
Remove the CFE calls from the specs

Question: Is this the best way to do this? Any suggestions on refactoring? Or preferred alternatives?

Also where would tests for these sit? I am assuming in `spec/controllers/estimates_controller_spec.rb` but wasn't sure `spec/requests` was more appropriate.

Still to do: Add tests for the moved methods.

Checklist

### DIFF
--- a/app/controllers/build_estimates_controller.rb
+++ b/app/controllers/build_estimates_controller.rb
@@ -34,7 +34,6 @@ class BuildEstimatesController < ApplicationController
     if @form.valid?
       session_data.merge!(@form.attributes)
       estimate = load_estimate
-      handler.save_data(cfe_connection, estimate_id, @form, session_data) if last_step_in_group?(estimate, step)
 
       redirect_to wizard_path next_step_for(estimate, step)
     else
--- a/app/controllers/estimates_controller.rb
+++ b/app/controllers/estimates_controller.rb
@@ -5,23 +5,15 @@ class EstimatesController < ApplicationController
 
   def create
     cfe_estimate_id = cfe_connection.create_assessment_id
-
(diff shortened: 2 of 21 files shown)


patrick-laa:
Ok, some broad opinions from me:

I'm not sure having so much logic in the controller is the right approach. I probably set a bad precedent by lazily stuffing the 'create applicant' logic into the EstimatesController. I think we're sliding towards the "[fat controller](https://blog.appsignal.com/2021/04/14/ruby-on-rails-controller-patterns-and-anti-patterns.html)" anti-pattern. The issue is that controllers are these very complicated objects that are closely tied to Rails-y, HTTP-y stuff. There are lots of 'secrets' in a controller - words you can't use because oops they overwrite some magic method or variable inherited from ApplicationController. Also I can't test them in isolation because I have no idea how to instantiate a controller instance.

So it would be safer, I think, to syphon all the CFE logic off into its own place, like a Service object, and that way the controller (or anything else), can call `CfeService.call(estimate)`. That way it's reusable, testable in isolation, and easy to extract into its own gem or whatever if we ever want to.

It might be overkill, but you could even do what Apply does, and have a service object for each CFE endpoint. The [CFE Submission Manager](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/app/services/cfe/submission_manager.rb) then iterates through each of them (they're packed into a [ServiceSet](https://github.com/ministryofjustice/laa-apply-for-legal-aid/blob/main/app/services/cfe/service_set.rb) array) and just fires them one by one. That way we can test each one in isolation if we need to.

PR #96, opened 2022-10-14
EL-376 complete results screen with client outgoings and capital

starswan:
https://dsdmoj.atlassian.net/browse/EL-376

Complete implementation of results screen with 3 accordions income, outgoings and capital

### DIFF
--- a/app/models/calculation_result.rb
+++ b/app/models/calculation_result.rb
@@ -1,4 +1,6 @@
 class CalculationResult
+  CFE_MAX_VALUE = 999_999_999_999
+
   include ActionView::Helpers::NumberHelper
   def initialize(api_response)
     @api_response = api_response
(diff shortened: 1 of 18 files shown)


app/models/calculation_result.rb
```diff
@@ -20,30 +20,95 @@ def income_contribution
...
+  def total_capital_limit
+    monetise(api_response.dig(:result_summary, :capital, :proceeding_types).map { |pt| pt.fetch(:upper_threshold) }.min)
+  end
+
   def client_income_rows
-    {
-      employment_income: extract_money(:result_summary, :disposable_income, :employment_income, :gross_income),
-      benefits: extract_money(:assessment, :gross_income, :state_benefits, :monthly_equivalents, :all_sources),
+    data = {
+      employment_income: api_response.dig(:result_summary, :disposable_income, :employment_income, :net_employment_income),
```
patrick-laa, line 53: In UAT of #81  Mike Hooper asked me to change from net_employment_income to gross_income for this field, because the `net` value CFE returns has removed not just NI and tax but also the £45 employment disregard, so shows a very user-unfriendly number if not explained further.

config/locales/en.yml
```diff
@@ -288,6 +288,18 @@ en:
       percentage_owned:
         input: What percentage share of their home does your client legally own?
         input_hint: Your client's name must be on the property deeds, lease or freehold.
+    property_entry:
```
patrick-laa, line 291: We seem to be adding this property_entry section twice - lines 291 and 408. But it's already present on line 279, so aren't both of these redundant?

starswan: Managed to remove lots of unused translations using the 'test' setting which is just about to be merged

PR #97, opened 2022-10-14
EL383 call CFE once at the end of the journey

willc-work:

Move the save_data methods to the build estimates controller to call CFE once at end of journey
refactor the CFE calls into a separate CFEService class

To: tests to be added

Checklist


app/controllers/estimates_controller.rb
```diff
@@ -1,5 +1,23 @@
...
+    CfeService.call(cfe_estimate_id, cfe_session_data)
+    @model = cfe_connection.api_result(cfe_estimate_id)
+
+    render :show
+  end
+
+  private
+
+  def cfe_estimate_id
+    @cfe_estimate_id ||= cfe_connection.create_assessment_id
```
starswan, line 17: Think this should be encapsulated inside the CfeService

app/services/cfe_service.rb
```diff
@@ -0,0 +1,156 @@
...
+  def self.call(cfe_estimate_id, cfe_session_data)
+    new.call(cfe_estimate_id, cfe_session_data)
+  end
+
+  def call(cfe_estimate_id, cfe_session_data)
+    applicant_screen_create cfe_estimate_id, Flow::ApplicantHandler.model(cfe_session_data)
+    save_employment_data cfe_estimate_id, Flow::EmploymentHandler.model(cfe_session_data)
+    save_monthly_income_data cfe_estimate_id, Flow::MonthlyIncomeHandler.model(cfe_session_data)
+    save_vehicle_value_data cfe_estimate_id, Flow::Vehicle::ValueHandler.model(cfe_session_data)
+    save_vehicle_finance_data cfe_estimate_id, cfe_session_data
```
starswan, line 11: I think this saves the vehicle data twice. One of the commented-out tests would probably show this up

app/services/cfe_service.rb
```diff
@@ -0,0 +1,156 @@
...
+    save_assets_data cfe_estimate_id, cfe_session_data
+    save_outgoings_data cfe_estimate_id, cfe_session_data
+    save_property_entry_data cfe_estimate_id, Flow::PropertyEntryHandler.model(cfe_session_data)
```
starswan, line 14: It would be nice to use the old code consistently in this layer, rather than sometimes passing naked session data and other times passing the handler that knows how to deal with it. I think the structure needs some more thought - I think something needs to understand 'Vehicle' data as a whole.

app/services/cfe_service.rb
```diff
@@ -0,0 +1,156 @@
...
+    create_applicant cfe_estimate_id, Flow::ApplicantHandler.model(cfe_session_data)
+  end
+
+  def cfe_connection
```
starswan, line 18: All these methods should be marked private as I don't think they are part of the public interface.

app/services/cfe_service.rb
```diff
@@ -0,0 +1,156 @@
...
+  # CFE doesn't understand about annual salary or 'total income in the last 3 months',
+  # so for both those use cases we must convert the figures we have into what they would
+  # be if they were paid as regular monthly income
+  def multiplier(form)
+    case form.frequency
+    when "annually"
+      1.0 / 12
+    when "total"
+      1.0 / 3
+    else
```
starswan, line 139: The scope of these methods (in particular) make me think that everything in here could be a class method rather than an instance method.

PR #101, opened 2022-10-17
EL-403: Enforce explicit l10n in tests

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-403)

- Adjusted config so the test runner will raise an exception if the L10n file doesn't contain a key that's requested by the code
- Made immediate fixes to exceptions raised in the exiting test suite.

Checklist

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -446,14 +446,3 @@ en:
       reduce_requests: Used to reduce the number of requests.
       save: Save cookie settings
       accept_cookies: Do you want to accept analytics cookies?
-  layouts:
-    application:
-      footer:
-        support: Support links
-        contact: Contact
-        feedback: Feedback
-        cookies: Cookies
-        privacy_policy: Privacy policy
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -446,3 +446,14 @@ en:
       reduce_requests: Used to reduce the number of requests.
       save: Save cookie settings
       accept_cookies: Do you want to accept analytics cookies?
+  layouts:
```
starswan, line 449: I think these are in layouts.yml, which needs to be renamed as en.layouts.yml to be picked up properly?

patrick-laa: Ooohhhh.... of course, sorry, that makes loads of sense.

starswan (approved):
Not sure if we needed the rename - it was conflated (for me) by the missing 'support' translation, but it doesn't do any harm.

PR #104, opened 2022-10-17
EL-406 return to same section

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-406)

When returning to check answers screen from check answers flow, specify an anchor so that we return to the relevant section of the screen.

Checklist


app/services/check_answers/relevant_section_finder_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      end
+    end
+
+    def matching_field?(subsection, step)
+      subsection[:fields]&.any? do |field|
+        field[:screen] == step.to_s
+      end
+    end
+  end
+end
```
starswan, line 28: I'm feeling that this data (that used to be owned by one module/class) has now 'escaped' into another module/class, and the old  'service' does nothing apart from loading the YML file. 
Could it not be re-organised into methods rather than hash access? [Sorry can't think of a better suggestion at the moment]

patrick-laa: You're right, we now have a Presenter and a Service that both know about an idiosyncratic data structure. I'll have a ponder.

patrick-laa:
RIght, I think this is better:

The only thing that knows about the yml file is one service. Then both the check answers view and a new service know that the original service outputs an array of Section objects that contain Subsections that contains Fields, and the new service does a minimal bit of work to essentially invert that tree in order to provide an answer.

app/services/check_answers/relevant_section_finder_service.rb
```diff
@@ -0,0 +1,30 @@
...
+              key: field.screen || subsection.screen || section.screen,
+              value: section.label,
+            }
+          end
+        end
+      end
+
+      dictionary = nested_pairs.flatten.map { [_1[:key], _1[:value]] }.to_h
+
+      dictionary[@step]
```
starswan, line 27: could this be dictionary.fetch(@step) ?

starswan, line 27:
```suggestion
      dictionary.fetch(@step)
```

starswan (approved):
Added 1 tiny comment

PR #108, opened 2022-10-19
EL 383 single call to  CFE

willc-work:

Refactor the CFE call so that instead of a call after each page we make one call at the end of the application.
Create a CFE service that calls individual services that process the data and call each CFE endpoint
Fix all the existing tests

Checklist


spec/features/applicant_page_spec.rb
```diff
@@ -11,6 +11,7 @@
     before do
       allow(CfeConnection).to receive(:connection).and_return(mock_connection)
       allow(mock_connection).to receive(:create_proceeding_type)
+      allow(mock_connection).to receive(:create_regular_payments)
```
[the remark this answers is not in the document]
willc-work: I have removed these mocks and the allow(CfeConnection) to where they are actually needed (below line 99)

app/controllers/estimates_controller.rb
```diff
@@ -1,22 +1,23 @@
...
+    redirect_to estimate_build_estimates_path SecureRandom.uuid
   end
 
   def create
-    # The last thing to do before retrieving the result is to tell CFE
-    # about the applicant
-    create_applicant
-
-    @model = cfe_connection.api_result(params[:cfe_id])
+    cfe_estimate_id
```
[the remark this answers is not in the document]
willc-work: Ignore me. It took less than 5 minutes to do and the tests passed first time so I have made the suggested change,

app/services/submit_vehicle_service.rb
```diff
@@ -0,0 +1,39 @@
...
+                                                   loan_amount_outstanding: 0,
+                                                   in_regular_use: value_form.vehicle_in_regular_use
+  end
+
+  def submit_vehicle_finance_data(cfe_estimate_id, cfe_session_data)
+    finance_form = Flow::Vehicle::FinanceHandler.model(cfe_session_data)
+
+    # line below is missing branch coverage. However I am not sure it is required. we validate on the page
+    # that a value is entered if vehicle_pcp is selected, if no value is entered the estimate cannot progress and therefore no submission
+    # to CFE is possible.
```
patrick-laa, line 27: Again, if we've got logic that's not being tested and _can't_ be tested by a high-level spec, we can either knock up a quick unit test, or, if you think this logic is completely redundant, you could just remove the untested logic.

app/controllers/estimates_controller.rb
```diff
@@ -1,22 +1,17 @@
...
-    redirect_to estimate_build_estimates_path cfe_connection.create_assessment_id
+    redirect_to estimate_build_estimates_path SecureRandom.uuid
   end
 
   def create
-    # The last thing to do before retrieving the result is to tell CFE
-    # about the applicant
-    create_applicant
-
-    @model = cfe_connection.api_result(params[:cfe_id])
```
patrick-laa, line 11: Since we're not doing this, we can remove `def cfe_connection` from ApplicationController, which is one of the places where we don't have line coverage

app/services/submit_vehicle_service.rb
willc-work: removed after ensuring that it didnt break anything else and that I was sure of my logic

PR #109, opened 2022-10-20
El-352 Benefits page

starswan:
Create pages to capture benefits, including a loop where multiple benefits are captured until user tells us capture is complete

https://dsdmoj.atlassian.net/browse/EL-352


app/lib/cfe_connection.rb
```diff
@@ -93,6 +93,23 @@ def create_regular_payments(assessment_id, income_form, outgoings_form)
     create_record(assessment_id, "regular_transactions", regular_transactions:) if regular_transactions.any?
   end
 
+  def create_benefits(assessment_id, benefits)
+    state_benefits = benefits.map do |benefit|
+      payments = [2, 1, 0].map do |v|
```
patrick-laa, line 98: I'll do a full review tomorrow, but one thing to flag up immediately: CFE won't believe that your payments are fortnightly or weekly if you only provide 3 instances of them. [It thinks](https://github.com/ministryofjustice/check-financial-eligibility/blob/main/app/services/utilities/regular_period_analyser.rb) that you are providing 3 months' worth of benefits payments, so if they're weekly there should be 12 or 13 of them, and if they are fortnightly there must be 6 or 7 of them. See the logic in EmploymentHandler (added as a bugfix for EL-402)

app/views/estimate_flow/benefits_more.html.slim
```diff
@@ -0,0 +1,25 @@
...
+  span.govuk-caption-l = t(".caption")
+  h1.govuk-heading-l = t(".heading.text", count: @form.benefits.size)
+
+  = govuk_table do |table|
+    - table.body do |body|
+      - @form.benefits.each do |item|
+        - body.row do |row|
+          - row.cell(text: item.fetch("benefit_type"))
+          - row.cell(text: item.fetch("benefit_amount"))
+          - row.cell(text: item.fetch("benefit_frequency"))
```
patrick-laa, line 14: Is this PR deliberately leaving out the change/remove links from here?

config/locales/en.yml
```diff
@@ -134,6 +151,29 @@ en:
     related_content_title: Related content
     not_applicable: Not applicable
   estimate_flow:
+    benefits_more:
+      caption: Your client's finances
+      heading:
+        text: You have added %{count} benefits
```
patrick-laa, line 157: You'll need to pluralize `benefit` appropriately so that it doesn't say "1 benefits", won't you?

app/views/estimate_flow/benefits_more.html.slim
starswan: oops yes forgot that bit

patrick-laa:
Closing as this functionality was merged in #121

PR #110, opened 2022-10-20
EL-359: Client outgoings page

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-359)

Rework the outgoings page to meet new designs, and send additional outgoings information to CFE.

Note that I'm not touching the results page because it looks like #96 has that covered

Checklist


app/lib/cfe_connection.rb
```diff
@@ -55,6 +55,13 @@ def create_employment(assessment_id, employment_income)
     create_record(assessment_id, "employments", employment_income:)
   end
 
+  CFE_OUTGOING_FREQUENCIES = {
```
exonian, line 58: Not necessarily a blocker but a note that usually (`OUTGOINGS_ATTRIBUTES` now, the form name, the handler etc) we  used `outgoings` but sometimes (`OUTGOING_ATTRIBUTES` previously, and here) we use 'outgoing' singular. If this isn't deliberate, we should be consistent with 'outgoings'.

However, if it _is_ to distinguish between 'outgoings' as in money paid out and 'outgoing' as in regarding an external call to CFE... 😅

config/locales/en.yml
```diff
@@ -53,8 +53,22 @@ en:
               greater_than: Other income must be greater than £0
         outgoings_form:
           attributes:
-            outgoings:
-              at_least_one_item:  Please select at least one option
+            housing_payments_value:
+              blank: Please enter a 0 if your client makes no housing payments
```
exonian, line 57: Is this copy from anywhere? I note that we use "enter 0" in the screengrab and "enter zero" in the additional acceptance criteria for the top of the page guidance. I can't see these error strings defined elsewhere but I just note that "a 0" struck me as odd in that it was different. It's entirely possible I'm just missing where this was designed

exonian (approved):
Couple of moments of pedantry. Happy for you to take them under advisement and either revise or press forward with it. If you make changes, I'll review again, but if you think they're fine as is I don't need to be a blocker by commenting now and you needing to come back to me.

The CFE calls were slightly beyond my confidence in spotting issues but not so far beyond that I have no confidence; they seem very plausible and nothing leapt out to me, so I think it's sensible to accept the risk.

patrick-laa:
@exonian both were valid points, so have made changes as requested.

exonian (approved):
This reads more nicely to me; it's still slightly different from the other text ("a zero" vs "zero") but let's just :shipit: because without specific intent I honestly don't know which is better in the context of an error message and this shouldn't delay us further

PR #111, opened 2022-10-20
EL-415: Remove irrelevant guidance

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-415)

Removed irrelevant immigration and asylum guidance.

Checklist


starswan (approved):
Should we comment out rather than delete?

patrick-laa:
I think given that all we had was unvetted wording for non-existent links, we lose nothing of value by leaving in something commented out that will definitely not be resurrected in time for private beta and, the future being uncertain, may never be resurrected.

PR #116, opened 2022-10-21
Refactor of StepsHelper into 3 independent sections

starswan:
Refactor StepsHelper into 3 sections (applicant, income/outgoings and capital)

This detangles StepsHelper quite a bit (moving property rules into the capital section rather than being attached to income) 
and is hopefully a good first step towards a better structure based on data groupings.

It also moves the CFE call from the 'income' section into the 'outgoings' handler (as they are always performed together)


patrick-laa (approved):
This ticks two boxes for me.

First, StepsHelper no longer makes me cry. (CapitalSection is still a bit hard to read, but I think the upcoming squashification of the vehicle screens is going to solve that problem for me)

Second, we've not made it _harder_ for ourselves to start splitting up the app's knowledge into sections (vehicles, benefits etc) rather than functions (decide which screen to show, save to CFE etc).

So this seems a good direction to me.

PR #117, opened 2022-10-21
EL-414: Add print results screen

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-414)

Added a "Print results" screen that is accessible from the results screen and reuses as much content as possible from it.

Checklist


app/controllers/estimates_controller.rb
```diff
@@ -14,6 +14,15 @@ def create
     render :show
   end
 
+  def print
+    # TODO: At the moment api_result cannot be called twice with the same ID
+    # Therefore this page cannot be accessed after the `create` page has been accessed.
```
exonian, line 19: It's unclear to me what the implication of not being able to access it after that page is. Does it advise how the following code is written, mitigating the problem but in an otherwise non-ideal way? Or is there an ongoing issue that we can't go to print page after that other page? If so, how are we avoiding doing that?

Sorry for being dense, but hopefully others after me would have the same uncertainty so explaining it like I'm 5 will help them too...

exonian (commented):
All looks nice to me, good splitting out into templates. I've commented on one thing, where I'm not confident I understand the inline comments (which are great to have there, and nearly brilliant) so just need it spelled out slightly for the likes of me, I think, as I couldn't tell someone else what the implications of them are

app/controllers/estimates_controller.rb
patrick-laa: @exonian sorry, yes, basically this is blocked by #108 - the single call to CFE - which means that in practice this controller action will _always_ raise an exception. I will amend the comment to make it clearer.

PR #121, opened 2022-10-24
EL-352 create benefits

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-352)

New screen to allow client to list benefits

Checklist


exonian (approved):
I'm afraid I can't speak to much of the CFE-calling aspects, and my understanding of the models and controllers is a bit all over the place, too. But it looks pretty sensible.

Couple of challenges on DRYness and test efficacy, respectively. Both non-blocking so I'll Approve it so that if it's decided that on balance no action is needed I'm not blocking integration by you needing to come back to me for formal approval.

PR #124, opened 2022-10-24
EL-411: Check benefits answers

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-411)

Allow individually entered benefits to be shown on the 'Check Answers' screen

Checklist

### DIFF
--- a/spec/features/check_answers_page_spec.rb
+++ b/spec/features/check_answers_page_spec.rb
@@ -1,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe "Check answers page" do
+  let(:estimate_id) { SecureRandom.uuid }
+
+  context "when I have entered benefits" do
+    before do
+      visit estimate_build_estimate_path(estimate_id, :benefits)
+      find(:css, "#benefits-form-add-benefit-true-field").click
+      click_on "Save and continue"
+      fill_in "Benefit type", with: "Child benefit"
+      fill_in "Enter amount", with: "150"
+      choose "Every week"
+      click_on "Save and continue"
+    end
+
+    scenario "I can modify those benefits from the 'check answers' screen" do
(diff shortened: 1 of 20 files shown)


spec/features/check_answers_page_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe "Check answers page" do
+  let(:estimate_id) { SecureRandom.uuid }
+
+  context "when I have entered benefits" do
+    before do
+      visit estimate_build_estimate_path(estimate_id, :benefits)
+      find(:css, "#benefits-form-add-benefit-true-field").click
```
starswan, line 9: there's a helper method for this - 
```suggestion
      select_boolean_value("benefits-form", "add-benefit", true)
```

patrick-laa: I hadn't spotted that. Thanks!

spec/features/check_answers_page_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    scenario "I can modify those benefits from the 'check answers' screen" do
+      visit estimate_build_estimate_path(estimate_id, :check_answers)
+
+      within("#field-list-benefits") { click_on "Change" }
+      fill_in "Benefit type", with: "Child Benefits"
+      click_on "Save and continue"
+      within("#field-list-benefits") do
+        expect(page).to have_content "Child Benefits"
+      end
+    end
```
starswan, line 26: IMHO it would be good to have a more complicated test here - both adding and removing benefits to make sure that this path works properly

patrick-laa: Ok, will do

starswan (approved):

PR #126, opened 2022-10-25
EL-418: Update vehicle details to be on a single screen

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-418)

Combine vehicle details screens as per new design

Checklist


app/services/submit_vehicle_service.rb
```diff
@@ -4,30 +4,20 @@ def self.call(cfe_estimate_id, cfe_session_data)
...
+  def submit_vehicle_finance_data(cfe_estimate_id, model)
+    date_of_purchase = model.vehicle_over_3_years_ago ? 4.years.ago.to_date : 2.years.ago.to_date
     cfe_connection.create_vehicle cfe_estimate_id,
                                   date_of_purchase:,
-                                  value: value_form.vehicle_value,
-                                  loan_amount_outstanding: finance_form.vehicle_finance.presence,
-                                  in_regular_use: value_form.vehicle_in_regular_use
+                                  value: model.vehicle_value,
+                                  loan_amount_outstanding: model.vehicle_finance.presence || 0,
+                                  in_regular_use: model.vehicle_in_regular_use
```
starswan, line 21: Think there is a tiny bug here - we should check vehicle_pcp being set, not the presence of vehicle_finance (in case they entered it and then changed their mind)

patrick-laa: Great spot! Thank you.

starswan (approved):

PR #129, opened 2022-10-26
El 386 send other income to CFE

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-386)

Add a method to submit other income as an irregular quarterly income
Display the value used for monthly equivalent on the accordion section of the results page

Checklist


app/services/submit_irregular_income_service.rb
```diff
@@ -6,21 +6,34 @@ def self.call(cfe_estimate_id, cfe_session_data)
   def call(cfe_estimate_id, cfe_session_data)
     form = Flow::MonthlyIncomeHandler.model(cfe_session_data)
 
-    if form.monthly_incomes.include?("student_finance")
+    if form.student_finance
```
patrick-laa, line 9: There's a niche edge case that will break this: If you check 'student finance' then fill in money input that appears in the conditional reveal, then _uncheck_ student finance (i.e. you change your mind), then we want to respect the fact that you've indicated you _don't_ want to tell CFE about student finance, even if you have actually given us a value for student finance.

The way around this is to keep `if form.monthly_incomes.include?("student_finance")` here, which tells us if you checked the box, and likewise a couple of lines down to replace `if form.other` with `if form.monthly_incomes.include?("other")`.

If you want to be _super_ defensive you could do both, i.e. `if form.monthly_incomes.include?("student_finance") && form.student_finance`, but that might be overkill, if we trust our model validation.

patrick-laa (approved):
One very small thing, otherwise LGTM.

PR #131, opened 2022-10-27
EL-413 use redis session store

starswan:
https://dsdmoj.atlassian.net/browse/EL-413

Convert session store from cookie to redis for production-like environments so that real users can perform lots of estimates without having to restart their browser or use Incognito mode. The redis keys are set to expire after 14 days just like their CFE counterparts - this is ephemeral data so it should have a limited shelf-life

This is dependant on the corresponding cloud platform PR being merged: https://github.com/ministryofjustice/cloud-platform-environments/pull/8981


config/environments/development.rb
```diff
@@ -32,6 +32,7 @@
 
     config.cache_store = :null_store
   end
+  config.session_store :cookie_store, key: LaaEstimateFinancialEligibilityForLegalAid::SESSION_COOKIE_NAME
```
patrick-laa, line 35: So I _keep_ hitting the CookieOverflow error in development, and I would personally prefer the dev environment to use Redis to avoid having to clear cookies and restart the flow every couple of days.

(Also, if we used Redis in dev as well, we could explore things like storing the CFE result in the session, obviating the need to call CFE _twice_ just to print the results page. At the moment the result payload is big enough to explode the cookie).

If it feels like too much faff to have working Redis in dev I'm OK to be overruled - but maybe for future devs we could add a line in the readme explaining that CookieOverflows are a dev-environment-specific error, and offering a workaround?

starswan: I don't really use the dev environment personally, it seems a bit overkill to have a different server running just for this purpose. I've added a commit to use the memory_store for development which seems to work well for me (and I think I've used this before in MoJ), could you try it out @patrick-laa and see if it works for you?

patrick-laa: @starswan great shout. Yes, that works for me, and I'm confident that the differences in environment between dev/test/prod are small enough that it won't come back to haunt us, and that by saying it explicitly out loud like this I'm not immediately jinxing it.

patrick-laa (approved):

config/environments/development.rb
starswan: @patrick-laa There are a few minor differences (from experience) but it's fine as long as you stick to primitive types and don't try to serialise/deserialise whole objects

PR #132, opened 2022-10-27
EL-427 change confirmation panel to a neutral blue colour

starswan:
https://dsdmoj.atlassian.net/browse/EL-427

Change confirmation panel colour to blue to be a neutral colour as the outcome can be 'unsuccessful'

### DIFF
--- a/app/assets/stylesheets/results.scss
+++ b/app/assets/stylesheets/results.scss
@@ -5,3 +5,8 @@ tr.solid-top-border {
 tr.solid-bottom-border {
   border-bottom: 2px solid #000
 }
+
+// make the confirmation panel a neutral colour
+.panel-blue {
+  background: govuk-colour("blue");
+}
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -412,14 +412,6 @@ en:
       submission_instructions: Now submit your eligibility check
       submission_explanation: We'll use your answers to give you an estimated result on your client's financial eligibility for legal aid.
       call_to_action: Submit eligibility check
-    results:
-      heading: Results Page
-      provided_client_information: You have provided your client's information
-      client-has-to-pay: Based on what you've told us, your client may need to pay towards controlled legal aid work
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/results.scss
```diff
@@ -4,4 +4,9 @@ tr.solid-top-border {
 
 tr.solid-bottom-border {
   border-bottom: 2px solid #000
-}
\ No newline at end of file
+}
+
+// make the confirmation panel a neutral colour
+.govuk-panel--confirmation {
```
patrick-laa, line 10: We _also_ use a `.govuk-panel--confirmation` on /referrals. We currently override it to be grey background, but this new CSS rule takes precedence and turns it blue!

I recommend adding a new class for blue panels that we can apply to the results screen to ensure only the intended panel gets changed

patrick-laa (changes requested):
As well as my below comment, there's an additional requirement in EL-427 about changing the wording in the panel if the user is unconditionally eligible. It's only in comments rather than acceptance criteria so far, because there's a little back and forth about the wording itself. I don't think we can complete this ticket until that's decided.

config/locales/en.yml
```diff
@@ -480,7 +472,7 @@ en:
       partially_eligible: Provisionally eligible
       ineligible: Ineligible
       summary: Your client appears %{eligibility} for legal aid based on the information provided.
-      we_calculated: "We calculated that your client may have to pay:"
+      we_calculated: "We estimate that your client may have to pay:"
```
patrick-laa, line 475: THis isn't quite the change that was requested on the ticket. The scenario in question is where the client is eligible without a contribution, in which case this string isn't shown. The requirement is to put, in its place: "We estimate that your client will not have to pay any financial contribution." (I agree that for consistency we should probably change the wording as you have done above _too_, but that's a separate issue - possibly a separate ticket?)

patrick-laa (approved):

PR #133, opened 2022-10-28
EL-429: Subject Matter of Dispute for property

starswan:
https://dsdmoj.atlassian.net/browse/EL-429

Add SMOD checkboxes to property and second property screens


app/services/submit_assets_service.rb
```diff
@@ -27,8 +28,23 @@ def call(cfe_estimate_id, cfe_session_data)
...
     end
 
-    cfe_connection.create_properties(cfe_estimate_id, main_home, second_property) if main_home.present? || second_property.present?
+    create_properties(cfe_estimate_id, main_home, second_property) if main_home.present? || second_property.present?
+  end
+
+private
+
+  def create_properties(assessment_id, main_property, second_property)
+    main_home = main_property ||
```
patrick-laa, line 40: :+1: to moving this logic

patrick-laa (approved):
A suggestion for a simplification; otherwise all good.

app/forms/property_entry_form.rb
```diff
@@ -18,4 +18,10 @@ class PropertyEntryForm
   validates :percentage_owned,
             numericality: { greater_than: 0, only_integer: true, less_than_or_equal_to: 100, allow_nil: true, message: :within_range },
             presence: true
+
+  attribute :house_in_dispute, array: true, default: []
```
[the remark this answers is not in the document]
starswan: Great thanks - originally used a boolean, but couldn't work out how to fit it in with the form submit. Think multiple: false is what I need

config/locales/en.yml
```diff
@@ -155,6 +155,10 @@ en:
     date: Date
     related_content_title: Related content
     not_applicable: Not applicable
+    dispute: This asset is a subject matter of dispute
```
willc-work, line 158: I think this should say "This asset is the subject matter of dispute" or "This asset is subject matter of dispute". The inclusion of the 'a' scans funny to me. And in the guidance it is referred to using either 'the' or no article at all

PR #137, opened 2022-10-28
EL-430: Add SMOD checkbox to vehicles page

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-430)

Add a 'smod' checkbox to the vehicle details page.
Send a subject_matter_of_dispute flag to CFE when describing vehicle if the checkbox is checked

Checklist

### DIFF
--- a/app/forms/vehicle_details_form.rb
+++ b/app/forms/vehicle_details_form.rb
@@ -6,7 +6,8 @@ class VehicleDetailsForm
                                   vehicle_pcp
                                   vehicle_over_3_years_ago
                                   vehicle_value
-                                  vehicle_in_regular_use].freeze
+                                  vehicle_in_regular_use
+                                  vehicle_in_dispute].freeze
 
   attribute :vehicle_pcp, :boolean
   validates :vehicle_pcp, inclusion: { in: [true, false], allow_nil: false }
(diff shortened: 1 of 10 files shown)


app/forms/vehicle_details_form.rb
```diff
@@ -26,4 +27,6 @@ class VehicleDetailsForm
 
   attribute :vehicle_in_regular_use, :boolean
   validates :vehicle_in_regular_use, inclusion: { in: [true, false], allow_nil: false }
+
+  attribute :vehicle_in_dispute, :boolean, default: false
```
starswan, line 31: Just worked out how to do this using govuk_check_box. It looks like there is a subtle bug that means that the 'magic' rails value doesn't show up by default, but passing a zero as the 3rd argument to govuk_check_box makes everything come together (rather than the default of false)

app/forms/vehicle_details_form.rb
```diff
@@ -26,4 +27,6 @@ class VehicleDetailsForm
 
   attribute :vehicle_in_regular_use, :boolean
   validates :vehicle_in_regular_use, inclusion: { in: [true, false], allow_nil: false }
+
+  attribute :vehicle_in_dispute, :boolean
```
starswan, line 31: It would nice to have a boolean validation to match the others. Got bitten by this in a previous project when the data started coming from another source...

PR #143, opened 2022-10-31
EL-435 re-order property prompt options to match prototype

starswan:
https://dsdmoj.atlassian.net/browse/EL-435

Property with mortgage option is supposed to be before owned_outright

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -155,6 +155,7 @@ en:
     date: Date
     related_content_title: Related content
     not_applicable: Not applicable
+    opens_in_new_tab: (opens in new tab)
   estimate_flow:
     applicant:
       heading: Your client's details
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -274,7 +274,7 @@ en:
         outright: Yes, owned outright
         with_mortgage: Yes, with a mortgage or loan
         none: "No"
-      guidance_on_properties: Guidance on properties (link needed)
+      guidance_on_properties: Guidance on properties (opens in new tab)
```
patrick-laa, line 277: Looking at the GDS guidance Lel linked to, the rule is that "(opens in new tab)" should appear _any_ time a link opens in a new tab. Since that's _every_ sidebar link, I think we should implement this by appending this as text to every link text passed to the sidebar partial, otherwise we'll have to manually go through and add this to _every_ guidance link.

patrick-laa (approved):

PR #149, opened 2022-11-02
EL-407: Specify adult and child dependants

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-407)

Add separate screens for dependants questions.


app/forms/dependant_details_form.rb
```diff
@@ -0,0 +1,11 @@
+class DependantDetailsForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  DEPENDANT_ATTRIBUTES = %i[adult_dependants child_dependants].freeze
+
+  DEPENDANT_ATTRIBUTES.each do |attr|
+    attribute attr, :integer
+    validates attr, presence: true
```
starswan, line 9: I think we should have a check for integer here?

spec/services/submit_dependants_service_spec.rb
```diff
@@ -5,31 +5,70 @@
...
+
           service.call(cfe_estimate_id, session_data)
         end
       end
 
       describe "when there are no dependants" do
         let(:dependant_boolean) { false }
-        let(:dependants) { nil }
+        let(:child_dependants) { 4 }
+        let(:adult_dependants) { 2 }
```
starswan, line 60: don't think we need these 2 lines?

app/forms/dependant_details_form.rb
patrick-laa: @starswan because we've defined these attributes as integers, any non-integer that gets assigned to them gets immediately cast to an integer _before_ it's validated. So when the validation is run, even if you passed in `3.7` or "forty-three", the check for integer will pass, as the cast value (`3` and `0` respectively in the examples given) _is_ an integer.

config/locales/en.yml
```diff
@@ -176,6 +181,25 @@ en:
...
+      caption: About your client
+      legend: Tell us about your client's dependants
+      hint: If some don't apply, enter 0.
+      adult_dependants:
+        label: Adult dependants
+        hint: For example, a parent, grandparent, aunt or uncle.
+        instruction: Enter number of adult dependants
+      child_dependants:
+        label: Child dependants
+        hint: For example, your client's child, grandchild, neice or nephew.
```
willc-work, line 201: niece is spelled incorrectly

spec/features/applicant_page_spec.rb
```diff
@@ -154,60 +119,17 @@
...
-      end
-
-      it "submits dependants" do
-        fill_in "applicant-form-dependant-count-field", with: "2"
-        click_on "Save and continue"
-        expect(page).to have_content benefits_header
-      end
+    it "has a back pointer to the applicant page" do
+      click_on "Back"
+      expect(page).to have_content applicant_header
```
willc-work, line 132: In the past for matching e.g. page titles I have used the following format:

expect(page.body).to include(I18n.t("estimate_flow.dependants.legend"))

This means you don't need the let at the top of the page. Happy to learn that there is a reason not to do it this way also.

starswan: IMHO it's useful to give the thing you're checking for a name - we actually want to say 'we are on page x' and there are many ways of doing that. Originally I started checking the URL, but I felt like it might be more fragile than checking for some basic header content.

PR #151, opened 2022-11-03
Refactor check answers page out of Wizard loop

starswan:
As check_answers is really the 'end' of the wizard loop rather than a 'step' in its own right, move it into the estimates controller to help implementing the more complex check answers behaviour required in EL-388


app/views/estimates/check_answers.html.slim
```diff
@@ -1,5 +1,5 @@
 - content_for :back
-  = back_link(:check_answers, @estimate, @back_buttons_invoke_browser_back_behaviour)
+  = back_link(:check_answers, @estimate, true)
```
patrick-laa, line 2: My instinct is that this back link on this screen should always go back to the last wizard step screen (which happens to always be the assets screen). By using the browser back behaviour, once the user has gone through a mini-loop, clicking back will take them to the final screen of the mini-loop, which feels wrong to me. Happy to be told I'm wrong if you disagree.

willc-work: I don't have strong feelings on this myself as both suggestions make sense, however govuk indicates that it should go back to the previous page:
https://design-system.service.gov.uk/components/back-link/
 >>Make sure the link takes users to the previous page they were on, in the state they last saw it. Where possible, ensure it works even when JavaScript is not available.<<

patrick-laa: Ok, I'm happy to accept the govuk official line on this.

patrick-laa (approved):

PR #155, opened 2022-11-03
EL-388: Complex check answers behaviour

starswan:
https://dsdmoj.atlassian.net/browse/EL-388

More complex check answers behaviour. If check answers screen results in the need to answer other questions, those questions are now answered in turn so that by the time the user returns to the check answers screen the data set is valid again


willc-work (approved):

patrick-laa (approved):

patrick-laa:
Oops, didn't spot that Will had already approved this. Well, you're doubly approved-of, then.

PR #162, opened 2022-11-04
add github action to delete UAT release on merge or close of PR

willc-work:
Add a github action to delete UAT releases when a PR is merged with main or closed.
Add secrets to the repo via the cloud platform namespace - https://github.com/ministryofjustice/cloud-platform-environments/pull/9101

The circleci job does not delete UAT releases when a PR is opened and subsequently closed without merging. This job will be removed in a subsequent PR

Checklist

### DIFF
--- a/.github/actions/delete-uat-release/action.yml
+++ b/.github/actions/delete-uat-release/action.yml
@@ -1,0 +1,79 @@
+name: "Delete UAT deployment"
+description: 'Deletes a UAT deployment with a name that matches the merged or closed branch'
+inputs:
+  k8s_cluster:
+    description: "Kubernetes cluster name"
+    required: true
+  k8s_cluster_cert:
+    description: "Kubernetes cluster certificate"
+    required: true
+  k8s_namespace:
+    description: "Kubernetes cluster namespace"
+    required: true
+  k8s_token:
+    description: "Kubernetes authentication token"
+    required: true
+outputs:
+  branch-name:
+    description: "Extracted branch name"
+    value: ${{ steps.extract_branch.outputs.branch }}
+  release-name:
+    description: "Extracted release name"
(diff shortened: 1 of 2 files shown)


.github/actions/delete-uat-release/action.yml
```diff
@@ -0,0 +1,79 @@
...
+        fi
+
+        echo "##[set-output name=branch;]$(echo $branch)"
+
+    - name: Extract release name
+      id: extract_release
+      shell: bash
+      run: |
+        branch=${{ steps.extract_branch.outputs.branch }}
+        truncated_branch=$(echo $branch | tr '[:upper:]' '[:lower:]' | sed 's:^\w*\/::' | tr -s ' _/[]().' '-' | cut -c1-20 | sed 's/-$//')
```
patrick-laa, line 48: Question: If we had two git branches with long names, where the first part of the name was the same (e.g. `el-500-do-some-really-complicated-thing-part-one` and `el-500-do-some-really-complicated-thing-part-two`), would this run the risk of picking the wrong release? Or would the fact the truncated names were identical mean this would have blown up when trying to create the release in the first place?

(I'm terrible at reading and understanding scripts like this, so apologies if this is a stupid question)

willc-work: In that situation the second option applies, i.e the deployment fails due to the similarity of the names. We can extend it from 20 chars but we end up hitting a limit on the ingress name which has to be made up from the release_name+repo_name+namespace. So at best we may get another 6 characters in the release name.

patrick-laa: Ok cool, so it's not something we need to worry about here. Thanks for explaining!

patrick-laa (approved):

PR #163, opened 2022-11-04
EL-419 Subject matter of dispute.

starswan:
https://dsdmoj.atlassian.net/browse/EL-419

This branch/PR is the main point of entry for the Subject matter of dispute work. All SMOD work will be merged here after review and this branch will then continue as the UAT point of reference to avoid branch merge/rebase issues


willc-work (approved):

willc-work:
I seem to have reviewed this in error and I cannot dismiss my review

app/services/check_answers/section_lister_service.rb
```diff
@@ -58,9 +58,20 @@ def build_field(field_data, label_set, parent_screen)
...
+                     @session_data[field_data[:disputed_attribute]]
+                   end
+                 else
+                   false
+                 end
+
       Field.new(label: "#{label_set}_fields.#{label}",
                 type: field_data[:type],
                 value:,
+                disputed?: disputed && value.present?,
```
patrick-laa, line 74: When the `value` variable is set to the property `property_owned` attribute, and it is set to `"none"`, the `value.present?` will still be truthy, leading to the potential for:
Screenshot 2022-11-09 at 16 05 19

app/lib/check_answers_fields.yml
```diff
@@ -108,55 +108,57 @@ sections:
...
       - label: property
+        screen: property
         fields:
           - type: select
             attribute: property_owned
-            screen: property
+            disputed_attribute: house_in_dispute
           - type: money
             attribute: house_value
-            screen: property_entry
```
patrick-laa, line 117: An unfortunate side effect of these `screen` removals is the presenter no longer understands that if the `property_entry` screen is to be skipped (i.e. the user has said they do not own their house), these fields should always be skipped. Leading to:
Screenshot 2022-11-09 at 16 06 45

patrick-laa: (this is my own damn fault for making these fields 'clever'. The solution might be to leave the link in between fields and screens in the yml file, for the purposes of knowing which fields to omit based on which screens to omit. But then _not_ use the field's `screen` attribute to construct a link.

app/services/check_answers/relevant_section_finder_service.rb
```diff
@@ -22,9 +22,7 @@ def call
         end
       end
 
-      dictionary = nested_pairs.flatten.map { [_1[:key], _1[:value]] }.to_h
-
-      dictionary.fetch(@step.to_s)
+      nested_pairs.flatten.detect { _1[:key] == @step }&.fetch(:value)
```
patrick-laa, line 25: Some sort of small regression has been added here: Previously if you went: CheckAnswersScreen->VehicleScreen->VehicleDetailsScreen->CheckAnswerScreen, the final redirect would add an `#assets-section` anchor to the URL. Now when I try it the anchor is `#-section` and therefore broken.

Note that  CheckAnswersScreen->VehicleScreen->CheckAnswerScreen _does_ still end up with the correct anchor, so it's something to do with multi-step check answer loops.

starswan: good spot, but sadly wasn't caused by this specific change. Rework to come

patrick-laa (approved):

PR #184, opened 2022-11-11
EL-468: New partner input screens

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-468)

- Replace 'Handler' logic with a couple of `SessionPersistable` concerns that allow form object to instantiate themselves with session data
- Add new 'normal' partner screens to the flow, reusing existing form objects and form partials.
- Add new partner benefits controller and flow

 
Checklist


app/controllers/flow/generic_handler.rb
```diff
@@ -0,0 +1,22 @@
+module Flow
+  class GenericHandler
+    def initialize(form_class)
+      @form_class = form_class
+    end
+
+    def model_from_session(session_data)
+      @form_class.new(session_data.slice(*@form_class::ATTRIBUTES.map(&:to_s))).tap { modify(_1, session_data) }
+    end
```
starswan, line 9: In EL-419, I used a #from_session method to allow classes to override this load. Might be easier to follow than this version, WDYT? Saves a blank 'modify' part-constructor - kind of a factory pattern?

app/controllers/flow/generic_handler.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    def model_from_params(params, session_data)
+      relevant_params = params.fetch(@form_class.name.underscore, {}).permit(*@form_class::ATTRIBUTES)
+      @form_class.new(relevant_params).tap { modify(_1, session_data) }
+    end
```
starswan, line 14: same here - could be model.from_params maybe?

app/controllers/flow/partner_handler.rb
```diff
@@ -1,13 +1,33 @@
...
+    end
+
+    # Ensure that the model's class name starts with 'Partner' so that it gets its own
+    # key for validation error i18n.
+    def partner_form_class
+      if Object.const_defined?(partner_form_class_name)
+        partner_form_class_name.constantize
+      else
+        Object.const_set(partner_form_class_name, Class.new(@form_class))
       end
```
starswan, line 26: frankly I don't understand this method. Why do we need it? It looks very low-level. I would have expected partner_form_class_name.constantize to always work...?

app/forms/vehicle_details_form.rb
```diff
@@ -1,29 +1,3 @@
...
-  attribute :vehicle_value, :gbp
-  validates :vehicle_value,
-            numericality: { greater_than: 0, allow_nil: true },
-            presence: true
-
-  attribute :vehicle_in_regular_use, :boolean
-  validates :vehicle_in_regular_use, inclusion: { in: [true, false], allow_nil: false }
+class VehicleDetailsForm < BaseVehicleDetailsForm
+  # TODO: Put SMOD attribute and validation here
 end
```
starswan, line 3: should these be called 'ClientVehicleXXX' to distinguish from 'PartnerVehicleXXX' as they are quite different (once SMOD is in)

app/helpers/application_helper.rb
```diff
@@ -40,9 +40,10 @@ def number_to_money(number)
...
   end
 
-  def variable_controller_benefit_path(action: nil, id: nil)
+  def variable_controller_benefit_path(action: nil, id: nil, partner: false)
     prefix = "#{action}_" if action
-    controller = controller_name == "check_answers" ? "check_benefits_answer" : "benefit"
+    main_name = partner ? "partner_benefit" : "benefit"
+    controller = controller_name == "check_answers" ? "check_#{main_name}s_answer" : main_name
 
     if id
```
starswan, line 48: This now looks *really* complicated. Can we look to refactor it? WDYT?

app/models/partner_benefit_model.rb
```diff
@@ -0,0 +1 @@
+class PartnerBenefitModel < BenefitModel; end
```
starswan, line 1: Is there some way of avoiding this class needing to exist? If there is no difference in behaviour..?

app/views/estimate_flow/assets.html.slim
```diff
@@ -6,63 +6,7 @@
...
-      .govuk-hint = t(".valuables.hint")
-
-      = render "shared/money_input",
-               form:,
-               field: :valuables,
-               width: 5,
-               label_text: t(".valuables.input")
-
-    = form.govuk_submit t("generic.save_and_continue")
+  = render "estimate_flow/forms/assets", i18n_key: "assets", show_smod_content: true
```
starswan, line 9: I'm a bit anti 'passing booleans into functions to change their behaviour' - is there a simpler way of breaking up the assets form so that we know that we're in partner-land (or not) to avoid this boolean?

app/controllers/flow/generic_handler.rb
patrick-laa: I have taken your suggestion and added `from_session` methods to Form objects via some new concerns.

app/controllers/flow/generic_handler.rb
patrick-laa: Done

app/forms/vehicle_details_form.rb
patrick-laa: Done

app/helpers/application_helper.rb
patrick-laa: Moved into benefits_helper and split up into some simpler parts.

app/models/partner_benefit_model.rb
patrick-laa: As discussed, the benefit of this class is it having its own i18n key for error messages. I have added a comment to this file to that effect.

app/views/estimate_flow/assets.html.slim
patrick-laa: As discussed, no obvious better solution has emerged. This may be the lesser of multiple evils.

starswan (approved):

PR #199, opened 2022-11-16
Re-add conditional logic around partner field display

patrick-laa:
The SMOD merge accidentally removed some work done after it to show partner employment and age summaries on the check answers screen. This re-adds that work (doing it differently this time to use the new `Section` object) and adds a test to make sure it can't go missing again.


exonian (approved):
Thanks @patrick-laa for explaining to me that the previous was memoization, so I could understand the syntax changing as well as the value being removed 👍 

Test looks good

PR #201, opened 2022-11-16
EL-483: Add flipper

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-483)

- Create a `partner` feature flag using Flipper, reading flag settings from environment variables
- Only show the 'do you have a partner?' screen if Flipper is true (if this screen is not shown, all additional partner screens will be omitted from the flow)
- Set the flag to enabled on UAT and disabled on staging


willc-work (approved):
👍 looks clear and easy to use

PR #204, opened 2022-11-17
EL-477 add employed attribute to applicant

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-477)

CFE expected an employed attribute for applicant as this was being used to determine a client's eligibility for childcare cost allowance in the method `eligible_for_childcare_costs?` in `childcare_collator.rb`

This adds the correct attribute

An employed applicant OR one who is in receipt of student finance, and has a dependant child is eligible for childcare costs to be discounted from eligibility (I think - need Mike to confirm that this is actually happening in UAT as I need clarity on what figures are expected)

Checklist


app/lib/cfe_connection.rb
```diff
@@ -25,11 +25,12 @@ def create_proceeding_type(assessment_id, proceeding_type)
     create_record(assessment_id, "proceeding_types", proceeding_types:)
   end
 
-  def create_applicant(assessment_id, date_of_birth:, receives_qualifying_benefit:)
+  def create_applicant(assessment_id, date_of_birth:, receives_qualifying_benefit:, employed:)
     applicant = {
```
patrick-laa, line 29: Over the past fortnight we've very slowly been moving logic out of CfeConnection into the Submit services that call it. `create_applicant` is one of the very few to be left. As part of this ticket, any chance you could move the creation of the applicant hash into SubmitApplicantService, and then pass it as a param to `create_applicant`? That would help us on our slow quest towards consistency.

patrick-laa (approved):

PR #213, opened 2022-11-21
EL-471: Partner results page

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-471)

Adjust results page to show partner fields where relevant

Checklist


app/models/calculation_result.rb
```diff
@@ -30,30 +30,58 @@ def income_contribution
   end
 
   def gross_income
-    monetise(api_response.dig(:result_summary, :gross_income, :total_gross_income))
+    sum = if has_partner?
+            api_response.dig(:result_summary, :gross_income, :total_gross_income) +
```
willc-work, line 34: I may be missing something... Isnt this calculation being done somewhere in CFEP, i.e combining the partner and client gross_income's to give us an overall value which we apply against the thresholds (same for disposable and capital below). If so, would it not make more sense to return that value from CFEP instead of piecing it together here?

patrick-laa: @willc-work you're right. At the moment CFEP adds the two together for the purpose of comparing to the thresholds, but it doesn't store that sum anywhere in the DB, and what CFEP returns to EFE is basically a representation of the database. But it would make a lot more sense for CFEP to tell us the combined total to avoid having to do maths here.

I'll work up a CFEP PR.

app/models/calculation_result.rb
```diff
@@ -62,143 +66,205 @@ def assessed_capital
...
+    capital_subtotal_rows(prefix: "partner_")
   end
 
 private
 
   attr_reader :api_response
 
-  def capital_items(key)
-    api_response.dig(:assessment, :capital, :capital_items, key)
+  def capital_items(key, prefix = nil)
```
starswan, line 158: I think I would prefer '' over nil here to show that the prefix is empty rather than missing. WDYT?

app/models/calculation_result.rb
```diff
@@ -62,143 +66,205 @@ def assessed_capital
...
+    }
+
+    data.transform_values { |value| monetise(value) }
+  end
+
+  def vehicle_disregards(prefix)
+    value = capital_items(:vehicles, prefix).sum(0) { |z| z.fetch(:value) }
+    pcp = capital_items(:vehicles, prefix).sum(0) { |z| z.fetch(:loan_amount_outstanding) }
+    assessed_value = capital_items(:vehicles, prefix).sum(0) { |z| z.fetch(:assessed_value) }
+    value - pcp - assessed_value
```
starswan, line 268: should we be doing this calculation here, or should CFE be doing it?

spec/features/partner_assets_page_spec.rb
```diff
@@ -3,11 +3,7 @@
 RSpec.describe "Partner assets page" do
   let(:partner_assets_heading) { I18n.t("estimate_flow.partner_assets.assets.legend") }
 
-  around do |example|
-    Flipper.enable(:partner)
-    example.run
-    Flipper.disable(:partner)
-  end
+  enable_partner_screens
```
starswan, line 6: There is a more idiomatic version of this in EL-472 - I think I will break it out into a separate PR

patrick-laa: Yes please, that looks great!

app/models/calculation_result.rb
patrick-laa: Sounds sensible, will do.

app/models/calculation_result.rb
patrick-laa: You're right, CFE should, and now that we're being more hands-on with CFE, it's within our ability to make that happen. I'll get on it.

patrick-laa:
As we're now using a lot of functionality that's only available in our fork, I've updated the default CFE URL to our fork, and regenerated all our VCR files accordingly, hence the large number of files here.

starswan (approved):

PR #217, opened 2022-11-23
EL-379 Change service name

starswan:
https://dsdmoj.atlassian.net/browse/EL-379

Change service name to reflect new name

### DIFF
--- a/app/views/start/index.html.slim
+++ b/app/views/start/index.html.slim
@@ -1,5 +1,5 @@
 .govuk-grid-column-two-thirds
-  h1 class="govuk-heading-xl" = t(".efe_heading")
+  h1 class="govuk-heading-xl" = t("service.name")
 
   p class="govuk-body" = t(".legal_aid_provider")
 
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -239,7 +239,7 @@ en:
             vehicle_finance:
               blank: Please enter the outstanding finance amount
   service:
-    name: Estimate financial eligibility for legal aid
+    name: Check if your client qualifies for legal aid
     email: eligibility@justice.gov.uk
     beta: Beta
     beta_message_html: This is a new service - your <a href="mailto:%{email}">feedback</a> will help us improve it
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -814,7 +814,7 @@ en:
       legal_aid: Legal Aid
       agency: Agency
       print_this_page: Print this page
-      check_eligibility_for_legal_aid: Check Eligibility For Legal Aid
+      check_eligibility_for_legal_aid: Check if your client qualifies for legal aid
```
patrick-laa, line 817: I'd also update or remove en.cookies.show.service_name

app/views/start/index.html.slim
```diff
@@ -1,5 +1,5 @@
 .govuk-grid-column-two-thirds
-  h1 class="govuk-heading-xl" = t(".efe_heading")
+  h1 class="govuk-heading-xl" = t("service.name")
```
patrick-laa, line 2: The ticket also says we need to update the header so that `gov.uk` on the left links to gov.uk.

starswan: I know - but the standard doesn't seem to have it and the standard component (which we're using) doesn't produce it. Will be triple-checking with Jo

patrick-laa: You can make the standard component do what you want by passing an explicit `homepage_url` _and_  `service_url` I believe: https://github.com/DFE-Digital/govuk-components/blob/main/app/components/govuk_component/header_component.rb#L22

starswan: Yes - but there's research that suggests it's confusing for users to take them back to gov.uk

davidread: ooh, interesting that research is point in this new direction - I've always found it confusing that link to gov.uk. However since this link is not a core part of our service, I suggest we don't overthink it. On this basis, how about we simply defer to the convention of linking to gov.uk?

starswan: @davidread yes I've gone with that in the end rather than digging the hole any deeper

davidread: 👍 I'll leave others to do the approval

config/locales/en.yml
davidread: Looks done in d3ecab2

davidread (approved):
Looks good on UAT. Simple enough code, so I'll approve, to get this significant change in 🚀

PR #220, opened 2022-11-24
EL-494 separate client and partner questions

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-XXX)

Move domestic abuse (proceeding type) and partner questions into the client details screen.
Add partner details page to capture over+60 and employed attributes for partner
update flow to use new partner page when there is a partner
Update all specs
Update check answers

Checklist


app/forms/applicant_form.rb
```diff
@@ -3,32 +3,18 @@ class ApplicantForm
   include ActiveModel::Attributes
   include SessionPersistable
 
-  PERMANENT_ATTRIBUTES = %i[passporting over_60 employed].freeze
-  CONTINGENT_ATTRIBUTES = %i[partner_over_60 partner_employed].freeze
+  PROCEEDING_TYPES = { domestic_abuse: "DA001", other: "SE003" }.freeze
 
-  ATTRIBUTES = PERMANENT_ATTRIBUTES + CONTINGENT_ATTRIBUTES.freeze
+  PROCEEDING_ATTRIBUTE = %i[proceeding_type].freeze
+  PERMANENT_ATTRIBUTES = %i[passporting over_60 employed partner].freeze
```
patrick-laa, line 9: I don't think `PERMANENT_ATTRIBUTES` is the right name here - all attributes listed are permanent. What sets these ones apart now is that they are _boolean_.

app/forms/applicant_form.rb
```diff
@@ -3,32 +3,18 @@ class ApplicantForm
...
 
-  attr_accessor :partner
+  ATTRIBUTES = PERMANENT_ATTRIBUTES + PROCEEDING_ATTRIBUTE.freeze
 
   PERMANENT_ATTRIBUTES.each do |attr|
     attribute attr, :boolean
     validates attr, inclusion: { in: [true, false] }
```
patrick-laa, line 15: I don't think we want this validation to be applied to the partner attribute if the Partner feature flag is switched off.

app/forms/partner_details_form.rb
```diff
@@ -1,9 +1,9 @@
-class PartnerForm
+class PartnerDetailsForm
   include ActiveModel::Model
   include ActiveModel::Attributes
   include SessionPersistable
 
-  ATTRIBUTES = %i[partner].freeze
+  ATTRIBUTES = %i[partner_over_60 partner_employed dependants].freeze
```
patrick-laa, line 6: Careful - because this model includes `SessionPersistable`, not `SessionPersistableForPartner`, the dependants attribute here will be saved to the session as `dependants`, overwriting the _client_ dependants attribute.

I would recommend including the partner persistable concern, so this line then becomes `ATTRIBUTES = %i[over_60 employed dependants]` and we know that `partner_` will get prepended to the relevant session. keys.

app/lib/applicant_case_details_section.rb
```diff
@@ -1,13 +1,11 @@
 class ApplicantCaseDetailsSection
   class << self
     def all_steps
-      %i[case_details partner applicant dependants dependant_details]
+      %i[applicant dependants dependant_details]
```
patrick-laa, line 4: I thought it was agreed that we would remove the dependants screen and ask the dependants question on the applicant screen?

app/lib/check_answers_fields.yml
```diff
@@ -39,19 +39,12 @@ sections:
         fields:
         - attribute: partner
```
patrick-laa, line 40: We will need to make sure that this attribute gets hidden if the partner feature flag gets switched off. (I think it's going to be a case of adjusting  `CheckAnswers::ApplicantSection` and in particular `#display_fields`)

app/services/check_answers/partner_section.rb
```diff
@@ -0,0 +1,28 @@
+module CheckAnswers
```
patrick-laa, line 1: I don' t think you actually need to add this model. The "vanilla" logic for Check Answers will ensure that the fields will only get displayed if the `partner_details` screen is part of the flow, and the `disputed_asset?` logic doesn't apply.

Basically we only need these section models for Check Answers if there's SMOD logic or if there's complex logic around when to display particular fields within a section.

app/views/estimate_flow/applicant.html.slim
```diff
@@ -8,23 +8,30 @@
...
+                                          :last,
+                                          ->(pair) { t(".proceeding_type.#{pair[0]}.label") },
+                                          legend: { text: t(".proceeding_type.legend") }
+
     = form.govuk_collection_radio_buttons :over_60,
                                           yes_no_options, :first, :last,
                                           legend: { text: t(".over_60.legend") },
                                           hint: { text: t(".over_60.hint") }
+
+    = form.govuk_collection_radio_buttons :partner,
```
patrick-laa, line 22: This will need to be wrapped in the feature flag.

spec/features/partner_employment_page_spec.rb
```diff
@@ -30,21 +30,21 @@
...
 
     it "shows the partner employment page" do
       expect(page).to have_content(partner_employment_page_header)
     end
 
-    it "has a back link to the assets page" do
-      click_link "Back"
-      expect(page).to have_content assets_page_header
-    end
+    # it "has a back link to the assets page" do
```
patrick-laa, line 44: I'm basically allergic to commented-out code. If the test works, it shouldn't be commented out. If it doesn't but should, let's fix it. If it doesn't work and shouldn't work, let's remove it.

patrick-laa (changes requested):

spec/features/partner_employment_page_spec.rb
willc-work: I meant to go back and fix this and couldn't find it. Updated the test and reinstated it

app/lib/applicant_case_details_section.rb
willc-work: Ah the DoD on the ticket and the acceptance criteria are different on the issue of dependants question.

app/controllers/flow/handler.rb
```diff
@@ -1,8 +1,6 @@
 module Flow
   class Handler
     CLASSES = {
-      case_details: ProceedingTypeForm,
-      partner: PartnerForm,
       applicant: ApplicantForm,
       dependants: DependantsForm,
```
patrick-laa, line 5: This line can go, as well as the DependantsForm itself.

app/services/check_answers/applicant_section.rb
```diff
@@ -30,11 +30,7 @@ def from_session(session)
     end
 
     def display_fields
-      if partner
-        FIELDS
-      else
-        FIELDS - %i[partner_employed partner_over_60]
```
patrick-laa, line 36: You should be able to delete the partner_employed and partner_over_60 attributes on lines 23 and 24 here, because they're no longer things that appear in the applicant section of check answers

patrick-laa (approved):
There are a couple of extra things you can delete, but everything looks good to me.

PR #225, opened 2022-11-25
EL-504 client content

starswan:
https://dsdmoj.atlassian.net/browse/EL-504

Add content from content designer - client (non-partner) only for this pass

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -285,22 +285,31 @@ en:
     smod_help_text: Regulation 38 (Civil Legal Aid (Financial Resources and Payment for Services) Regulations 2013) provides for a disregard of the amount or value of assets that are the subject matter of dispute (SMOD); however the total amount disregarded must not exceed £100,000. Please read the related content on subject matter of dispute (Regulation 38).
   estimate_flow:
     partner:
+      partner_guidance:
+        text: "Guidance on partners"
+        link: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1045682/Lord_Chancellor_s_guidance_on.pdf#page=17
       caption: About your client
       legend: Does your client have a partner?
       hint: This is a spouse, someone with whom they are in a civil relationship, a cohabiter or someone they share their finances with. They cannot be the opponent in any legal proceedings.
     applicant:
+      passporting_guidance:
+        text: Guidance on passporting
+        link: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1045682/Lord_Chancellor_s_guidance_on.pdf#page=11
+      pensioner_guidance:
+        text: "Guidance on Pensioner disregards"
+        link: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1045682/Lord_Chancellor_s_guidance_on.pdf#page=71
       heading: Tell us about your client
       heading_with_partner: Tell us about your client and their partner
       caption: About your client
       over_60:
-        legend: Is your client over 60 years old?
+        legend: Is your client 60 or over?
         hint: Clients over 60 may have some of their capital excluded from a financial eligibility assessment.
       employed:
         legend: What is your client's employment status?
         hint: ""
       passporting:
         legend: Does your client receive a passporting benefit?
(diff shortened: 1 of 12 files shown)


config/locales/en.yml
```diff
@@ -892,19 +917,20 @@ en:
...
-        - "You'll need information from your client (and any partner) about their:"
+        - "You'll need information about your client's (and their partner, if applicable.):"
       required_information_types:
-        - income, including any benefits they claim
-        - outgoings
-        - savings and investments
-        - assets, like property and vehicles
+        - "Income including any salary, benefits and financial support received"
+        - "Outgoings such as housing costs, childcare costs, maintenance and any criminal legal aid contribution payments your client may be making"
+        - savings, investments and other assets
```
patrick-laa, line 924: Grammar: Either all these bullet points should start with a capital, or none should.

config/locales/en.yml
```diff
@@ -892,19 +917,20 @@ en:
       before_start_heading: Before you start
       before_start_paragraphs:
         - It should take about 15 minutes to answer all the questions.
-        - "You'll need information from your client (and any partner) about their:"
+        - "You'll need information about your client's (and their partner, if applicable.):"
```
patrick-laa, line 920: The sentence construction here is a bit tortured, but if we are going to do it like this, `partner` should be `partner's`, and there shouldn't be a full stop in side the parenthesis. I'd recommend going back to Chris and asking exactly how this sentence should read, because I don't think this one has been completely thought out.

config/locales/en.yml
```diff
@@ -305,18 +314,18 @@ en:
...
-        hint: For example, your client's child, grandchild, niece or nephew.
+        hint: For example your client’s children or another child relative who lives in their household. Do not include any foster children.
         instruction: Enter number of child dependants
     employment:
       caption: Your client's finances
       heading: Add your client's salary breakdown
       second_caption: You can find this on your client's payslips.
       gross_income:
-        input: Employment income
+        input: Gross Employment Income
```
patrick-laa, line 328: I know the Google doc has these all capitalised, but I think that may not have been in the conscious awareness that that would be a convention-breaking change. I am 99% certain that Chris would confirm that, for consistency, this should be 'Gross employment income' if asked.

config/locales/en.yml
```diff
@@ -295,7 +304,7 @@ en:
       hint: "A dependant is a child or adult who:"
       lives_with: lives with your client
       depends: depends on your client for financial support
-      no_partner: Don't include your client's partner if they have one
+      no_partner: Do not include your client’s partner as a dependant.
```
patrick-laa, line 307: Is the intention here to only show this if the client has a partner? (Because without the 'if' clause, this sentence looks strongly like it doesn't belong when there is no partner.)

config/locales/en.yml
```diff
@@ -267,22 +267,31 @@ en:
...
     partner:
+      partner_guidance:
+        text: "Guidance on partners"
+        link: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/1045682/Lord_Chancellor_s_guidance_on.pdf#page=17
       caption: About your client
       legend: Does your client have a partner?
       hint: This is a spouse, someone with whom they are in a civil relationship, a cohabiter or someone they share their finances with. They cannot be the opponent in any legal proceedings.
     applicant:
+      passporting_guidance:
+        text: "Guidance on Passporting"
```
patrick-laa, line 278: With this and the pensioner guidance again I would be surprised if adding a capital P in the middle was a deliberate, convention-breaking choice. I'd confirm with Chris.

patrick-laa (commented):
A bunch of small grammar things that, while they're what's literally specified, are things I'm pretty sure are not intended. I've asked Chris on the Google doc in each case.

config/locales/en.yml
starswan: Yes this is only shown if the client has a partner IIRC

patrick-laa: Looking at the view file, it doesn't have any conditional logic in it yet - because previously the "if" in the wording meant we could just about get away without it. Now that we don't have the "if" any more, I think we need to add in the conditional logic.

starswan: I think the challenge here is that I'm working on 2 tickets with dependants. The other one does have 'if' logic - I've updated the content here for now pending the dependants ticket that is being reworked anyway

patrick-laa: Ohhh, I see. Fair enough!

patrick-laa (approved):

PR #232, opened 2022-11-28
EL-513 don't show partner_other_income section when no partner

starswan:
https://dsdmoj.atlassian.net/browse/EL-513

Remove partner_benefits section when no partner so that partner_other_income section collapses to nothing

### DIFF
--- a/app/services/check_answers/section_lister_service.rb
+++ b/app/services/check_answers/section_lister_service.rb
@@ -80,25 +80,35 @@ module CheckAnswers
                 alt_value: @session_data[field_data[:alt_attribute]])
     end
 
-    def benefits_fields(session_key: "benefits", field_type: "benefit")
-      return [] unless StepsHelper.valid_step?(@model, :benefits)
-
-      if @session_data[session_key].blank?
-        return [Field.new(label: I18n.t("generic.not_applicable"),
-                          type: field_type)]
-      end
-
-      @session_data[session_key].map do |benefit|
-        Field.new(label: benefit["benefit_type"],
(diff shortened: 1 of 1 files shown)


app/services/check_answers/section_lister_service.rb
```diff
@@ -98,7 +98,11 @@ def benefits_fields(session_key: "benefits", field_type: "benefit")
     end
 
     def partner_benefits_fields
-      benefits_fields(session_key: "partner_benefits", field_type: "benefit")
+      if StepsHelper.valid_step?(@model, :partner_benefits)
```
patrick-laa, line 101: It looks to me like the buggy line is really line 84, in `#benefits_fields` which makes the assumption that the relevant step is `:benefits`, when in fact, when called from `#partner_benefits_fields`, the relevant step is `:partner_benefits`.

If we don't fix that buggy line then if there's ever a situation (I know, I know, it's unlikely) where `:benefits` is an _invalid_ step but `:partner_benefits` is a valid one, line 84 will bite us again, because we'll _not_ show the partner benefits section even when we should.

So I reckon the 'best' way to fix this bug is to fix the bad assumption in line 84, rather than try to work around it.

patrick-laa: Oh, sorry, I jumped the gun, I didn't see this PR was only a draft. Sorry!

starswan: review appreciated anyway!

patrick-laa (approved):

PR #236, opened 2022-11-29
EL-495 add partner dependants section into journey

starswan:
https://dsdmoj.atlassian.net/browse/EL-494

Add partner dependants screens to journey. 
Add partner dependants to CFE partner endpoint

### DIFF
--- a/app/lib/check_answers_fields.yml
+++ b/app/lib/check_answers_fields.yml
@@ -32,10 +32,11 @@
 sections:
   - label: case_details
     subsections:
-  - label: client_details
-    screen: applicant
+  - label: about_your_client
     subsections:
-      - model: ApplicantSection
+      - label: client_details
+        screen: applicant
+        model: ApplicantSection
         fields:
         - attribute: partner
           type: boolean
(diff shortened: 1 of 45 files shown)


app/services/check_answers/applicant_section.rb
```diff
@@ -18,6 +23,11 @@ class ApplicantSection
     attribute :dependants, :boolean
     attribute :child_dependants, :integer
     attribute :adult_dependants, :integer
+    attribute :partner_employed, :boolean
```
patrick-laa, line 26: I'm not sure why we're adding these attributes here? None of these fields belong in the applicant section of check answers, so I don't think they should belong in CheckAnswers::ApplicantSection.

patrick-laa (approved):

app/lib/check_answers_fields.yml
```diff
@@ -45,13 +45,16 @@ sections:
           type: boolean
         - attribute: passporting
           type: boolean
-        - attribute: dependants
-          type: boolean
-          screen: dependants
         - attribute: child_dependants
+          type: boolean
+          screen: dependant_details
+        - attribute: child_dependants_count
```
patrick-laa, line 51: If I say I have 2 child dependants, then go back and change my answer about child dependants from 'Yes' to 'No', the check answers screen will say "**Has child dependants** No"/"**Number of child dependants** 2" which seems wrong. I'd suggest adding a `skip_unless: child_dependants` to this attribute (and similar to the other three `_dependants_count` attributes.

starswan: It was done for partner version, just not for client version

app/lib/check_answers_fields.yml
```diff
@@ -45,13 +45,16 @@ sections:
           type: boolean
         - attribute: passporting
           type: boolean
-        - attribute: dependants
-          type: boolean
-          screen: dependants
         - attribute: child_dependants
```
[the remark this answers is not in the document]
starswan: Spoken to Kate - she agreed so have implemented

PR #240, opened 2022-11-30
EL-517: Partner housing benefit

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-517)

Added the partner equivalent of the new housing benefit screens

### DIFF
--- a/app/lib/partner_income_section.rb
+++ b/app/lib/partner_income_section.rb
@@ -1,16 +1,27 @@
 class PartnerIncomeSection
   class << self
     def all_steps
-      %i[partner_employment partner_benefits partner_other_income partner_outgoings]
+      %i[partner_employment partner_housing_benefit partner_housing_benefit_details partner_benefits partner_other_income partner_outgoings]
     end
 
     def steps_for(estimate)
       if estimate.passporting || !estimate.partner
         []
       else
-        employment_step = estimate.partner_employed ? [:partner_employment] : []
-        (employment_step + %i[partner_benefits partner_other_income partner_outgoings]).map { [_1] }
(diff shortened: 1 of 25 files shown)


app/lib/partner_income_section.rb
```diff
@@ -1,16 +1,27 @@
...
     def steps_for(estimate)
       if estimate.passporting || !estimate.partner
         []
       else
-        employment_step = estimate.partner_employed ? [:partner_employment] : []
-        (employment_step + %i[partner_benefits partner_other_income partner_outgoings]).map { [_1] }
+        employment_steps(estimate) + housing_benefit_steps(estimate) + other_steps
       end
     end
+
```
starswan, line 14: could we lob a 'private' in here to indicate where the public interface stops?

app/services/submit_partner_service.rb
```diff
@@ -44,10 +44,15 @@ def regular_transactions
   end
 
   def state_benefits
-    form = PartnerBenefitsForm.from_session(@cfe_session_data)
-    return [] if form.benefits.blank?
+    benefits_form = PartnerBenefitsForm.from_session(@cfe_session_data)
+    housing_benefit_form = PartnerHousingBenefitForm.from_session(@cfe_session_data)
+    return [] if benefits_form.benefits.blank? && !housing_benefit_form.housing_benefit
 
-    CfeParamBuilders::StateBenefits.call(form)
```
starswan, line 50: This change suggests that the StateBenefits param builder has an optional second parameter that maybe isn't optional any more?

starswan (approved):

PR #241, opened 2022-11-30
EL-514 hide partner

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-514)

Hid (with the feature flag) or removed all references to partner in the non-partner journey where they did not jive with the journey. Ran through the journey twice using CTRL+F to find all mentions of partner and confirmed that in the translation file any remaining uses were hidden behind a feature flag e.g partner_details pages or partner_dependants pages

Also includes the start page changes detsiled on el-520 to remove warning text, add bullet point etc

There is probably a better way to hide the inset text on the dependants question on applicant.html.slim, I tried using Flipper within the lambda but it complained about the formatting, which may be due to me trying to implement it incorrectly. I tried using a ternary and could remove or show the text but removing the class caused it to fail.

Checklist


app/services/check_answers/property_section.rb
```diff
@@ -21,7 +21,7 @@ def from_session(session_data)
 
     def display_fields
       if owns_property?
-        FIELDS
+        Flipper.enabled?(:partner) ? FIELDS : FIELDS - %i[joint_ownership joint_percentage_owned]
```
patrick-laa, line 24: This is a good spot, but I think it's not enough to check whether the partner flag is enabled, because we also don't want to show these fields if partner flag is enabled _but the client doesn't have a partner_. 

So I think we need to add a `partner` attribute to this model, and make sure that it gets assigned in `#from_session` (maybe `new session_data.slice(*FIELDS + [:partner])`?).

Then this line becomes `partner ? FIELDS : FIELDS - %i[joint_ownership joint_percentage_owned]`

(If the feature flag is not enabled, `partner` will always be false-y).

app/views/estimate_flow/applicant.html.slim
```diff
@@ -18,35 +18,48 @@
...
-            hint: lambda { \
-                            tag.p(t(".dependants.hint")) + \
-                            tag.ul(class: "govuk-list govuk-list--bullet") do \
-                            tag.li(t(".dependants.lives_with")) + \
-                            tag.li(t(".dependants.depends")) \
-                            end + \
-                            tag.div(t(".dependants.no_partner"), class: "govuk-inset-text") \
-                            }
+    - if Flipper.enabled?(:partner)
+      = form.govuk_collection_radio_buttons :dependants,
```
patrick-laa, line 22: Just to warn you that in https://github.com/ministryofjustice/laa-estimate-financial-eligibility-for-legal-aid/pull/236 this field is moving to a different screen, so brace yourself for a merge conflict if that gets merged first.

app/views/estimate_flow/applicant.html.slim
```diff
@@ -18,35 +18,48 @@
...
-            legend: { text: t(".dependants.legend") },
-            hint: lambda { \
-                            tag.p(t(".dependants.hint")) + \
-                            tag.ul(class: "govuk-list govuk-list--bullet") do \
-                            tag.li(t(".dependants.lives_with")) + \
-                            tag.li(t(".dependants.depends")) \
-                            end + \
-                            tag.div(t(".dependants.no_partner"), class: "govuk-inset-text") \
-                            }
+    - if Flipper.enabled?(:partner)
```
patrick-laa, line 21: Given how gross the hint lambdas are, I think this is as good a way of changing the hint text depending on the feature flag as any. :+1:

patrick-laa (commented):
One suggested change for CheckAnswers::PropertySection, everything else looks good.

app/services/check_answers/property_section.rb
willc-work: thanks, this is much better and is one less place we need to remove Flipper at a later date

app/views/estimate_flow/applicant.html.slim
willc-work: 👍 🏎️

patrick-laa (approved):

PR #258, opened 2022-12-05
EL-527: Update content

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-527)

* Update content
* Remove leading zero from dates

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -623,7 +623,7 @@ en:
         none: "No"
         legend: Does your client and/or their partner own the home they live in?
       trapped_capital:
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -614,7 +614,7 @@ en:
         none: "No"
         legend: Does your client and/or their partner own the home they live in?
       trapped_capital:
-        text: "If your client’s capital includes assets that may be inaccessible (e.g. trapped equity in a property) please refer to the following guidance:"
+        text: "Guidance on trapped capital"
```
starswan, line 617: One of these 2 looks like a duplicate that could be removed

PR #267, opened 2022-12-07
EL-532 add 'blind' component for disregarded benefits

starswan:
https://dsdmoj.atlassian.net/browse/EL-532

### DIFF
--- a/app/views/estimate_flow/benefits.html.slim
+++ b/app/views/estimate_flow/benefits.html.slim
@@ -4,14 +4,11 @@
 .govuk-grid-column-two-thirds
   - if @form.benefits.blank?
 
-    = form_for(@form, url: variable_controller_benefit_path(action: :add)) do |form|
-      = form.govuk_error_summary t("generic.error_summary_title")
-      span.govuk-caption-l = t(".caption")
-      = form.govuk_collection_radio_buttons :add_benefit, yes_no_options, :first, :last,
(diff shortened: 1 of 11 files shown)


app/views/estimate_flow/benefits.html.slim
```diff
@@ -35,8 +35,16 @@
 
     = form_for(@form, url: variable_controller_benefit_path(action: :add)) do |form|
 
-      = form.govuk_collection_radio_buttons :add_benefit, yes_no_options, :first, :last,
-              legend: { text: t(".add_more_legend"), size: "l" }
+      = form.govuk_radio_buttons_fieldset :add_benefit,
+              legend: { text: t(".add_more_legend"), size: "l" },
+              hint: { text: t(".add_more_hint") } do
+
+        = render "shared/disregarded_benefits"
```
patrick-laa, line 42: Something to clear up: When Chris talks about the "Does your client receive any other benefits?" page, he's not (just) talking about the benefits page when the user has already entered one benefit on it. "Does your client receive any other benefits?" is supposed to be the new title of the page even when a user has just reached it (because 'other' refers to the Housing Benefit screen the user has just submitted).

So we need to add this partial on line ~10 of this view as well, and (although it's not explicitly specified in the ticket) update the legend for that radio button fieldset to include the word 'other'

patrick-laa (approved):

PR #272, opened 2022-12-08
EL-565: smarter number fields

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-565)

- Allow users to enter non-number characters into integer input fields
- If the user enters something non-numeric into an integer or money field, retain the string value and use a `numericality` validation to flag its issues
- Display the original string value back to the user alongside the validation message if user-entered value is non-numeric

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -49,6 +49,7 @@ en:
               blank: Income tax cannot be blank
             national_insurance:
               blank: National Insurance cannot be blank
+              not_a_number: National Insurance must be a number, if this does not apply enter 0
             frequency:
               blank: Please select how frequently you receive employment income
         partner_employment_form:
--- a/spec/features/employment_page_spec.rb
+++ b/spec/features/employment_page_spec.rb
@@ -80,6 +80,21 @@ RSpec.describe "Employment page" do
         end
       end
 
+      context "when I enter something invalid" do
+        before do
+          fill_in "employment-form-gross-income-field", with: "5,000"
+          fill_in "employment-form-income-tax-field", with: "1000"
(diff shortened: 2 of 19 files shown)


spec/features/employment_page_spec.rb
```diff
@@ -90,11 +90,26 @@
...
+        fill_in "employment-form-gross-income-field", with: "5,000"
+        fill_in "employment-form-income-tax-field", with: "1000"
+        fill_in "employment-form-national-insurance-field", with: "foo"
+        select_radio_value("employment-form", "frequency", "monthly")
+        click_on "Save and continue"
+      end
+
+      it "shows me an error message" do
+        expect(page).to have_content employment_page_header
+        expect(page).to have_content "There is a problem"
```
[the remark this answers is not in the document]
patrick-laa: Fair enough. The only reason I hadn't is that the finalised error message to display here is covered by another ticket. But I take the point that the test should check that a relevant error message is displayed!

config/locales/en.yml
```diff
@@ -49,6 +49,7 @@ en:
               blank: Income tax cannot be blank
             national_insurance:
               blank: National Insurance cannot be blank
+              not_a_number: National Insurance must be a number, if this does not apply enter 0
             frequency:
```
starswan, line 53: Now we know this, every fully_validatable_integer needs a translation for not_a_number...

patrick-laa: @starswan we have a separate ticket for this, https://dsdmoj.atlassian.net/browse/EL-566

patrick-laa:
@starswan could you hit the 'Approve' button when you get a mo?

starswan (approved):

PR #277, opened 2022-12-09
ELL-499 Increase delay before probes start checking, because of conce…

davidread:
…rn about early temrination


willc-work (approved):

davidread:
Closing - I think we concluded the app started up comfortably within 30s

PR #287, opened 2022-12-12
EL-558 retry CFE 502 errors

starswan:
https://dsdmoj.atlassian.net/browse/EL-558

Retry CFE 502 errors to (help) prevent issues with transient errors


app/lib/cfe_connection.rb
```diff
@@ -95,6 +95,12 @@ def cfe_connection
     @cfe_connection ||= Faraday.new(url: CFE_HOST, headers: { "Accept" => "application/json;version=5" }) do |faraday|
       faraday.request :json
       faraday.response :json
+
+      # retry 502 errors from CFE requests
+      faraday.request :retry, max: 3, interval: 0.05,
+                      interval_randomness: 0.5, backoff_factor: 2,
+                      retry_statuses: [502],
```
patrick-laa, line 102: We _occasionally_ see 504s as well, which I _think_ are another symptom of the same problem - can we add that as a status to the list?

patrick-laa (approved):

app/lib/cfe_connection.rb
starswan: I'm wondering if we should avoid retrying GETs as we know they fail - but maybe it's ok as if it works we know that the 500 never made it as far as the application itself

PR #293, opened 2022-12-13
EL-562 update income labels and spacing

willc-work:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-562)

Update size of labels to match first field, adjust content and amend spacing below each field in the employment input fields

Checklist

### DIFF
--- a/app/views/shared/_money_input.html.slim
+++ b/app/views/shared/_money_input.html.slim
@@ -1,9 +1,7 @@
--label_size ||= nil
 = form.govuk_text_field field,
                         width:,
-                        label: { text: label_text, size: label_size },
+                        label: { text: label_text },
                         prefix_text: BuildEstimatesHelper::POUND,
                         value: decimal_as_money_string(form, field),
                         pattern: "[0-9]*",
                         inputmode: "numeric"
-div class="govuk-!-padding-bottom-3"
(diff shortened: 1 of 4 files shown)


app/views/shared/_money_input.html.slim
```diff
@@ -1,7 +1,8 @@
 = form.govuk_text_field field,
                         width:,
-                        label: { text: label_text },
+                        label: { text: label_text, size: "m" },
```
patrick-laa, line 3: Hang on, this is going to change the size of the label for money inputs on _every_ screen, which means we'll end up with some fields on screens looking distinctly odd, e.g.:
Screenshot 2022-12-13 at 15 23 13
(Until now the type and amount labels have been the same size)

Do we want to pass in size as an optional parameter to this partial, so that it defaults to its default ('s'?), and can be overridden when used on the employment screen?

willc-work: Done that.

app/views/shared/_money_input.html.slim
```diff
@@ -1,7 +1,9 @@
+-label_size ||= "s"
```
patrick-laa, line 1: Sorry, I was wrong, the default _isn't_ "s". The size is now right, but this is still causing the label to be *bold* on all the other screens when previously it wasn't.

patrick-laa (approved):

patrick-laa (approved):

PR #297, opened 2022-12-14
EL-558 re-use faraday connection for every CFE request

starswan:
https://dsdmoj.atlassian.net/browse/EL-558

Re-use Faraday connection object for each of the requests to CFE, so that HTTP keep-alives can be active as we will make multiple calls down the same pipe (hopefully)


patrick-laa:
This seems like a reasonable refactor in and of itself, but I think to get any benefit by reusing the same pipe we need to tell Faraday to use the [NetHttpPersistent adapter](https://github.com/lostisland/faraday-net_http_persistent) - c.f. https://www.zorched.net/2013/10/07/3-5x-increase-in-performance-with-a-one-line-change/, https://github.com/lostisland/faraday/pull/778

patrick-laa (approved):

PR #313, opened 2022-12-21
Tidy PR template

davidread:
Jira ticket: none

Hopefully adding this formatting is agreeable and not simply my pedancy gone rogue...

Checklist


willc-work (commented):
We had a checklist like this on apply that required it to be ticked off and it became a bit of a pain. If it is marked ready for review but some of the boxes are not ticked should we refrain from reviewing it, or ask for the boxes to be ticked off before we start? When we added the template we agreed to not use tick boxes and just use a reminder list. Happy to discuss further, I am happy with the other changes

davidread:
@willc-work cool, I've removed the checkboxes now

willc-work (approved):

PR #326, opened 2023-01-04
EL-570: Add cost of living wording

patrick-laa:
[Link to the Jira ticket](https://dsdmoj.atlassian.net/browse/EL-570)

Note, please do not deploy this to production before 10 Jan 2023. Yes, this is not ideal, but the mechanism to introduce copy changes automatically on specific dates has not yet been built.

[EL-570]: https://dsdmoj.atlassian.net/browse/EL-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -439,7 +439,7 @@ en:
           - Widow’s Pension lump sum payments
           - Windrush Compensation Scheme payments
     benefits:
-      hint_html: For example, Child Benefit or tax credits. Do not include Housing Benefit, government Cost of Living Payments, or any other disregarded benefits.
+      hint: For example, Child Benefit or tax credits. Do not include Housing Benefit, government Cost of Living Payments, or any other disregarded benefits.
   estimate_flow:
     partner:
       partner_guidance:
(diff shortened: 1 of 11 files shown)


willc-work:
Re: releasing on 10th I have added a slack reminder in the eligibility-team channel to remind us to release this the morning of the 10th

config/locales/en.yml
```diff
@@ -385,8 +385,23 @@ en:
       text: "Guidance on trapped capital"
       link: "https://legalaidlearning.justice.gov.uk/course/view.php?id=186"
     disregarded_benefits:
-      text: Disregarded benefits
-      hint: The following benefits are disregarded from the financial eligibility calculation
+      text: Government Cost of Living Payments and disregarded benefits
+      cost_of_living:
+        text: Government Cost of Living Payments
+        list:
+          - Government Cost of Living Payments are disregarded from the financial eligibility calculation, for example
```
willc-work, line 392: should this line end with a colon ":", its not in the ticket but it does precede a list, which is where we would usually use one.

patrick-laa: I'll double-check with Chris. You're right, we use a colon prior to a list in several places elsewhere.

patrick-laa: @willc-work Chris has confirmed that in this case he does _not_ want a colon.

willc-work (approved):
I have some queries on translations but they are not blockers

PR #333, opened 2023-01-06
EL-601: Missing fields on check answers

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-601)

## What changed and why

We weren't showing summaries for the 'passporting' or 'proceeding type' attributes on the 'check answers' screen.

## Checklist

### DIFF
--- a/app/services/check_answers/applicant_section.rb
+++ b/app/services/check_answers/applicant_section.rb
@@ -3,18 +3,25 @@ module CheckAnswers
     include ActiveModel::Model
     include ActiveModel::Attributes
 
-    FIELDS = %i[over_60
+    FIELDS = %i[domestic_abuse
+                over_60
                 partner
-                employed].freeze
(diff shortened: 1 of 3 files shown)


app/services/check_answers/applicant_section.rb
```diff
@@ -3,18 +3,25 @@ class ApplicantSection
     include ActiveModel::Model
     include ActiveModel::Attributes
 
-    FIELDS = %i[over_60
+    FIELDS = %i[domestic_abuse
```
starswan, line 6: not sure domestic_abuse should be a field - it's derived from proceeding type?

patrick-laa: You're right, I forgot that we have a mechanism in check answers to handle enums like `proceeding_type` so I didn't need to convert it to a boolean.

PR #343, opened 2023-01-10
EL-607 fix crash when submitting a partner during a passported application

starswan:
https://dsdmoj.atlassian.net/browse/EL-607

## What changed and why

When filling in a passported application with a partner, the 'employed' question is never asked of the partner, so the 'employed' field was nil rather than false. This upset CFE which wanted true/false only. This fixes this so we only send true or false during this process

### DIFF
--- a/app/services/submit_partner_service.rb
+++ b/app/services/submit_partner_service.rb
@@ -32,10 +32,12 @@ class SubmitPartnerService < BaseCfeService
   end
 
   def employments
-    return [] unless partner[:employed] && !@applicant_form.passporting
-
-    form = PartnerEmploymentForm.from_session(@cfe_session_data)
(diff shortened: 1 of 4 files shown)


app/services/submit_partner_service.rb
```diff
@@ -32,10 +32,12 @@ def irregular_incomes
   end
 
   def employments
-    return [] unless partner[:employed] && !@applicant_form.passporting
-
-    form = PartnerEmploymentForm.from_session(@cfe_session_data)
-    CfeParamBuilders::Employments.call(form)
+    if partner[:employed] && !@applicant_form.passporting
```
patrick-laa, line 35: 👍 for readability improvement!

patrick-laa (approved):

PR #387, opened 2023-01-30
EL-619: New confirmation panel on results page

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-619)

## What changed and why

Updated the blue panel on the results screen (and top section on the print screen) with new copy

## Checklist

### DIFF
--- a/app/assets/stylesheets/results.scss
+++ b/app/assets/stylesheets/results.scss
@@ -13,4 +13,5 @@ tr.solid-bottom-border {
 // make the confirmation panel a neutral colour
 .panel-blue {
   background: govuk-colour("blue");
+  color: white !important;
 }
--- a/app/assets/stylesheets/results.scss
+++ b/app/assets/stylesheets/results.scss
@@ -13,5 +13,7 @@ tr.solid-bottom-border {
 // make the confirmation panel a neutral colour
 .panel-blue {
   background: govuk-colour("blue");
-  color: white !important;
+  * {
+    color: white !important;
+  }
 }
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/results.scss
```diff
@@ -9,4 +9,7 @@ tr.solid-bottom-border {
 // make the confirmation panel a neutral colour
 .panel-blue {
   background: govuk-colour("blue");
-}
+  * {
+    color: white !important;
+  }
+}
```
starswan, line 15: This looks quite extreme? Think a comment would be useful

app/models/calculation_result.rb
```diff
@@ -6,21 +6,10 @@ def initialize(api_response)
     @api_response = api_response
   end
 
-  def eligible?
-    decision == "eligible"
-  end
-
   def decision
     api_response.dig(:result_summary, :overall_result, :result) || "ineligible"
   end
```
starswan, line 11: Must confess this method has always made me slightly nervous - and now more so that we're rending a template based on the API result (which has more valid values than we are rendering, we just know that they will never happen for our inputs). Ideally we should have 2 (or 3) classes so that you can't ask the contribution if you're ineligible (or eligible now?) - but I'm not quite sure how much we should do. A comment and/or limiting the return values might be enough. WDYT?

app/assets/stylesheets/results.scss
patrick-laa: Good idea.

app/models/calculation_result.rb
patrick-laa: Good point! I think adding a comment and limiting the return values as you suggest is enough for now. I'll do that.

willc-work (approved):

PR #388, opened 2023-01-31
Decision doesn't need to be defaulted 

starswan:
decision it is always present in the API response, so it doesn't need to be defaulted.

It is worth noting that the style a = something || 0 doesn't get caught by the coverage tools - the || 0 is not considered a branch.


patrick-laa (approved):

patrick-laa:
I've now made this change in #387 as I had to update that line of code anyway.

PR #402, opened 2023-02-02
EL-645 benefits guidance for controlled work

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-645)

Make the disregarded benefits partial more flexible for levels of help 

## Checklist

### DIFF
--- a/app/views/benefits/edit.html.slim
+++ b/app/views/benefits/edit.html.slim
@@ -18,4 +18,5 @@
 
     = render "form", form:
 
-= render "shared/disregarded_benefits_sidebar"
+  - level_of_help = @estimate.level_of_help == "controlled" ? "controlled" : "certificated"
+= render "shared/disregarded_benefits_sidebar", translation_text: level_of_help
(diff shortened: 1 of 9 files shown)


app/views/benefits/edit.html.slim
```diff
@@ -18,4 +18,8 @@
 
     = render "form", form:
 
-= render "shared/disregarded_benefits_sidebar"
+- if FeatureFlags.enabled?(:controlled)
```
patrick-laa, line 21: I don't think you need to check the feature flag here: If `:controlled` is _not_ enabled then @estimate.level_of_help will never be equal to `"controlled"`, so your `level_of_help` assignment below still works. So lines 21, 24 and 25 can be deleted, because lines 22 and 23 cover all cases.

willc-work: I ended up wrapping these sidebars in the feature flag check because it was blowing up some tests without it. turns out @estimate was not available in edit/update hence only some tests breaking.

patrick-laa (approved):

PR #405, opened 2023-02-03
EL-681 remove disregarded and passported benefits from suggestions

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-681)

remove the disregarded and passported benefits from the list returned by cfe before we display it as suggestions

## Checklist

### DIFF
--- a/app/lib/cfe_connection.rb
+++ b/app/lib/cfe_connection.rb
@@ -13,8 +13,7 @@ class CfeConnection
   end
 
   def state_benefit_types
-    response = cfe_connection.get("state_benefit_type")
-    response.body.map { _1["name"] }.uniq
+    cfe_connection.get("state_benefit_type").body
   rescue StandardError
     []
   end
--- a/app/models/benefit_model.rb
+++ b/app/models/benefit_model.rb
@@ -36,8 +36,7 @@ class BenefitModel
   end
 
   def benefit_list
-    full_list = cfe_benefit_list.pluck("label", "name", "exclude_from_gross_income")
-    display_list = full_list.reject { |item| item[2] == true || item[0].in?(PASSPORTED_BENEFITS) }
-    display_list.map { |row| row[1] }
+    display_list = cfe_benefit_list.reject { _1["exclude_from_gross_income"] || _1["label"].in?(PASSPORTED_BENEFITS) }
+    display_list.map { _1["name"] }.uniq
   end
 end
--- a/app/models/benefit_model.rb
+++ b/app/models/benefit_model.rb
@@ -5,6 +5,15 @@ class BenefitModel
 
   FREQUENCY_OPTIONS = %w[every_week every_two_weeks every_four_weeks monthly].freeze
 
+  PASSPORTED_BENEFITS = %w[
+    age_related_payment
+    universal_credit
+    income_support
+    jobseekers_allowance
+    employment_support_allowance
+    pension_credit
+  ].freeze
+
   attribute :id
 
   EDITABLE_ATTRIBUTES = %i[benefit_type benefit_amount benefit_frequency].freeze
@@ -22,7 +31,13 @@ class BenefitModel
     FREQUENCY_OPTIONS.map { [_1, I18n.t("estimate_flow.benefits.frequencies.#{_1}")] }
   end
 
+  def cfe_benefit_list
+    CfeConnection.connection.state_benefit_types
+  end
+
   def benefit_list
-    @benefit_list ||= CfeConnection.connection.state_benefit_types
+    full_list = cfe_benefit_list.pluck("label", "name", "exclude_from_gross_income")
+    display_list = full_list.reject { |item| item[2] == true || item[0].in?(PASSPORTED_BENEFITS) }
+    display_list.map { |row| row[1] }
   end
 end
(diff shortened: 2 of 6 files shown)


app/models/benefit_model.rb
```diff
@@ -5,6 +5,51 @@ class BenefitModel
...
 
+  PASSPORTED_BENEFITS = [
+    "Universal Credit",
+    "Income Support",
+    "Income-based Jobseeker's Allowance (JSA)",
+    "Income-related Employment and Support Allowance (ESA)",
+    "Guarantee Credit element of Pension Credit",
+  ].freeze
+
+  DISREGARDED_BENEFITS = [
```
patrick-laa, line 16: This list doesn't quite match up to what CFE returns. E.g. CFE has `Carer's Allowance` which is one character different from `Carer’s Allowance` which is here.

The good news is that CFE also returns an `exclude_from_gross_income` attribute for each benefit, and if we ignore every item that has that set to true, we exclude all the below from the list.

app/models/benefit_model.rb
```diff
@@ -5,6 +5,51 @@ class BenefitModel
 
   FREQUENCY_OPTIONS = %w[every_week every_two_weeks every_four_weeks monthly].freeze
 
+  PASSPORTED_BENEFITS = [
+    "Universal Credit",
+    "Income Support",
+    "Income-based Jobseeker's Allowance (JSA)",
```
patrick-laa, line 11: CFE doesn't return this, but it does return `Jobseeker’s Allowance (JSA)`. Likewise it returns `Employment and Support Allowance (ESA)` and `Pension Credit`. Do we want to remove those elements from the list? If so we should modify those strings in this array. If not, we should remove these strings.

app/lib/cfe_connection.rb
```diff
@@ -14,7 +14,7 @@ def status
 
   def state_benefit_types
     response = cfe_connection.get("state_benefit_type")
-    response.body.map { _1["name"] }.uniq
+    response.body.map { |benefit| [benefit["name"], benefit["exclude_from_gross_income"]] }.uniq
```
patrick-laa, line 17: This is a bit subjective, but I don't like having `state_benefit_types` return an array of arrays containing a bunch of booleans with no indication (unless you read the code inside the method) of what those booleans mean. I'd prefer it to return an array of hashes where each hash has a `name` and `exclude_from_gross_income` key. (At which point you could just return the whole body, since the body already has exactly those things.

app/models/benefit_model.rb
```diff
@@ -5,6 +5,14 @@ class BenefitModel
 
   FREQUENCY_OPTIONS = %w[every_week every_two_weeks every_four_weeks monthly].freeze
 
+  PASSPORTED_BENEFITS = [
+    "Universal Credit",
+    "Income Support",
+    "Jobseeker's Allowance (JSA)",
```
patrick-laa, line 11: I think CFE return's `Jobseeker’s Allowance (JSA)` (different type of apostrophe). I wonder if, to avoid this sort of fragility, it would be better to exclude benefits based on the _label_ attribute (`jobseekers_allowance` in this case), which has less fiddly formatting.

app/lib/cfe_connection.rb
patrick-laa: (Feel free to ignore this one if you disagree - like I say, this is subjective)

willc-work: I agree. Makes sense that cfe_connection is doing as little as possible to the data so I have set it to return everything

app/models/benefit_model.rb
```diff
@@ -22,7 +31,13 @@ def benefit_options
...
   end
 
+  def cfe_benefit_list
+    CfeConnection.connection.state_benefit_types
+  end
+
   def benefit_list
-    @benefit_list ||= CfeConnection.connection.state_benefit_types
+    full_list = cfe_benefit_list.pluck("label", "name", "exclude_from_gross_income")
+    display_list = full_list.reject { |item| item[2] == true || item[0].in?(PASSPORTED_BENEFITS) }
```
patrick-laa, line 40: Another aesthetic thing: you don't really need the `pluck`. You could do 

```ruby
display_list = cfe_benefit_list.reject { _1["exclude_from_gross_income"] || _1["label"].in?(PASSPORTED_BENEFITS }
display_list.map { _1["name"] }
```

And that would mean whenever you use an attribute you can refer to it by its name rather than a number.

app/lib/cfe_connection.rb
```diff
@@ -13,8 +13,7 @@ def status
   end
 
   def state_benefit_types
-    response = cfe_connection.get("state_benefit_type")
-    response.body.map { _1["name"] }.uniq
```
patrick-laa, line 17: I think we still need a `uniq` - otherwise Child Benefit crops up twice in our list. Maybe it needs to move into BenefitModel#benefit_list?

spec/unit/cfe_connection_spec.rb
```diff
@@ -6,6 +6,7 @@ def post(*); end
 RSpec.describe CfeConnection do
   let(:connection) { described_class.new }
   let(:root_url) { "https://check-financial-eligibility-partner-staging.cloud-platform.service.justice.gov.uk/assessments" }
+  let(:root_url_one) { "https://check-financial-eligibility-partner-staging.cloud-platform.service.justice.gov.uk" }
```
patrick-laa, line 9: I think this is redundant?

app/lib/cfe_connection.rb
willc-work: I have moved the `uniq` to the `benefit_model` and specifically to the `benefit_list` method used to determine what to display. I think this method should just get whatever CFE sends it and pass that on... I also just properly read the second sentence, so yes i agree. 🤦

patrick-laa (approved):

PR #419, opened 2023-02-09
EL-678 remove cost of living feature flag

starswan:
https://dsdmoj.atlassian.net/browse/EL-678

### DIFF
--- a/app/views/estimate_flow/benefits.html.slim
+++ b/app/views/estimate_flow/benefits.html.slim
@@ -10,7 +10,7 @@
       = render "add_benefit_form",
               form:,
               legend: { text: t(".legend"), size: "xl", tag: "h1" },
-              hint_text: t("generic.benefits.hint_with_col")
+              hint_text: t("generic.benefits.hint")
   - else
     = form_for(@form, url: variable_controller_benefit_path(action: :add)) do |form|
       = form.govuk_error_summary t("generic.error_summary_title")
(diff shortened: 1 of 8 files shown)


app/views/estimate_flow/forms/_other_income.html.slim
```diff
@@ -26,7 +26,7 @@
     h2.govuk-heading-m class="govuk-!-margin-top-9 govuk-!-margin-bottom-2"
       = t("estimate_flow.#{i18n_key}.#{type}.label")
     span.govuk-caption-m class="govuk-!-margin-bottom-4"
-      - if FeatureFlags.enabled?(:disregard_cost_of_living) && type == :other
+      - if type == :other
```
patrick-laa, line 29: If you rename the `estimate_flow.assets(/partner_assets).other.hint_with_col` to `estimate_flow.assets(/partner_assets).other.hint` then we can remove this if/else block entirely.

app/views/estimate_flow/benefits.html.slim
```diff
@@ -37,7 +37,7 @@
       = render "add_benefit_form",
                form:,
                legend: { text: t(".legend"), size: "l" },
-               hint_text: FeatureFlags.enabled?(:disregard_cost_of_living) ? t(".add_more_hint_with_col") : t(".add_more_hint")
```
patrick-laa, line 40: I think we should remove the "orphaned" strings from en.yml, like `estimate_flow.benefits.add_more_hint` here.

app/views/estimate_flow/forms/_other_income.html.slim
starswan: Thanks - I've just worked out that hint_with_col means 'cost_of_living'

patrick-laa (approved):

PR #423, opened 2023-02-10
EL-652: Stop using git-crypt

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-652)

## What changed and why

Move secrets out of source control. Contrary to the ticket, I've not replaced the secrets file with a config map; rather I'm directly referencing values files from the _envs file. This is because, having been the one to add env vars from time to time, I've found the extra layer of abstraction annoying and I can't see the benefit of having it there, so eliminating it makes sense to me here.

Note that the main reason I've moved the values files into a new folder is that otherwise Git doesn't show their new unencrypted contents in the diff of this PR, because I suppose it respects the git-crypt settings of the branch it's moving _into_.

## Checklist


willc-work (approved):
Looks good but would be good to get @exonian 's view on it too

exonian (approved):
This is excellent. Great call on the removal of complexity.

I feel like I have a preference for `values/staging.yaml` over `values/values-staging.yaml`, but it's non-blocking and there may also be compelling reasons not to.

PR #424, opened 2023-02-10
EL-696 passporting asylum support applications

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-696)
Add in asylum support page, translations and check answers

Splits the ApplicantCaseDetailsSection into ApplicantDetailsSection and CaseDetailsSection as the current section was heavily weighted towards case and not applicant. Happy to revert this if anyone feel sit is unnecessary.

## Checklist

### DIFF
--- a/app/forms/applicant_form.rb
+++ b/app/forms/applicant_form.rb
@@ -3,7 +3,7 @@ class ApplicantForm
   include ActiveModel::Attributes
   include SessionPersistable
 
-  attr_accessor :asylum_support, :level_of_help
+  attr_accessor :level_of_help
 
   PROCEEDING_TYPES = { domestic_abuse: "DA001", other: "SE003" }.freeze
 
--- a/app/forms/applicant_form.rb
+++ b/app/forms/applicant_form.rb
@@ -3,6 +3,8 @@ class ApplicantForm
   include ActiveModel::Attributes
   include SessionPersistable
 
+  attr_accessor :asylum_support, :level_of_help
+
   PROCEEDING_TYPES = { domestic_abuse: "DA001", other: "SE003" }.freeze
 
   PROCEEDING_ATTRIBUTE = %i[proceeding_type].freeze
@@ -15,8 +17,6 @@ class ApplicantForm
             presence: true,
             inclusion: { in: PROCEEDING_TYPES.values, allow_nil: true },
             if: -> { level_of_help != "controlled" }
-
-  attr_accessor :level_of_help
 
   BOOLEAN_ATTRIBUTES.each do |attr|
     attribute attr, :boolean
--- a/app/forms/asylum_support_form.rb
+++ b/app/forms/asylum_support_form.rb
@@ -1,0 +1,10 @@
+class AsylumSupportForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include SessionPersistable
+
+  ATTRIBUTES = %i[asylum_support].freeze
+
+  attribute :asylum_support, :boolean
+  validates :asylum_support, inclusion: { in: [true, false], allow_nil: false }
+end
--- a/app/lib/case_details_section.rb
+++ b/app/lib/case_details_section.rb
@@ -1,7 +1,7 @@
-class ApplicantCaseDetailsSection
+class CaseDetailsSection
   class << self
     def all_steps
-      %i[level_of_help matter_type applicant dependant_details]
+      %i[level_of_help matter_type asylum_support]
     end
 
     def steps_for(estimate)
@@ -9,7 +9,7 @@ class ApplicantCaseDetailsSection
     end
 
     def steps(estimate)
-      [level_of_help, matter_type(estimate), :applicant, dependant_details(estimate)].compact
+      [level_of_help, matter_type(estimate), asylum_support(estimate)].compact
     end
 
     def level_of_help
@@ -20,8 +20,8 @@ class ApplicantCaseDetailsSection
       :matter_type if estimate.controlled? && FeatureFlags.enabled?(:asylum_and_immigration)
     end
 
-    def dependant_details(estimate)
-      :dependant_details unless estimate.passporting
+    def asylum_support(estimate)
+      :asylum_support if estimate.level_of_help == "controlled" && estimate.upper_tribunal?
     end
   end
 end
--- a/app/lib/income_section.rb
+++ b/app/lib/income_section.rb
@@ -5,7 +5,7 @@ class IncomeSection
     end
 
     def steps_for(estimate)
-      if estimate.passporting
+      if estimate.passporting || estimate.asylum_support
         []
       else
         employment_steps(estimate) + housing_benefit_steps(estimate) + other_steps
--- a/app/models/estimate_model.rb
+++ b/app/models/estimate_model.rb
@@ -4,6 +4,7 @@ class EstimateModel
   include SessionPersistable
 
   ESTIMATE_BOOLEANS = %i[upper_tribunal
+                         asylum_support
                          over_60
                          passporting
                          vehicle_owned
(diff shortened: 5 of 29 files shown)


app/controllers/build_estimates_controller.rb
```diff
@@ -8,7 +8,8 @@ def update
       session_data.merge!(@form.session_attributes)
       estimate = load_estimate
 
-      next_step = StepsHelper.next_step_for(estimate, step)
+      next_step = estimate.asylum_support ? false : StepsHelper.next_step_for(estimate, step)
```
patrick-laa, line 11: Trying to add this logic here causes some weird side effects. For example, if you click yes to asylum support, then click the back button a few times to get you back to the first question in the flow, then try to go forwards again, this logic runs when trying to work out the next screen, so you get jumped to the check answers screen wrongly.

The "proper" way of doing this, which I think is the way we have to do it, is to modify each of the Section classes (apart from the CaseDetailsSection) to return an empty array in `steps_for(estimate)` if `estimate.asylum_support` is true. Much like we do for most sections if `estimate.passporting` is true.

app/lib/case_details_section.rb
```diff
@@ -20,12 +20,12 @@ def tribunal(estimate)
       :tribunal if estimate.level_of_help == "controlled" && FeatureFlags.enabled?(:asylum_and_immigration)
     end
 
-    def matter_type(estimate)
-      :matter_type if estimate.upper_tribunal
+    def asylum_support(estimate)
+      :asylum_support if estimate.level_of_help == "controlled" && FeatureFlags.enabled?(:asylum_and_immigration) && estimate.upper_tribunal
```
patrick-laa, line 24: You don't need the FeatureFlags check here, because `estimate.upper_tribunal` can only be true if the flag is enabled anyway.

app/services/check_answers/section_id_finder.rb
```diff
@@ -3,6 +3,7 @@ class SectionIdFinder
     IDS = {
       level_of_help: "section-level-of-help-header",
       tribunal: "subsection-case_details-header",
+      asylum_support: "subsection-case_details-header",
```
patrick-laa, line 6: I think you want this to be "subsection-asylum_support-header" - that way when you return to this section after clicking the asylum support change link, you're brought to exactly the right subsection you clicked on.

willc-work: updated, weirdly both work and take you to the correct page 🤷

app/lib/income_section.rb
```diff
@@ -5,7 +5,7 @@ def all_steps
     end
 
     def steps_for(estimate)
-      if estimate.passporting
+      if estimate.passporting || estimate.asylum_support
```
patrick-laa, line 8: To avoid a possible edge case we should add this to the three partner sections as well: If a user initially answers `no` to the asylum support question, then on the applicant details screen indicates they have a partner, then modifies their responses at some point to now say `yes` to the asylum support question, we need to be sure that they are _not_ then prompted to fill out the partner details.

app/forms/asylum_support_form.rb
```diff
@@ -0,0 +1,10 @@
+class AsylumSupportForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include SessionPersistable
+
+  ATTRIBUTES = %i[asylum_support].freeze
+
+  attribute :asylum_support, :boolean
```
patrick-laa, line 8: We also need to send this boolean to CFE as an applicant attribute (although CFE will only know what to do with it when https://github.com/ministryofjustice/check-financial-eligibility-partner/pull/89 is merged)

app/lib/case_details_section.rb
```diff
@@ -1,15 +1,15 @@
-class ApplicantCaseDetailsSection
+class CaseDetailsSection
   class << self
     def all_steps
-      %i[level_of_help tribunal matter_type applicant dependant_details]
+      %i[level_of_help tribunal asylum_support matter_type]
```
patrick-laa, line 4: One more thing, sorry - as a result of a bunch of discussion over the last week, tribunal screen is being merged into the matter type screen in EL-718 (now in progress), so asylum support will need to come _after_ matter type now.

app/services/submit_applicant_service.rb
```diff
@@ -2,12 +2,17 @@ class SubmitApplicantService < BaseCfeService
...
 
-    applicant = {
+    base_attributes = {
       date_of_birth: estimate.over_60 ? 70.years.ago.to_date : 50.years.ago.to_date,
       has_partner_opponent: false,
       receives_qualifying_benefit: estimate.passporting,
       employed: estimate.employed,
     }
+    applicant = if FeatureFlags.enabled?(:asylum_and_immigration)
+                  base_attributes.merge({ asylum_support: estimate.asylum_support })
```
patrick-laa, line 12: The CFE API attribute [is called](https://github.com/ministryofjustice/check-financial-eligibility-partner/blob/873bd8fa23696acc9fc7cd6fd140851878a4473c/spec/requests/swagger_docs/v5/applicants_spec.rb#L39) `receives_asylum_support

patrick-laa (commented):
One tiny change, other than that I think we're there.

patrick-laa (approved):

app/forms/applicant_form.rb
```diff
@@ -3,6 +3,8 @@ class ApplicantForm
   include ActiveModel::Attributes
   include SessionPersistable
 
+  attr_accessor :asylum_support, :level_of_help
```
patrick-laa, line 6: I think this change is redundant - this accessor doesn't get used

app/models/estimate_model.rb
```diff
@@ -4,6 +4,7 @@ class EstimateModel
   include SessionPersistable
 
   ESTIMATE_BOOLEANS = %i[upper_tribunal
```
patrick-laa, line 6: I should have removed this `upper_tribunal` boolean when I got rid of the tribunal screen. Sorry. To avoid further confusion, can you remove it from the list here?

app/models/estimate_model.rb
```diff
@@ -39,4 +41,9 @@ def partner_owns_property?
   def controlled?
     level_of_help == "controlled"
   end
+
+  def upper_tribunal?
+    # controlled_proceeding_type != "SE003"
```
patrick-laa, line 46: Let's remove this commented out code.

app/models/estimate_model.rb
```diff
@@ -39,4 +41,9 @@ def partner_owns_property?
...
+    controlled_proceeding_type.in?(%w[IM030 IA031])
```
patrick-laa, line 47: Let's make this method safer - at the moment this will explode if controlled_proceeding_type is nil. Maybe `controlled_proceeding_type&.in?(...)`?

patrick-laa (commented):
A couple of bugs to iron out here, and a couple of tidy-ups

app/services/submit_applicant_service.rb
```diff
@@ -2,12 +2,18 @@ class SubmitApplicantService < BaseCfeService
...
-    applicant = {
+    base_attributes = {
       date_of_birth: estimate.over_60 ? 70.years.ago.to_date : 50.years.ago.to_date,
       has_partner_opponent: false,
       receives_qualifying_benefit: estimate.passporting,
       employed: estimate.employed,
     }
+    applicant = if FeatureFlags.enabled?(:asylum_and_immigration)
+                  base_attributes.merge({ receives_asylum_support: session_data["asylum_support"],
+                                          receives_qualifying_benefit: session_data["asylum_support"] })
```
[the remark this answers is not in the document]
willc-work: I have used a combination solution to merge the "receives_asylum_support: asylum_support_form.asylum_support || false" if the flag is in place as otherwise it breaks a load of tests.

patrick-laa (approved):

PR #430, opened 2023-02-15
EL-709: De-identify user in logs

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-709)

## What changed and why

Separate estimate_id into concepts of assessment_code and assessment_id. This way we have one thing that appears in URLs (and therefore also appears in logs alongside user IP), and something else that is the key to all the user entered financial values stored in Redis. 

While I'm here, rename all other instances of `estimate_id` that aren't baked in by the framework, as neither CCQ nor CFE officially deals with estimates, they deal with checks and assessments.

## Checklist

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,28 +4,32 @@ class ApplicationController < ActionController::Base
 
 private
 
-  def session_data(id = estimate_id)
-    session[session_key(id)] ||= {}
+  def assessment_id
+    Digest::SHA256.hexdigest(assessment_code + (cookies["SessionData"] || ""))
   end
 
-  def session_key(id)
-    "estimate_#{id}"
+  def session_data
+    session[assessment_id] ||= {}
   end
 
-  def track_page_view(assessment_id: nil, page: page_name)
-    track("page_view", assessment_id:, page:)
+  def track_page_view(page: page_name)
+    track("page_view", page:)
   end
 
-  def track_validation_error(assessment_id: nil, page: page_name)
-    track("validation_message", assessment_id:, page:)
-    track_page_view(assessment_id:, page:)
+  def track_validation_error(page: page_name)
+    track("validation_message", page:)
+    track_page_view(page:)
   end
 
-  def track(event_type, assessment_id:, page:)
+  def track(event_type, page:)
     AnalyticsService.call(event_type:,
                           browser_id: cookies[BROWSER_ID_COOKIE],
-                          assessment_id:,
+                          assessment_code:,
                           page:)
+  end
+
+  def assessment_code
+    nil
   end
 
   def page_name
--- a/app/controllers/estimates_controller.rb
+++ b/app/controllers/estimates_controller.rb
@@ -4,12 +4,9 @@ class EstimatesController < ApplicationController
   end
 
   def create
-    redirect_to estimate_path(params[:estimate_id])
-  end
-
-  def show
     @model = CfeService.call(session_data)
     track_page_view(page: :view_results)
+    render :show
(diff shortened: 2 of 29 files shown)


app/controllers/application_controller.rb
```diff
@@ -4,30 +4,34 @@ class ApplicationController < ActionController::Base
...
     AnalyticsService.call(event_type:,
                           browser_id: cookies[BROWSER_ID_COOKIE],
-                          assessment_id:,
+                          assessment_code:,
                           page:)
   end
 
+  def assessment_code
+    nil
+  end
```
starswan, line 33: This looks strange - should this method exist if it's being overridden by derived classes?

app/controllers/estimates_controller.rb
```diff
@@ -4,32 +4,34 @@ def new
...
   def create
-    @model = CfeService.call(cfe_session_data(:estimate_id))
-    track_page_view(assessment_id: params[:estimate_id], page: :view_results)
-    render :show
+    redirect_to estimate_path(params[:estimate_id])
+  end
+
+  def show
+    @model = CfeService.call(session_data)
+    track_page_view(page: :view_results)
```
starswan, line 12: Not sure I'm happy with this change. IMHO the 'GET' should not invoke the service to create the assessment . The 'create' call was a little strange in rendering the 'show' template - this could just have been called 'create' (c.f. rails generated templates, which do just that but common-up the 'created' and 'shown' views for DRY purposes.

app/controllers/estimates_controller.rb
```diff
@@ -4,32 +4,34 @@ def new
...
-    track_page_view(assessment_id: params[:id], page: :download_results)
+    track_page_view(page: :download_results)
     html = render_to_string({
       template: "estimates/print",
       layout: "print_application",
       locals: {
-        :@model => CfeService.call(cfe_session_data(:id)),
-        :@answers => CheckAnswersPresenter.new(cfe_session_data(:id)),
+        :@model => CfeService.call(session_data),
+        :@answers => CheckAnswersPresenter.new(session_data),
```
starswan, line 34: This code generated warnings for me about being non rails 7.1 compatible. Not sure how to fix it through? Do we just set local variables?

app/services/analytics_service.rb
```diff
@@ -1,11 +1,11 @@
 class AnalyticsService
   class << self
-    def call(event_type:, page:, assessment_id: nil, browser_id: nil)
+    def call(event_type:, page:, assessment_code: nil, browser_id: nil)
```
starswan, line 3: this service is only ever called with 4 parameters, so we can remove the defaults

app/controllers/estimates_controller.rb
patrick-laa: OK. I do think there's a case for turning this into a GET/show (we invoke the service on other GETs like `print` and `download`, and I don't think the result's page should fundamentally be less bookmarkable than any of the other pages. But a change like that is controversial enough that it doesn't belong in this PR.

app/controllers/estimates_controller.rb
patrick-laa: I'll experiment and find out.

app/controllers/application_controller.rb
patrick-laa: It's not overriden by all derived classes, but it's _used_ by all of them (via `#track`). In cases where there is no assessment, but the track method want to be able to retrieve whatever the assessment code is, I think it's helpful having a `nil` default.

starswan (approved):

PR #449, opened 2023-02-21
EL-713: Start using postgres to record analytics

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-713)

## What changed and why

We now keep analytics in a postgres database.

## Checklist

### DIFF
--- a/app/services/analytics_service.rb
+++ b/app/services/analytics_service.rb
@@ -1,14 +1,12 @@
 class AnalyticsService
   class << self
     def call(event_type:, page:, assessment_code:, browser_id:)
-      hash = {
-        key: "CCQ Analytics Datum",
+      AnalyticsEvent.create(
         event_type:,
         page:,
         assessment_code:,
         browser_id:,
-      }
-      Rails.logger.info hash.to_json
+      )
     end
   end
 end
--- a/app/services/analytics_service.rb
+++ b/app/services/analytics_service.rb
@@ -1,12 +1,14 @@
 class AnalyticsService
   class << self
     def call(event_type:, page:, assessment_code:, browser_id:)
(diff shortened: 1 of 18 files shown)


app/services/analytics_service.rb
```diff
@@ -1,14 +1,12 @@
 class AnalyticsService
   class << self
     def call(event_type:, page:, assessment_code:, browser_id:)
-      hash = {
-        key: "CCQ Analytics Datum",
+      AnalyticsEvent.create(
```
starswan, line 4: should this be create! (although if it fails, I'm not sure I want a hard crash...?)

db/migrate/20230221125618_create_analytics_events.rb
```diff
@@ -0,0 +1,12 @@
+class CreateAnalyticsEvents < ActiveRecord::Migration[7.0]
+  def change
+    create_table :analytics_events do |t|
+      t.string :event_type
+      t.string :page
+      t.string :assessment_code
+      t.string :browser_id
```
starswan, line 7: do we want some null: falses and some indexes in here?

app/services/analytics_service.rb
patrick-laa: Good idea! I'll add a `!` and then catch and Sentrify any exceptions generated.

db/migrate/20230221125618_create_analytics_events.rb
patrick-laa: Cool. I think the final say about the right indexes will come out of EL-717, but as a start I think it makes sense to index the created_at and assessment_code columns, as a lot of queries are based around those.

starswan (approved):

PR #451, opened 2023-02-22
EL 638 matter type page certificated

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-638)

Added domestic abuse to the matter_type page
This change is to the certificated flow but it only visible when the controlled work flag is enabled, so the original flow remains unchanged.

## Checklist

### DIFF
--- a/app/forms/applicant_form.rb
+++ b/app/forms/applicant_form.rb
@@ -16,7 +16,7 @@ class ApplicantForm
   validates :proceeding_type,
             presence: true,
             inclusion: { in: PROCEEDING_TYPES.values, allow_nil: true },
-            if: -> { level_of_help != "controlled" }
+            if: -> { !FeatureFlags.enabled?(:controlled) && level_of_help != "controlled" }
 
   BOOLEAN_ATTRIBUTES.each do |attr|
     attribute attr, :boolean
--- a/app/forms/applicant_form.rb
+++ b/app/forms/applicant_form.rb
@@ -7,16 +7,16 @@ class ApplicantForm
 
   PROCEEDING_TYPES = { domestic_abuse: "DA001", other: "SE003" }.freeze
 
-  PROCEEDING_ATTRIBUTE = %i[proceeding_type].freeze
+  PROCEEDING_ATTRIBUTE = %i[legacy_proceeding_type].freeze
   BOOLEAN_ATTRIBUTES = %i[over_60 employed partner passporting].freeze
 
   ATTRIBUTES = BOOLEAN_ATTRIBUTES + PROCEEDING_ATTRIBUTE.freeze
 
-  attribute :proceeding_type
-  validates :proceeding_type,
+  attribute :legacy_proceeding_type
+  validates :legacy_proceeding_type,
             presence: true,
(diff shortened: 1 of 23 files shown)


app/forms/applicant_form.rb
```diff
@@ -14,7 +14,7 @@ class ApplicantForm
   validates :proceeding_type,
             presence: true,
             inclusion: { in: PROCEEDING_TYPES.values, allow_nil: true },
-            if: -> { level_of_help != "controlled" }
+            if: -> { !FeatureFlags.enabled?(:controlled) && level_of_help != "controlled" }
```
patrick-laa, line 17: I think this is the wrong feature flag. We're only introducing the `matter_type` screen behind the `asylum_and_immigration` feature flag, so if that flag is switched off, this is the only screen that certificated proceeding types can be added

app/views/estimate_flow/applicant.html.slim
```diff
@@ -9,13 +9,13 @@
 
     = form.govuk_error_summary t("generic.error_summary_title")
 
-    - if @form.level_of_help != "controlled"
+    - if !FeatureFlags.enabled?(:controlled) && @form.level_of_help != "controlled"
```
patrick-laa, line 12: As above, this should be the asylum_and_immigration feature flag.

app/views/estimate_flow/matter_type.html.slim
```diff
@@ -12,6 +12,9 @@
               label: { text: t(".immigration") }, hint: { text: t(".immigration_hint") }
       = form.govuk_radio_button :controlled_proceeding_type, MatterTypeForm::PROCEEDING_TYPES[:asylum],
               label: { text: t(".asylum") }, hint: { text: t(".asylum_hint") }
+      - if @estimate.level_of_help != "controlled"
```
patrick-laa, line 15: We'll also need to adjust the SubmitProceedingsService so that when this screen is shown and the domestic abuse case is not on the applicant form, we use the value from the MatterTypeForm (which should be renamed from `controlled_proceeding_type` to just `proceeding_type` I guess and not the value from ApplicantForm (which, to avoid duplicate keys in the session overwriting one another, should probably be renamed to something else)

app/helpers/applicant_helper.rb
```diff
@@ -1,15 +1,8 @@
...
-    links = {
+    {
       t("estimate_flow.applicant.passporting_guidance.text") => t("estimate_flow.applicant.passporting_guidance.#{level_of_help}.link"),
       t("estimate_flow.applicant.pensioner_guidance.text") => t("estimate_flow.applicant.pensioner_guidance.#{level_of_help}.link"),
     }
-    if level_of_help == "controlled"
-      links
-    else
-      links.merge({
-        t("estimate_flow.applicant.domestic_abuse_guidance.text") => t("estimate_flow.applicant.domestic_abuse_guidance.link"),
```
patrick-laa, line 11: I think we still want those the domestic abuse guidance link when `!FeatureFlags.enabled?(:asylum_and_immigration) && level_of_help != "controlled"`

patrick-laa (approved):
3 tiny things but I think we're there.

PR #475, opened 2023-03-02
EL-764: SubmitServices defer to StepsHelper

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-764)

## What changed and why

SubmitXServices now use StepsHelper's knowledge to reduce logic duplication (so that if the flow changes we don't have to update the services)

## Checklist


starswan (approved):
At one point I thought we should have a 'model' that reflected the data, so that we could ask it questions rather than constructing all these form objects inside the CFE interface. I guess this is a 'halfway-house' to that idea? 

Looks really nice, good to reduce the duplication of 'model data' logic.

PR #476, opened 2023-03-03
EL-757 - bugfix for pension disregards

willc-work:
display pension disregard only if it is used and alt hint text

[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-757)

## Checklist


app/models/calculation_result.rb
```diff
@@ -67,9 +67,14 @@ def capital_upper_threshold
   end
 
   def client_assessed_capital
-    monetise(capital_row_items(prefix: "").values.compact.sum(0))
+    monetise([api_response.dig(:result_summary, :disposable_income, :assessed_capital), 0].compact.max)
```
starswan, line 70: This should no longer be needed - assessed_capital cannot be negative

app/models/calculation_result.rb
```diff
@@ -67,9 +67,14 @@ def capital_upper_threshold
...
   end
 
+  # Why is this being used instead of the assessed capital from the result? This looks like CCQ doing calculations??
+  # def client_assessed_capital
+  #   monetise(capital_row_items(prefix: "").values.compact.sum(0))
+  # end
+
```
starswan, line 77: remove commented-out code. It was adding a+b+c+d because the total asset value wasn't available from CFE

app/models/calculation_result.rb
```diff
@@ -91,6 +96,10 @@ def pensioner_disregard_applied?
     api_response.dig(:result_summary, :capital, :pensioner_capital_disregard).positive?
   end
 
+  def pensioner_disregard_value
```
patrick-laa, line 99: You only use this in one place, and to check whether `pensioned_disregard_applied` is positive. So I think you're better off, instead of creating this method, updating the now incorrect `pensioner_disregard_applied?` - if it checked `:pensioner_disregard_applied` rather than `:pensioner_capital_disregard` it would be more accurate, and save us from having the term "pensioner_disregard_applied" refer to 2 different things.

app/models/calculation_result.rb
```diff
@@ -148,13 +157,21 @@ def client_capital_rows
...
 
   def client_capital_subtotal_rows
-    rows = {
-      total_capital: monetise(api_response.dig(:result_summary, :capital, :total_capital)),
-      pensioner_capital_disregard: monetise(-api_response.dig(:result_summary, :capital, :pensioner_capital_disregard)),
-      smod_disregard: monetise(-api_response.dig(:result_summary, :capital, :subject_matter_of_dispute_disregard)),
-    }
-
-    if has_partner? && pensioner_disregard_applied?
+    rows = if has_partner?
```
patrick-laa, line 160: I think this `if` statement is based on a slight misunderstanding of the ticket: there should never be a situation where pensioner capital disregard appears above smod disregard. The ticket is saying "if pensioner disregard is shown _at all_ in this table, show it beneath smod".

app/models/calculation_result.rb
```diff
@@ -148,13 +157,21 @@ def client_capital_rows
...
+             }
+           else
+             {
+               total_capital: monetise(api_response.dig(:result_summary, :capital, :total_capital)),
+               pensioner_capital_disregard: monetise(-api_response.dig(:result_summary, :capital, :pensioner_disregard_applied)),
+               smod_disregard: monetise(-api_response.dig(:result_summary, :capital, :subject_matter_of_dispute_disregard)),
+             }
+           end
+
+    if (has_partner? && pensioner_disregard_applied?) || (pensioner_disregard_applied? && !pensioner_disregard_value.positive?)
```
patrick-laa, line 174: As above, if we update `pensioner_disregard_applied` then the logic here gets a lot simpler. We only show the pensioner capital disregard row if (a) its value is positive and (b) there is no partner. So we exclude that row if the value is zero or there is a partner.

config/locales/en.yml
```diff
@@ -1472,12 +1472,15 @@ en:
...
           hint: ""
+          alt_hint: ""
         pensioner_capital_disregard:
           text: Pensioner disregard
           hint: Applied to total capital and capped at £100,000
+          alt_hint: Applied to total capital up to a maximum of £100,000
         smod_disregard:
           text: Disputed asset disregard
           hint: Equal to the assessed value of all assets marked as disputed and capped at £100,000
+          alt_hint: Applied to remaining capital after disputed asset disregard has been applied, and up to a maximum of £100,000
```
patrick-laa, line 1483: This is in the wrong place. We want to show this text on the _pensioner_ disregard line _if_ there is non-zero SMOD.

config/locales/en.yml
```diff
@@ -1472,12 +1472,15 @@ en:
         total_capital:
           text: Total capital
           hint: ""
+          alt_hint: ""
         pensioner_capital_disregard:
           text: Pensioner disregard
           hint: Applied to total capital and capped at £100,000
+          alt_hint: Applied to total capital up to a maximum of £100,000
```
patrick-laa, line 1479: This text is supposed to _replace_ the "capped at £100,000" text, not be an alternative to it. I had to clarify on the ticket with Kate, but what it's saying is if there is a non-zero pensioner disregard applied, show the new text. Otherwise don't show this line at all.

patrick-laa (approved):

PR #482, opened 2023-03-06
EL-731: Reshape tests

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-731)

## What changed and why

Restructure the tests as per the test plan to make them more explicit about what they are testing and why they expect what they do.

## Checklist


spec/end_to_end/with_partner_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+    fill_in_add_partner_benefit_screen
+    fill_in_partner_benefits_screen
+    fill_in_partner_other_income_screen(values: { friends_or_family: "200", other: "100" }, frequencies: { friends_or_family: "Every week" })
+    fill_in_partner_outgoings_screen
+    fill_in_partner_property_screen(choice: "Yes, with a mortgage or loan")
+    fill_in_partner_property_details_screen
+    fill_in_partner_vehicle_screen(choice: "Yes")
+    fill_in_partner_vehicle_details_screen
+    fill_in_partner_assets_screen(values: { valuables: "700" })
+    click_on "Submit"
```
starswan, line 89: Yikes - this looks pretty hard to write a new test - it seems to be a backward step from the syntax we had before where only the 'interesting' screens needed to be specified in the test

spec/features/applicant_page_spec.rb
```diff
@@ -1,260 +0,0 @@
-require "rails_helper"
-
-RSpec.describe "Applicant Page" do
-  let(:applicant_header) { I18n.t("estimate_flow.applicant.heading") }
-
-  describe "errors" do
-    before do
```
starswan, line 7: Personally I'm feeling very nervous that we have deleted over 1,000 lines of test just in this block. I'm totally not confident that some important test hasn't been deleted as part of this (and presumably others in this PR)

spec/flows/asylum_and_immigration_flag_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+
+RSpec.describe "Asylum and immigration flag", :controlled_flag, :asylum_and_immigration_flag, type: :feature do
+  it "adds a new matter type screen" do
+    start_assessment
+    fill_in_provider_screen
+    fill_in_level_of_help_screen
+    fill_in_matter_type_screen(choice: "Another legal matter")
+    confirm_screen("applicant")
```
starswan, line 9: There's lots of duplication here. Could it be extracted into common methods or before blocks?

spec/flows/benefits_flow_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+    start_assessment
+    fill_in_provider_screen
+    fill_in_applicant_screen
+    fill_in_dependants_screen
+    fill_in_housing_benefit_screen
+    fill_in_benefits_screen(choice: "Yes")
+    fill_in_add_benefit_screen
+    click_on "Change"
+    fill_in_edit_benefit_screen
+    confirm_screen("benefits")
```
starswan, line 18: see comments passim

spec/end_to_end/with_partner_spec.rb
patrick-laa: @starswan I would argue that here, in these end to end and flow tests (of which there are very few), we are explicitly testing the order of the screens. That's what these tests are about. Whereas in the new form tests, which replace the bulk of the old feature tests, we are now entirely free from implicit assumptions about flow (on the `alt-partner` branch it's proving to be an absolute nightmare to swap around the order of the screens because tests in completely unrelated feature tests keep breaking).

starswan (commented):
I guess someone else from CCQ needs to look at this, but it's enormous. I think it's a shame to lose the methods that go to the check_answers screen with a callback.

spec/end_to_end/with_partner_spec.rb
patrick-laa: I've now updated these tests so that, in the cases where the test is not about explicitly asserting the screens that are filled in, we use a new `fill_in_forms_until(:target_screen_name)` method so that we can skip to the interesting bits.

willc-work (approved):
looks great

PR #485, opened 2023-03-06
[xxx] Update README

HettieS:

## What changed and why

Whilst onboarding I noticed a few steps that might be useful to add to the README. These have been added in this PR.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,25 +14,60 @@ see https://stackoverflow.com/questions/71262775/how-do-i-ensure-assets-are-pres
 
 - Ruby version
 
-  - Ruby version 3.x
+  - Ruby version 3.1.3
   - Rails 7.0.x
 
 - System dependencies
   - postgres
   - yarn
 
-You can install postgres with `brew install postgresql` if you have homebrew. Then please run `bundle exec rails db:create` to set up the development and test databases.
+## Setting up the app
 
-Install dependencies:
+You can use [Homebrew](https://brew.sh/) to install any dependencies. The Homebrew documentation has lots of useful commands.
+
+If using Homebrew:
+
+Install postgres. You will need to select a version if using brew. Run the below command, changing the version number as appropriate.
+
+```
+brew install postgresql@14
+```
+
+You will be prompted on the command line to start the server with something like:
+
+```
+brew services start postgresql@14
+```
+
+If you are running Ruby version 3.1.3, then [Bundler](https://bundler.io/) should already be installed. You may run into an error here if you are not using the correct Ruby version.
 
 ```
 bundle install
+```
+
+Create the development and test databases and run migrations
+
+```
+bundle exec rails db:create
+bundle exec rails db:migrate
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,19 +21,54 @@ see https://stackoverflow.com/questions/71262775/how-do-i-ensure-assets-are-pres
...
+brew install postgresql@14
+```
+
+You will be prompted on the command line to start the server with something like:
 
 ```
+brew services start postgresql@14
+```
+
+Install [Bundler](https://bundler.io/). You may run into an error where you need to update your Ruby version. Use rbenv or another version manager to do this.
```
patrick-laa, line 42: I _believe_ that as of Ruby 2.6, Bundler is built into Ruby, so you don't actually have to do a `gem install bundler` to get bundler to be available for a `bundle install`. That being said, flagging up that you need to have the correct version of Ruby installed - specifically 3.1.3 at the moment. It might be worth updating line 17 to make that explicit.

README.md
```diff
@@ -21,19 +21,54 @@ see https://stackoverflow.com/questions/71262775/how-do-i-ensure-assets-are-pres
...
+
+```
+gem install bundler
 bundle install
+```
+
+Create the development and test databases
+
+```
+bundle exec rails db:create
```
patrick-laa, line 52: You will also need to have the databases loaded up with the correct table structure, via either `bundle exec rails db:migrate` (to apply the migrations one by one), or `bundle exec rails db:schema:load` (to load the schema file). That's probably worth noting (and apologies, I was the one who added a database to the app and didn't update the readme properly)

README.md
```diff
@@ -21,19 +21,54 @@ see https://stackoverflow.com/questions/71262775/how-do-i-ensure-assets-are-pres
...
 yarn build:css
 ```
+
 ## Tests
 
 We test with RSpec and enforce 100% line and branch coverage with SimpleCov.
 
+You can run tests with the command `bundle exec rspec`. 
+
+In order to get tests running, you will need to make sure you have set up the correct environment config. Another developer can pair with you to make sure you have the correct config in your local .env and .env.test files.
```
patrick-laa, line 70: This isn't quite accurate. To get the tests running you shouldn't need anything in an `.env.test` file. But you do need _not_ to have anything in your `.env` file. This has tripped us up in the past, so I think it's worth spelling it out here: `.env` files are kinda dangerous, because they apply to both the development environment and the test environment. Better to copy the pre-populated `.env.sample` to `.env.development`, and you can then monkey with that in development to your heart's content, and then not have either a `.env` or ` .env.test` file.

patrick-laa (commented):
I wholeheartedly approve of this update, as the readme has been rather neglected recently. I've added some comments with a couple of tweaks you might consider.

README.md
HettieS: I think I might have done a `db:reset` now I think about it but I will document the way you've suggested instead!

patrick-laa (approved):

PR #489, opened 2023-03-07
[EL 776] Remove hint text from About their home and bullet from Level of help

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-776)

## What changed and why

We need to remove some hint text from the 'About their home' section for controlled and certificated flows. We also need to remove one of the bullets from the 'Level of help' page.

This PR removes the hint text and the bullet.

## Checklist

### DIFF
--- a/app/views/estimate_flow/property_entry.html.slim
+++ b/app/views/estimate_flow/property_entry.html.slim
@@ -8,7 +8,6 @@
           i18n_key: "property_entry",
           show_smod: true,
           show_joint_ownership: @estimate.partner.present?,
-          show_mortgage: @estimate.property_owned == "with_mortgage",
-          hint: t("estimate_flow.property_entry.second_caption_#{@estimate.level_of_help}")
+          show_mortgage: @estimate.property_owned == "with_mortgage"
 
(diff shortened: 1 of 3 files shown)


app/views/estimate_flow/property_entry.html.slim
```diff
@@ -8,7 +8,6 @@
           i18n_key: "property_entry",
           show_smod: true,
           show_joint_ownership: @estimate.partner.present?,
-          show_mortgage: @estimate.property_owned == "with_mortgage",
-          hint: t("estimate_flow.property_entry.second_caption_#{@estimate.level_of_help}")
```
patrick-laa, line 12: Now that neither of the views that renders the `app/views/estimate_flow/forms/_property_entry.html.slim` partial uses the "hint" param any more, I think it would be a good idea to remove mention of that param from the partial, so keep things tidy.

HettieS: I think I have implemented your feedback @patrick-laa , let me know if I've missed something!

willc-work (approved):
👍

PR #490, opened 2023-03-07
EL-753: Alt partner results page

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-753)

## What changed and why

Update the results page with new order, separate 'household outgoings' section and up to two vehicles sections

## Checklist

### DIFF
--- a/spec/models/calculation_result_spec.rb
+++ b/spec/models/calculation_result_spec.rb
@@ -62,4 +62,30 @@ RSpec.describe CalculationResult do
       expect(described_class.new(data).client_additional_property_rows.keys).to include(:deductions)
     end
   end
+
+  describe "#additional_vehicle_rows" do
+    it "can cope with multiple vehicles" do
+      data = FactoryBot.build(
+        :api_result,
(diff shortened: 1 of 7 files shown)


spec/models/calculation_result_spec.rb
```diff
@@ -62,4 +62,30 @@
...
+
+  describe "#additional_vehicle_rows" do
+    it "can cope with multiple vehicles" do
+      data = FactoryBot.build(
+        :api_result,
+        vehicles: FactoryBot.build_list(:vehicle_api_result, 2),
+      )
+
+      expect(described_class.new(data).additional_vehicle_rows).not_to be_blank
+      expect(described_class.new(data).additional_vehicle_assessed_value).not_to be_blank
```
starswan, line 74: think it would be nice to assert what's in here rather than just not blank

willc-work: would it be better to assert the keys contained in the hash or the actual values associated with the keys? Asserting the keys makes more sense to me, but I could easily be persuaded I am wrong.

PR #495, opened 2023-03-08
EL-730: CFE Service unit tests

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-730)

## What changed and why

Updated the structure of the CfeService code, and added enough unit tests to get us to 100% coverage without relying on higher-level tests.

## Checklist


HettieS:
just in the process of reviewing this - on first look I wonder if it might be nice to move the `session_data` in [spec/services/cfe_service_spec.rb](https://github.com/ministryofjustice/laa-estimate-financial-eligibility-for-legal-aid/pull/495/files#diff-8a29ef1df5154919ed2ba56528cbb23abef98e7c8f130453991d89395f912f4d) into its own fixture file? Just to reduce the file size maybe and make it a little easier to digest

patrick-laa:
@HettieS you're right. My thinking was that it's not good to set specific expectations of output (in this case the payload of calls made to CFE) if you've not been explicit about the input. Test that say "if it feed in whatever the standard fixture content is then I expect this field to end up with a value of exactly 15.427" are rubbish. 

BUT you're right, what I've ended up with is massively unwieldy and indigestible. Which leads me to suspect that I've written the wrong test.

I think the cfe_service_spec should say "If CfeService is called with a full session's worth of data (pulled from a fixture - we don't care about the precise values), then it should result in a bunch of HTTP requests to these specific endpoints, but again, we don't care about precise values. 

Then we break up the expectations about how specific input data is turned into specific output payloads into the unit tests of the individual SubmitXService classes, so that that's done in more manageable chunks.

Leave it with me, I'll rework this and put it back into review when it's ready.

app/services/cfe/submit_proceedings_service.rb
```diff
@@ -0,0 +1,24 @@
...
+      proceeding_type = if estimate.use_legacy_proceeding_type?
+                          applicant_form.legacy_proceeding_type
+                        elsif relevant_form?(:matter_type)
+                          matter_type_form = MatterTypeForm.from_session(@session_data)
+                          matter_type_form.proceeding_type
+                        else
+                          ApplicantForm::PROCEEDING_TYPES[:other]
+                        end
+
+      proceeding_types = [
```
HettieS, line 14: did you move this logic here to make the testing easier?

spec/services/cfe/submit_vehicle_service_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+             in_regular_use: false,
+             loan_amount_outstanding: 4445,
+             subject_matter_of_dispute: true,
+             value: 5556 }],
+        )
+        described_class.call(mock_connection, cfe_assessment_id, session_data)
+      end
+    end
+
+    context "when the client is asylum supported" do
```
HettieS, line 34: should we set a feature flag in this test for asylum and immigration?

HettieS (approved):
Looks great! I've approved but left a couple questions/comments

app/services/cfe/submit_proceedings_service.rb
patrick-laa: Yes, and to make CfeConnection more consistent: in every other case, the connection blindly turns what it's given into the payload of a request. Here it used to have knowledge of what should be _in_ that request, for no good reason.

spec/factories/session.rb
```diff
@@ -0,0 +1,124 @@
+FactoryBot.define do
+  factory :full_session, class: Hash do
```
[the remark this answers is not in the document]
patrick-laa: I think you might be right. I guess the question is what counts as a "basic" session: one where the client is asylum-supported so only has to enter about 3 screens-worth of information? Or one where they've filled out every screen but entered "no"/0 to everything? Or even one where it's not got much in it yet because it's incomplete? I guess there are lots of reasons why a session might have less than every possible value in it. I wonder if ultimately we'd want a fixture for each of them, named for that reason. Definitely open to ways of making the tests more expressive.

spec/services/cfe/submit_assets_service_spec.rb
```diff
@@ -19,6 +67,25 @@
       end
     end
 
+    context "when there are no homes" do
+      let(:session_data) do
```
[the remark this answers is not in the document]
patrick-laa: I do see the merit, but I think in this case, at the level of a unit test I think we want the test to be explicit about what a session with no homes in it looks like. Here, the first 4 key-value pairs of the session data we've specified are very much what the test is about - I wouldn't want to relegate them to a trait. That being said, the final three lines aren't relevant to the test. They just happen to be what you would need to have valid session data to cause the SubmitAssetsService not to also do some other thing here. So maybe this would be best expressed as something like:

```ruby
let(:session_data) do
  FactoryBot.build(:session, 
                   :without_capital_assets, 
                   property_owned: "none",
                   partner_property_owned: "none",
                   property_value: "0,
                   partner_property_value: 0,
 ```
 
 That way the stuff we care about is explicit, the stuff we don't is implicit.
 
 I dunno. I think there's at least one more iteration to do on these tests. Maybe correct use of fixtures is at the heard of it.

spec/services/cfe/submit_vehicle_service_spec.rb
HettieS: I actually just reflected on this - I don't think you need to as you're just giving it the right proceeding type so should just 'work' 🤷‍♀️ - I assume!

patrick-laa: I agree. We actually had a bug last week where a check was started before a feature flag was set and completed after it was set. Part of the CFE logic was making assumptions about whether a flag was _currently_ set, and that broke. So I think we're safer having the services neither know nor care about feature flags if instead they can simply know and care about the contents of the session (and trust that feature flags have determined what gets _into_ the session.

spec/services/cfe/submit_assets_service_spec.rb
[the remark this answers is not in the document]
HettieS: Yeah my choice of trait was a silly one 🤦‍♀️ but definitely! Stuff that doesn't need to be explicit could be potentially handled in a trait/a fixture. It might also help someone reading the tests to work out what kind of data needs to be set and why. That could be nice but probably needs some extra thought around the architecture for that, perhaps something for a future iteration 👍

PR #496, opened 2023-03-09
EL-762: Property tables have SMOD and % owned

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-762)

## What changed and why

Update property tables as per ticket

## Checklist

### DIFF
--- a/app/models/calculation_result.rb
+++ b/app/models/calculation_result.rb
@@ -107,8 +107,9 @@ class CalculationResult
     capital_items(:properties).dig(:main_home, :value)&.positive?
   end
 
-  def client_main_home_rows
-    main_home_rows(prefix: "")
+  def main_home_data
+    main_home = capital_items(:properties, "").fetch(:main_home)
+    property_data(main_home, property_type: :main)
   end
 
   def client_owns_additional_property?
@@ -119,12 +120,12 @@ class CalculationResult
     partner_capital_items(:properties)&.dig(:additional_properties).present?
   end
 
-  def client_additional_property_rows
-    additional_property_rows(prefix: "")
+  def client_additional_property_data
+    additional_property_data(prefix: "")
   end
 
-  def partner_additional_property_rows
-    additional_property_rows(prefix: "partner_")
+  def partner_additional_property_data
+    additional_property_data(prefix: "partner_")
   end
 
   def vehicle_owned?
@@ -175,15 +176,15 @@ class CalculationResult
     }
   end
 
-  def client_assessed_equity
+  def main_home_assessed_equity
     monetise(capital_items(:properties).dig(:main_home, :assessed_equity))
   end
 
-  def client_additional_equity
+  def client_additional_property_assessed_equity
     monetise(capital_items(:properties)[:additional_properties].first.fetch(:assessed_equity))
   end
 
-  def partner_additional_equity
+  def partner_additional_property_assessed_equity
     monetise(partner_capital_items(:properties)[:additional_properties].first.fetch(:assessed_equity))
   end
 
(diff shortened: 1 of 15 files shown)


app/models/calculation_result.rb
```diff
@@ -280,32 +281,46 @@ def capital_rows(prefix:)
...
+      value: property.fetch(:value),
+      mortgage: -property.fetch(:outstanding_mortgage),
+      transaction_allowance: -property.fetch(:transaction_allowance),
+      smod_allowance: -property.fetch(:smod_allowance),
+      main_home_disregard: -property.fetch(:main_home_equity_disregard),
     }
 
-    transaction_allowance = home.fetch(:transaction_allowance)
-    data[:deductions] = monetise(-transaction_allowance) if transaction_allowance.positive?
+    data.each_key do |key|
```
HettieS, line 298: would there ever be any value in creating a new data object with the relevant fields rather than modifying the existing one with `delete` ?

app/models/calculation_result.rb
```diff
@@ -280,32 +281,46 @@ def capital_rows(prefix:)
...
 
-  def additional_property_rows(prefix:)
-    home = capital_items(:properties, prefix)[:additional_properties].first
+  def property_data(property, property_type:)
     data = {
-      value: monetise(home.fetch(:value)),
-      mortgage: monetise(-home.fetch(:outstanding_mortgage)),
+      value: property.fetch(:value),
+      mortgage: -property.fetch(:outstanding_mortgage),
+      transaction_allowance: -property.fetch(:transaction_allowance),
```
HettieS, line 293: general question about `transaction_allowance` - how is this calculated? When is it sent to us by CFE?

app/models/calculation_result.rb
```diff
@@ -107,8 +107,9 @@ def client_owns_main_home?
     capital_items(:properties).dig(:main_home, :value)&.positive?
   end
 
-  def client_main_home_rows
-    main_home_rows(prefix: "")
+  def main_home_data
```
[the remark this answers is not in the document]
patrick-laa: Good question! So for additional properties, the `client_` or `partner_` prefix indicates who _owns_ the additional property. Now, CFE acts as though the client's main home is owned by the client, but CCQ knows that sometimes the client's main home is actually owned by the partner, or even is jointly owned (we just _tell_ CFE it's owned by the client to keep things simple).

So while it is the client's main home, it's not necessarily the client-_owned_ main home, and as such I thought it would be worth dropping the prefix, which does carry those extra connotations.

app/models/calculation_result.rb
patrick-laa: Ok, so broadly what happens is we collect information about the client's finances on each successive form screen, and then, when the user hits 'Submit' on the "Check your answers" screen, we send all the information over to CFE in one go (well, not quite in one go - the CFE API is designed so that we have to call a series of endpoints one by one to pass the data we've collected to it), and CFE then crunches all the numbers and passes back to us a big payload containing the overall result ("eligible"/"ineligible"/"contribution_required") as well as a breakdown of all of the workings. As part of the property calculation, in specific circumstances it applies a 3% transaction allowance to represent the cost of sale - if your home is worth £250k and you sold it, you wouldn't end up with £250k in your bank account thanks to lawyers fees etc. So CFE's "property" payload contains everything to say e.g "You said your home was worth £250k, we've subtracted 3% of that, and we've also subtracted a £100k disregard because it's your main home, so for the purposes of deciding how much capital you have we've decided your property is the equivalent of £142,500 in cash"

app/models/calculation_result.rb
patrick-laa: I buy that in general, immutable objects are more predictable, so in many circumstances yes, that would make sense. But in this case we've only just created `data` a couple of lines ago, and we've created it for the sole purpose of making a return value for this method, so I don't feel bad about monkeying with it.

PR #497, opened 2023-03-09
[EL 789] Show bullet conditionally for Immigration and Asylum

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-789)

## What changed and why

We haven't enabled the Asylum and Immigration flow yet but we display a bullet point about it on the Level of Help page. This caused some user confusion. Hide this bullet unless the feature flag for the asylum and immigration flow is enabled.

## Checklist

### DIFF
--- a/app/views/estimate_flow/level_of_help.html.slim
+++ b/app/views/estimate_flow/level_of_help.html.slim
@@ -19,6 +19,8 @@
     ul.govuk-list.govuk-list--bullet
       - t(".controlled_chambers").each do |bullet|
         li = bullet
+      - if FeatureFlags.enabled?(:asylum_and_immigration)
+        li = t(".asylum_and_immigration_chamber")
 
     h2.govuk-heading-s = t(".certificated_header")
     p.govuk-body = t(".certificated_explanation")
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -515,7 +515,7 @@ en:
       controlled_chambers:
         - the Health, Education and Social Care Chamber of the First-tier Tribunal under the Mental Health Act 1983 or paragraph 5(2) of the Schedule to the Repatriation of Prisoners Act 1984
         - the Mental Health Review Tribunal for Wales
-        - the Immigration and Asylum Chamber of the First tier Tribunal
+      asylum_and_immigration_chamber: the Immigration and Asylum Chamber of the First tier Tribunal
       certificated_header: Civil certificated or licensed legal work
       certificated_explanation: This includes full representation and family help (higher).
       select_a_level: Select a level of help


app/views/estimate_flow/level_of_help.html.slim
```diff
@@ -18,6 +18,8 @@
     p.govuk-body = t(".also_includes")
     ul.govuk-list.govuk-list--bullet
       - t(".controlled_chambers").each do |bullet|
+        - next if bullet.downcase.include?("immigration") && !FeatureFlags.enabled?(:asylum_and_immigration)
```
HettieS, line 21: I'm not sure if there is a better place for this logic rather than the view - but I couldn't see any view_object files to use instead

HettieS: refactored based on my own comment :)

app/forms/level_of_help_form.rb
```diff
@@ -9,4 +9,8 @@ class LevelOfHelpForm
 
   attribute :level_of_help
   validates :level_of_help, presence: true, inclusion: { in: LEVELS_OF_HELP }
+
+  def hide_asylum_and_immigration?(text)
+    text.downcase.include?("immigration") && !FeatureFlags.enabled?(:asylum_and_immigration)
```
patrick-laa, line 14: As it stands this code will break if we add Welsh translations (on the roadmap) before we enable the A&I feature flag. In general, having to know anything about the content of a user-facing text string is probably to be avoided.

In this instance I'd lean towards modifying en.yml so that the iffy bullet has its own key. So then we can do something more like:

```slim
  ul.govuk-list.govuk-list--bullet
    - t(".controlled_chambers").each do |bullet|
      li = bullet
    - if form.show_asylum_and_immigration_content?
      li =  t(".asylum_and_immigration_chamber")
```

HettieS: ah yes good point on the Welsh! Will amend 👍

app/views/estimate_flow/level_of_help.html.slim
patrick-laa: And it's a good refactor! We have been a bit lax about letting quite a bit of logic sneak into the view files. The `form` classes are a better place, but we don't have one of them for every page. Maybe in the next dev team time we could chat about better approaches? I've not used view_objects, but from the name they sound like they might be just the ticket. Perhaps you could walk us through how you've used them before?

patrick-laa (approved):
Looks good!

app/views/estimate_flow/level_of_help.html.slim
HettieS: I've just switched the implementation back to having the conditional with the feature flag in the view - chatted with @willc-work and we decided that is probably easier for the time being. But definitely up for chatting about how we can move the logic elsewhere/into a view object - they're quite nice if you know you're going to need to manipulate the UI a bit more

PR #503, opened 2023-03-10
EL-733: Controlled matter type content

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-733)

## What changed and why

Add new content for matter type screen for controlled work.

## Checklist

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -541,12 +541,11 @@ en:
       tribunal_guidance_link: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/983065/Lord_Chancellor_s_guidance_on_determining_controlled_work_and_family_mediation.pdf
     matter_type:
       title: Which type of matter is this?
-      immigration_controlled: Immigration in the First-tier Tribunal
-      asylum_controlled: Asylum in the First-tier Tribunal
-      other_controlled: Another legal matter
+      immigration_controlled: Legal representation for an immigration matter (First-tier Tribunal)
+      asylum_controlled: Legal help for an immigration or asylum matter, or help at court or legal representation for an asylum matter (First-tier Tribunal)
       immigration_certificated: Immigration (Upper Tribunal)
       asylum_certificated: Asylum (Upper Tribunal)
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -560,14 +559,26 @@ en:
...
+      immigration_and_asylum_controlled_paragraph: These must be for matters in the Immigration and Asylum Chamber of the First-tier Tribunal.
+      immigration_header_controlled: Legal representation for an immigration matter
+      immigration_controlled_paragraph_one: This must be for legal representation for an immigration matter in the Immigration and Asylum Chamber of the First-tier Tribunal.
+      immigration_controlled_paragraph_two_html: Guidance for this is in paragraphs 25 to 29, 32(1), 32A(1) of Part 1 of Schedule 1 to the <a href='https://www.legislation.gov.uk/ukpga/2012/10/schedule/1' target='_blank' rel='noreferrer noopener'>Legal Aid, Sentencing and Punishment of Offenders Act 2012</a> (LASPO).
+      asylum_header_controlled: Legal help for an immigration or asylum matter, or help at court or legal representation for an asylum matter
+      asylum_controlled_paragraph_one: "This must be for a matter in the Immigration and Asylum Chamber of the First-tier Tribunal, covering:"
```
HettieS, line 575: should these quotation marks be here?

HettieS (commented):
One small question about the quotation marks before I approve this one

config/locales/en.yml
patrick-laa: Very good question, and the answer is yes, and the reason is because YAML isn't as smart as we'd like. YAML is generally smart enough to know that `foo: bar` means `bar` is a string, i.e. `"bar"`. Because of that we generally write `foo: bar` instead of the more explicit `foo: "bar"` because it's less punctuation to type.

BUT where strings contain special characters like `:`, YAML isn't sure if the strings are strings or if they're actually something else, so it freaks out.

So that's a long way of saying, because of the `:` at the end of the line, we have to wrap the text in quotation marks!

HettieS (approved):
👍

PR #511, opened 2023-03-13
EL-771: Remove SMOD UI when it doesn't apply

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-771)

## What changed and why

Upper tribunal cases do not consider SMOD

## Checklist


app/services/cfe/submit_assets_service.rb
```diff
@@ -31,15 +32,16 @@ def call(cfe_assessment_id)
           value: property_entry_form.house_value,
           outstanding_mortgage: (property_entry_form.mortgage if property_form.owned_with_mortgage?) || 0,
           percentage_owned:,
+          subject_matter_of_dispute: (property_entry_form.house_in_dispute && !estimate.upper_tribunal?) || false,
```
HettieS, line 35: because we have `!estimate.upper_tribunal?` in so many places, I wonder if it's worth pulling it into its own method? Or own module?

app/views/estimate_flow/property_entry.html.slim
```diff
@@ -6,7 +6,7 @@
 .govuk-grid-column-two-thirds
   = render "estimate_flow/forms/property_entry",
           i18n_key: "property_entry",
-          show_smod: true,
+          show_smod: !@estimate.upper_tribunal?,
```
HettieS, line 9: are `show_smod` and `show_smod_content` different in any way? If not maybe we could take this opportunity to make that consistent

spec/factories/session.rb
HettieS, line 1: nice traits!

HettieS (commented):
Nice! Just a couple questions

app/forms/client_property_entry_form.rb
```diff
@@ -4,7 +4,6 @@ class ClientPropertyEntryForm < BasePropertyEntryForm
   attr_accessor :partner
 
   attribute :house_in_dispute, :boolean
-  validates :house_in_dispute, inclusion: { in: [true, false] }, allow_nil: false
```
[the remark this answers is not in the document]
patrick-laa: You're right. I was shying away from it because I think to do it properly requires a bit of a refactor that risked a can of worms, but on balance it's worth doing.

HettieS (approved):
Looks great! Nice refactor. It would be good to catch up on how the `SessionPersistable` stuff works with the estimate and various forms as not sure I entirely follow the code - but the tests are passing so all looks great! 👍

app/services/cfe/submit_vehicle_service.rb
```diff
@@ -4,7 +4,8 @@ def call(cfe_assessment_id)
       return unless relevant_form?(:vehicle_details)
 
       details_model = ClientVehicleDetailsForm.from_session(@session_data)
-      vehicles = CfeParamBuilders::Vehicles.call(details_model, in_dispute: details_model.vehicle_in_dispute)
+      vehicles = CfeParamBuilders::Vehicles.call(details_model,
+                                                 in_dispute: details_model.vehicle_in_dispute && estimate.smod_applicable?)
```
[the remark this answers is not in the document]
patrick-laa: Only that, because it's only called once in this class, and is called explicitly rather than being inferred from a symbol representing the name, it would have involved more typing to add a `delegate` line than not. I don't think it really matters either way, but for me conciseness won out.

PR #517, opened 2023-03-16
EL-805: Collect new user journey data and push derived metrics to Geckoboard

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-805)

## What changed and why

Store summary information for each check completed on the service, and push associated metrics to Geckoboard as part of our regular cron job. 

## Checklist


app/services/journey_logger_service.rb
```diff
@@ -0,0 +1,31 @@
...
+      end
+    rescue StandardError => e
+      ErrorService.call(e)
+    end
+
+    def build_attributes(calculation_result, estimate)
+      {
+        certificated: !estimate.controlled?,
+        partner: estimate.partner || false,
+        person_over_60: estimate.over_60 || (estimate.partner && estimate.partner_over_60) || false,
```
willc-work, line 21: could we just have `estimate&.partner_over_60` in the first OR section?

willc-work: and the same with `partner_vehicle_owned?`

willc-work (approved):
one query, not a blocker

app/services/journey_logger_service.rb
patrick-laa: Annoyingly no - because of the edge case where the user has said the client has a partner over 60 then gone back and said the client doesn't have a partner after all. In that case there is no person over 60, even though estimate.partner_over_60 will be `true`.

PR #519, opened 2023-03-17
EL-786 hide irrelevant rows from vehicle results

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-786)

## What changed and why

**Before**
If a vehicle was _not_ in regular use (only possible for certificated checks), the `assessed value` of the vehicle was equal to the `value.` Outstanding finance was not deducted, and the “disregards and deductions” figure we got back from CFE was actually the outstanding finance figure multiplied by -1. 

**After**
We now show neither of those rows in the vehicle table when `in_regular_use` is false in the vehicle payload.

For vehicles not in regular use, the vehicle table shown on the results page only has the “Value” and “Assessed value” rows.

## Checklist


patrick-laa:
(I'm not going to review this one as I paired writing the code so have biased eyes. I'll leave it to Hettie or Will.)

PR #521, opened 2023-03-17
EL-809 Change to dependant & partner allowances: 10/04/23

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-809)

## What changed and why

Made changes to allowance text displayed for both Controlled and Certificated flows, as the rates have changed to coincide with the start of the financial year. These changes will take effect from 10/04/23.

## Checklist


app/lib/feature_flags.rb
```diff
@@ -1,6 +1,7 @@
 class FeatureFlags
   ENABLED_AFTER_DATE = {
     example_2125_flag: "2125-01-01",
+    el_809: "2023-04-10",
```
willc-work, line 4: a more descriptive name would be better, such as dependance_allowance_increase, so we know what it is without needing to reference the jira ticket

app/views/estimate_flow/forms/_dependant_details.html.slim
```diff
@@ -1,3 +1,6 @@
+- dependant_allowance = FeatureFlags.enabled?(:el_809) ? "new" : "old"
```
willc-work, line 1: i would suggest using legacy instead of old, just ti be consistent with how we have handled these changes in the past. certainly not a blocker though if you want to keep it like this, as ultimately it will be removed

willc-work (changes requested):
I think the flag name should be changed to something more descriptive, the other commnet is just a suggestion but i dont feel too strongly about it

app/views/estimate_flow/forms/_dependant_details.html.slim
willc-work: I should have been more clear here, if we use legacy instead of old and dont enter anything for 'new' so it looks like this:
FeatureFlags.enabled?(:el_809) ? "" : "legacy"

This way we just need to remove the legacy translations at some point and then there is no need to change the translations to remove the "new" part. It just makes removing feature flag items easier.

patrick-laa: Although we'd need to be a little careful here, because a blank string here would mean the translation key ended up being "controlled_income_exclusion_" with a dangling underscore at the end, which is not ideal. 

You could do:

```
suffix = FeatureFlags.enabled?(:el_809) ? "" : "_legacy"
```

And then on the next line remove the underscore i.e.:

```
income_exclusion_key = "#{@estimate.level_of_help}_income_exclusion#{suffix}"
```

PR #551, opened 2023-03-28
EL-818: Extra ineligibility line on results page

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-818)

## What changed and why

Add missing ineligibility explanation at bottom of tables on results page

## Guidance to review

The CFE response shows whether each section (gross income, disposable income, capital) is ineligible by associating an "ineligible" result string with one or more of the proceeding types that are attached to that section. But in CCQ there is only ever one proceeding type, because we only ever feed CFE one proceeding type for a check.

## Checklist

### DIFF
--- a/app/models/calculation_result.rb
+++ b/app/models/calculation_result.rb
@@ -27,6 +27,10 @@ class CalculationResult
     api_response.dig(:result_summary, section, :proceeding_types).none? { _1[:result] == "pending" }
   end
 
+  def ineligible?(section)
+    api_response.dig(:result_summary, section, :proceeding_types).all? { _1[:result] == "ineligible" }
+  end
+
   def capital_contribution
     monetise(api_response.dig(:result_summary, :overall_result, :capital_contribution))
   end
--- a/app/views/estimates/_ineligible_explanation.html.slim
+++ b/app/views/estimates/_ineligible_explanation.html.slim
@@ -1,0 +1,8 @@
+- if @model.ineligible?(value_type)
+  - subject = t("estimates.show.ineligible_explanation.#{@model.has_partner? ? 'subject_with_partner' : 'subject_no_partner'}")
+  - outcome = t("estimates.show.ineligible_explanation.#{@model.has_partner? ? 'outcome_with_partner' : 'outcome_no_partner'}")
+  p.govuk-body class="govuk-!-margin-top-8"
+    = t("estimates.show.ineligible_explanation.text",
+        subject:,
+        outcome:,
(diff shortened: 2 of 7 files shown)


app/models/calculation_result.rb
```diff
@@ -26,6 +26,10 @@ def calculated?(section)
     api_response.dig(:result_summary, section, :proceeding_types).none? { _1[:result] == "pending" }
   end
 
+  def ineligible?(section)
+    api_response.dig(:result_summary, section, :proceeding_types).all? { _1[:result] == "ineligible" }
```
HettieS, line 30: interesting! So would CFE ever return eligible for some sections and ineligible for others? And is the result always present for each section? I guess I'm wondering how this looks to the user and how CFE determines the change to "ineligible"...

HettieS (approved):
Approving this although I'm intrigued how this result gets calculated per section from CFE

app/models/calculation_result.rb
patrick-laa: CFE will go section by section, and as soon as the result for a section is "ineligible" it will stop and return the other sections as "pending". So you might get gross income as "eligible", disposable income as "ineligible" and capital as "pending", with an overall result of "ineligible". We've got logic (see the method above this one) to skip any sections that are "pending" in the UI.

HettieS: Nice, yes I have just spotted the "pending" thing! So this ineligibility sentence will only ever appear once. Got it! Do we think this will change with the one shot API?

patrick-laa: As it stands the one-shot API still uses the same bit of logic that short-circuits the eligibility check as soon as any section is ineligible. Changing that behaviour would be a breaking change, and the promise for the one-shot API is that it has the same behaviour, just in a different form factor. So I think we're save to assume this will continue to work going forward.

app/views/estimates/_ineligible_explanation.html.slim
```diff
@@ -0,0 +1,6 @@
+- if @model.ineligible?(value_type)
```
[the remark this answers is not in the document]
patrick-laa: If the client selected domestic abuse, CFE will never tell us that they are ineligible. It sets the upper thresholds for income and capital on certificated domestic abuse cases infinitely high, so for a domestic abuse case the result for any section will always be either "eligible" or "contribution_required" (it's just that if you're very wealthy, that contribution will be very, very high.)

PR #554, opened 2023-03-30
LEP-77 - Point CCQ at new validation end point

starswan:
https://dsdmoj.atlassian.net/browse/LEP-77

Temp draft PR for testing LEP-77


davidread:
Not needed any more

PR #562, opened 2023-04-03
EL-826 Removed trailing whitespace: outstanding_mortgage field

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-826)

## What changed and why

Relates to this [Runtime Error](https://ministryofjustice.sentry.io/issues/4058281497/?project=4504177640275968&referrer=slack) where if a user leaves a trailing white space after the number on the 'outstanding mortgage' field they can't see the results page.

## Guidance to review

Chosen to tidy the data we send to CFE instead of form validation, for a better user journey. 

## Checklist

### DIFF
--- a/app/services/cfe/submit_assets_service.rb
+++ b/app/services/cfe/submit_assets_service.rb
@@ -16,7 +16,7 @@ module Cfe
       if asset_form.property_value.positive?
         second_property = {
           value: asset_form.property_value,
-          outstanding_mortgage: asset_form.property_mortgage.strip,
+          outstanding_mortgage: asset_form.property_mortgage.to_s.strip.to_f,
           percentage_owned: asset_form.property_percentage_owned,
           subject_matter_of_dispute: (asset_form.property_in_dispute? && smod_applicable?) || false,
         }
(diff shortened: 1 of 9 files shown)


MazOneTwoOne:
This bug, where a user leaves a trailing white space, may also apply to other number fields.

app/services/cfe/submit_assets_service.rb
```diff
@@ -16,7 +16,7 @@ def call(cfe_assessment_id)
       if asset_form.property_value.positive?
         second_property = {
           value: asset_form.property_value,
-          outstanding_mortgage: asset_form.property_mortgage,
+          outstanding_mortgage: asset_form.property_mortgage.to_s.strip.to_f,
```
willc-work, line 19: might be worth testing to see if `to_f `will work on  its own. According to the docs it should.
https://apidock.com/ruby/v2_5_5/String/to_f

HettieS:
Nice spot on this! I wonder if we want to consider stripping the data in the form rather than at the point of sending the data? I wonder if it might allow us to have some clearer, more shared logic if we do it across the forms?

MazOneTwoOne:
- Early attempt to fix bug, when it was assumed it affected only 1 monetary filed. 
- Changed values to existing tests from string to integers. 
- Applied data validation when sending the data to CFE instead of the form attribute.

PR #563, opened 2023-04-04
update cast method to remove spaces from string values

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-826)

updated the cast method to remove spaces entered into monetary value fields

## Checklist


spec/features/number_fields_spec.rb
```diff
@@ -4,9 +4,9 @@
 INVALID_INTEGER_VALUES = ["two", "2 00", "2.37"].freeze
 INVALID_MONEY_VALUES = ["two", "2 00", "$54"].freeze
 VALID_MONEY_VALUES = [
-  ["2,000", "2,000"],
+  ["2,000", "2,000", "2,000 "],
```
exonian, line 7: What does adding this value do? I can't see where the test actually uses the third item in these. My understanding is that the first is an input value and the second is how it is displayed when you return to the form.

Additionally, because it tests what's loaded back into the form when you revisit the page, this is the result of first using `cast` to serialize the data and then `decimal_as_money_string` to deserialize it. Is there a way to test what's on the model directly, rather than once it's been through the display method? It's not vital here, I suppose, as display isn't putting trailing spaces back, but I think this would be better as a unit test, or at least closer to one.

patrick-laa, line 7: If you look at how VALID_MONEY_VALUES gets used in the tests, only the first and second array elements of each sub-array ever get read. The elements you've added will be ignored.

What you want to do is add a new sub-array with the first (what gets written) having a space, and the second (what is expected to be read) not having one.

And since we've discovered that the formatting here is confusing, maybe worth changing these sub-arrays to hashes with `input` and `output` as the keys?

config/initializers/attribute_types.rb
```diff
@@ -1,9 +1,9 @@
 class Gbp < ActiveModel::Type::Decimal
   def cast(value)
-    if value.is_a?(Numeric) || value.blank? || value.gsub(/[0-9,.£]/, "").empty?
+    if value.is_a?(Numeric) || value.blank? || value.gsub(/[0-9,.£]+\s?\z/, "").empty?
       # When a monetary value is provided by a form submission, we are only interested
       # in the value to 2 decimal places, and discard any more fine detail immediately
-      super(value&.to_s&.delete(",")&.delete("£"))&.round(2)
+      super(value&.to_s&.chomp&.delete(",")&.delete("£"))&.round(2)
```
patrick-laa, line 6: `chomp` doesn't remove trailing spaces, it removes newlines. I think you're thinking of `strip`?

patrick-laa (changes requested):

config/initializers/attribute_types.rb
patrick-laa: But also, you don't actually need to make this change at all, because whitespace gets stripped by the superclass `cast` method anyway.

config/initializers/attribute_types.rb
```diff
@@ -1,9 +1,9 @@
 class Gbp < ActiveModel::Type::Decimal
   def cast(value)
-    if value.is_a?(Numeric) || value.blank? || value.gsub(/[0-9,.£]/, "").empty?
+    if value.is_a?(Numeric) || value.blank? || value.gsub(/[0-9,.£]+\s?\z/, "").empty?
```
patrick-laa, line 3: This will only work for a single space at the end of the string. Multiple spaces, or tabs, or whitespace at the beginning of the string, will suffer the same problem. So how about instead you do:

```ruby
value.is_a?(Numeric) || value.blank? || value.gsub(/[0-9,.£]/, "").blank?
```

spec/features/number_fields_spec.rb
patrick-laa: On this second point, you could either create a unit test of a `Form` object with a `gbp` attribute or, as a half-way house, you could modify these tests with an additional expectation about `session_contents`: i.e. get this test to expect that the value gets stored in the session as a decimal, as well as expecting it to be displayed back to the user as a formatted string.

exonian: Cool yeah, either of those sound reasonable.

config/initializers/attribute_types.rb
exonian: Oh so the thing that needs to change is making something like "123 " reach this line at all, which it currently isn't because of the regex in the `if` statement?

patrick-laa: Exactly. The assumption was that the 'if' statement would only evaluate `false` in situations where validation also failed, so that we could show a validation message and prepopulate the form with the invalid value provided by the user exactly as entered. The bug arises where the if statement is falsey but the validation _passes_, leading to the value exactly as entered being stored in the session and consequently being passed on to CFE.

config/initializers/attribute_types.rb
willc-work: this lets you enter `123 000`, which is accepted but is converted to `123` so we get inaccurate results. We(ruby?) are already removing whitespace at the start of any value that is entered.

willc-work: i have amended the regex to remove all trailing whitespace instead

patrick-laa: If you enter `  £20000` as it stands that will still cause a string rather than a decimal to be stored in the session, leading to a CFE crash. So I think we really do need to deal with leading whitespace.  Maybe it's `value.strip.gsub(/[0-9,.£]/, "").empty?`, because that will return true for leading and trailing whitespace but false for middle-of-string whitespace.

spec/features/number_fields_spec.rb
```diff
@@ -61,12 +88,13 @@
     end
 
     VALID_MONEY_VALUES.each do |pair|
-      it "allows me to enter '#{pair[0]}'" do
-        fill_in "employment-form-national-insurance-field", with: pair[0]
+      it "allows me to enter '#{pair[:input]}'" do
+        fill_in "employment-form-national-insurance-field", with: pair[:input]
         click_on "Save and continue"
+        expect(session_contents["national_insurance"]).to eq pair[:input].to_s.delete(",").delete("£").to_f
```
patrick-laa, line 94: You're right, @willc-work , this is not hugely helpful. Try adding a new `stored` key to each of the test hashes, e.g:

```ruby
{
  input: "  £2000",
  output: "2,000",
  stored: 2000.0,
}

...

expect(session_contents["national_insurance"]).to eq pair[:stored]

```

patrick-laa (approved):

app/forms/base_assets_form.rb
```diff
@@ -2,6 +2,9 @@ class BaseAssetsForm
   include ActiveModel::Model
   include ActiveModel::Attributes
   include SessionPersistable
+  include NumberValidatable
+
+  validate :numbers_are_all_valid
```
patrick-laa, line 7: You don't actually need to specify this in each Form file, because the `included` block in NumberValidatable means that this validation gets automatically declared by anything that includes it.

app/forms/dependant_details_form.rb
```diff
@@ -13,6 +16,7 @@ class DependantDetailsForm
   FORM_ATTRIBUTES.each do |boolean_field, integer_field|
     attribute boolean_field, :boolean
     validates boolean_field, inclusion: { in: [true, false] }
+    # validates_format_of integer_field, with: /\A\d*\.?\d+\z/, message: "%{value} is not a valid number"
```
patrick-laa, line 19: Let's remove this.

spec/features/number_fields_spec.rb
```diff
@@ -1,14 +1,49 @@
 require "rails_helper"
 
-VALID_INTEGER_VALUES = ["2", "2,000"].freeze
+VALID_INTEGER_VALUES = ["2", "2,000", "1234   "].freeze
 INVALID_INTEGER_VALUES = ["two", "2 00", "2.37"].freeze
```
patrick-laa, line 4: Let's add some invalid integer and money values based on what we found in testing today? So things like "-20", "+20", "10_000" and "123.45e1" are the things me and Mas found.

app/forms/outgoings_form.rb
```diff
@@ -2,6 +2,9 @@ class OutgoingsForm
   include ActiveModel::Model
   include ActiveModel::Attributes
   include SessionPersistable
+  include NumberValidatable
```
patrick-laa, line 5: Also, I think we need to include this in `BenefitModel`, as that has a :gbp attribute as well.

patrick-laa (approved):

PR #569, opened 2023-04-11
EL-816: Check model

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-816)

## What changed and why

The 'EstimateModel' class had 3 things wrong with it:

1) CCQ doesn't deal in 'Estimates'; we deal in 'Checks'
2) It only provided access to a subset of the attributes associated with a specific check
3) For the attributes it gave access to, it didn't consider whether that attribute was relevant. E.g if the user said the client had a partner, then later said the client received asylum support, the model would return `true` for the partner attribute even though when a client received asylum support the question of whether they have a partner is no longer valid.

This PR tries replacing EstimateModel with a new model called `Check` which fixes these three defects.

## Guidance to review

This work comes out of EL-816 and in particular tries to solve the "I've used the 'back' button and changed an answer to question A that renders my previous answer to question B irrelevant" problem. I think this will be helpful when we come to fill in CW PDFs because we won't accidentally fill in attributes that weren't used in the eligibility calculation.

StepsHelper now gets handed raw session data rather than an EstimateModel, because the new replacement for the EstimateModel relies on StepsHelper to function, so we needed to avoid a circular dependency.

## Checklist


README.md
```diff
@@ -155,6 +155,21 @@ We use Grover to save pages as PDF files for download, which in turn uses Puppet
...
+**NavigationHelper** contains methods that directly interrogate a session object for a few specific attributes that affect navigation flow through the form.
+It knows how answers to certain questions affect the relevance of certain other questions.
+
+**StepsHelper** uses NavigationHelper to determine which screens, or steps, should be displayed for a given check, based on the answers provided so far.
+
+**Flow::Handler** knows, for any given step, which Form object to populate to back the HTML form displayed on screen
+
+**Check** provides access to all _relevant_ data for a given check. For any given attribute it uses StepsHelper and Flow::Handler to determine whether,
+given the other answers given, the attribute is relevant. If not, when asked for that attribute it will return `nil`. Otherwise it will return that attribute.
+
```
MazOneTwoOne, line 172: This is super helpful.

patrick-laa:
@HettieS the problem re your comment 

> I'm wondering if that new helper is needed or if the model could handle the boolean logic

is just that all of the low-level boolean logic in NavigationHelper is needed by StepsHelper, which is in turn needed to get Check to function at all. Because of those dependencies I can't reference Check _within_ StepsHelper.

I agree that NavigationHelper isn't a hugely helpful name. I've renamed it to StepsLogic to emphasise that it's really a bunch of low-level methods needed for StepsHelper to work (that also just happen to sometimes be useful elsewhere).

README.md
```diff
@@ -155,6 +155,21 @@ We use Grover to save pages as PDF files for download, which in turn uses Puppet
...
+**StepsLogic** contains methods that directly interrogate a session object for a few specific attributes that affect navigation flow through the form.
+It knows how answers to certain questions affect the relevance of certain other questions.
+
+**StepsHelper** uses StepsLogic to determine which screens, or steps, should be displayed for a given check, based on the answers provided so far.
+
+**Flow::Handler** knows, for any given step, which Form object to populate to back the HTML form displayed on screen
+
+**Check** provides access to all _relevant_ data for a given check. For any given attribute it uses StepsHelper and Flow::Handler to determine whether,
+given the other answers given, the attribute is relevant. If not, when asked for that attribute it will return `nil`. Otherwise it will return that attribute.
```
willc-work, line 171: pedantic i know but the repetition of given, perhaos the second instance could we switched to 'provided', 'entered' or 'submitted'

app/controllers/benefits_controller.rb
```diff
@@ -3,12 +3,12 @@ class BenefitsController < EstimateFlowController
 
   def new
     @model = model_class.new
-    @estimate = load_estimate
+    @check = load_check
```
willc-work, line 6: would this file benefit from a` before_action :load_check` that omitted `:add`?

README.md
patrick-laa: Very fair point!

app/lib/steps_logic.rb
```diff
@@ -0,0 +1,68 @@
...
+    def upper_tribunal?(session_data)
+      session_data["proceeding_type"].in?(MatterTypeForm::PROCEEDING_TYPES.slice(:immigration, :asylum).values)
+    end
+
+    def asylum_supported?(session_data)
+      upper_tribunal?(session_data) && session_data["asylum_support"]
+    end
+
+    def passported?(session_data)
+      !asylum_supported?(session_data) && session_data["passporting"]
```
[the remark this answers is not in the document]
willc-work: that makes sense, so it is a belt and braces approach to ensure that passported really is passported.

PR #571, opened 2023-04-11
EL-651: Pre-fill CW1

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-651)

## What changed and why

- We add PDFTK as a dependency in CircleCI and  our Dockerfile, with instructions for how to download it for local development
- We use pdf-forms to populate a CW1 form when the user requests one, using a new ControlledWorkDocumentPopulationService
- A new ControlledWorkDocumentContent model exposes all the values needed to populate the form
- ControlledWorkDocumentValueMappingService reads a yaml file that specifies how to map fields in the CW1 form to appropriate methods on the model

## Guidance to review

- I'm building this PR off #569 as that provides a handy model that lets me access session attributes without having to worry if they're redundant/irrelevant
- The idea is that for other controlled forms we just need to add a new PDF file in `lib`,  then add a new entry in `ControlledWorkDocumentPopulationService::TEMPLATES` so the service knows where to find the file, and finally a new YAML mapping file (the assumption is that the arbitrary-seeming names of fields in CW1 form won't be exactly duplicated in other forms even if the content of the fields is the same, so each form will need a unique mapping; I am OK with being pleasantly surprised on that front).

## Checklist


app/services/controlled_work_document_population_service.rb
```diff
@@ -0,0 +1,47 @@
...
+      when "checkbox"
+        mapping[:checked_value] if value(mapping, check, calculation_result)
+      when "text"
+        value(mapping, check, calculation_result)
+      end
+    end
+
+    def value(mapping, check, calculation_result)
+      case mapping[:source]
+      when "answers"
```
HettieS, line 38: i wonder if 'user' as the source might be clearer? As the data is user inputted

app/services/controlled_work_document_population_service.rb
```diff
@@ -0,0 +1,47 @@
...
+        check.send(mapping[:attribute])
+      when "result"
+        calculation_result.dig(*mapping[:attribute].split("."))
+      when "static"
```
HettieS, line 42: is 'static' the source when we just want to populate the name of the field as opposed to any data?

HettieS (commented):
Nice!

app/services/controlled_work_document_population_service.rb
```diff
@@ -0,0 +1,47 @@
...
+      check = Check.new(session_data)
+      calculation_result = session_data["api_response"]
+      mappings = YAML.load_file(Rails.root.join("app/lib/controlled_work_mappings/#{form_type}.yml")).with_indifferent_access
+      mappings[:mappings].map { [_1[:name], convert(_1, check, calculation_result)] }.to_h
+    end
+
+    def convert(mapping, check, calculation_result)
+      case mapping[:type]
+      when "checkbox"
+        mapping[:checked_value] if value(mapping, check, calculation_result)
```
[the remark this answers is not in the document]
patrick-laa: Yup, at the moment this is rendering all "static" checkboxes as checked. Just to help me sanity check that I have control over all the checkboxes even if I'm not yet sure what they're all called. As above, the "static" stuff is just scaffolding while I figure out the names of each field.

app/services/controlled_work_document_population_service.rb
```diff
@@ -0,0 +1,47 @@
...
+    }.freeze
+
+    def template_path(form_type)
+      Rails.root.join(TEMPLATES.fetch(form_type))
+    end
+
+    def values(session_data, form_type)
+      check = Check.new(session_data)
+      calculation_result = session_data["api_response"]
+      mappings = YAML.load_file(Rails.root.join("app/lib/controlled_work_mappings/#{form_type}.yml")).with_indifferent_access
```
HettieS, line 23: Nice. Am I right in thinking we can just have one YAML file with the fields that are across all forms?

patrick-laa: I think it depends a _lot_ on what we find when we start exploring the CW2 forms etc. If the fields all have the same names then we can have one YAML file. But if they have slightly different field names then I think we have a YML file for each of them. But that's an EL-798 problem, I think!

HettieS: Makes sense 😄 the only other thing I was thinking around the mappings is rather than YAML, maybe doing some sort of hash mapping instead like:

```
module ControlledWork
  class Mappings
    CHECKBOX = {
      Checkbox_box1: some_value,
      Checkbox_box2: some_value2,
      Checkbox_box3: some_value3,
    }.freeze

    TEXT = {
      undefined_10: some_attribute,
      undefined_11: some_attribute11,
      undefined_12: some_attribute12,
    }.freeze
  end
end

ControlledWork::Mappings:CHECKBOX.map do |pdf_key, attribute|
  some things to make it return the right output for the pdf
end
```
which might be a completely terrible idea but just wanted to throw it out there. The yml file looks good though .

HettieS: I think what you've done looks great, that yml file looks nice and clear now, and easy to amend. This looks like a good way of building the forms out to me!

app/models/controlled_work_document_content.rb
```diff
@@ -0,0 +1,206 @@
...
+  def client_non_employment_income
+    non_employment_income
+  end
+
+  def partner_non_employment_income
+    non_employment_income("partner_")
+  end
+
+  def client_mortgage
+    main_home_owned? ? net_housing_costs : 0
```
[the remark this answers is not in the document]
patrick-laa: So this one comes from the requirements in the spreadsheet which says: "Any field not populated should contain 0."

It's worth saying that due to the `skip_unless: client_income_relevant?` line in the config file, this zero will only appear if the client _has_ entered housing costs, it's just that we've inferred that these costs are rent rather than mortgage payments. If the client receives passporting benefits or asylum support, this whole section will be moot so `client_income_relevant?` will be false so this field will be skipped and left blank.

app/services/controlled_work_document_value_mapping_service.rb
```diff
@@ -0,0 +1,50 @@
+# This file uses the rules defined by an array of "mappings" to generate a hash with
+# values derived from the session_data passed in.
+
+# For the most part it does this by calling methods on a ControlledWorkDocumentContent
+# model.
+
+# The mappings have the following attributes:
+# name - the name of a field in the PDF, which should be a key of the output hash
+# type - what sort of field this is (either `text`,  `checkbox` or `boolean_radio` (i.e. "Yes" and "No" radio buttons))
+# value_type - where to get the value from (either `from_attribute` or `from_cfe_payload`)
```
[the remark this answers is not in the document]
patrick-laa: Oh yes, that's much clearer. Will change!

app/models/controlled_work_document_content.rb
```diff
@@ -0,0 +1,206 @@
...
+
+  def additional_non_smod_properties_value
+    additional_non_smod_properties_sum "value"
+  end
+
+  def additional_non_smod_properties_mortgage
+    additional_non_smod_properties_sum "outstanding_mortgage"
+  end
+
+  def additional_non_smod_properties_percentage_owned
```
willc-work, line 81: My concern with returning a blank is that we will forget about this method and it will look like a bug to a user, i.e they enter something relevant that is not subsequently included on the form. Would it be better to highlight somewhere on the form that we cannot display due to [reason], instead of appearing to ignore it

app/lib/controlled_work_mappings/cw1.yml
```diff
@@ -0,0 +1,495 @@
...
+# combined with the output from CFE.
+
+# The mappings conform to the format described in ControlledWorkDocumentValueMappingService
+
+# "The client is directly or indirectly in receipt of S.4 or S.95 Asylum Support (Immigration and Asylum category work only):"
+- name: Go to question 2
+  type: boolean_radio
+  source: from_attribute
+  attribute: asylum_support
+  yes_value: "Yes_2"
```
willc-work, line 11: I was a bit confused about the yes_value/no_value entries, but I realised that these suffixes like "_2" were  probably used to identify where on a page these were appearing. As such I assume that all suffixes on yes/no answers need to be removed now.

patrick-laa: @willc-work sorry, I should probably have documented this better. For form radio buttons, each button in the set has its own value. To mark that button as selected you have to assign that value to the radio set. So in this case, if you want the 'Yes' button to be checked for the "Does the client receive asylum support?" question, you have to assign the value "Yes_2" to the set called "Go to question 2", because "Yes_2" happens to be the value associated with the "Yes" button in the "Go to question 2" set.

app/models/controlled_work_document_content.rb
patrick-laa: That's a really good idea. Maybe it should say "Unknown" or something in the field?

willc-work (approved):

app/lib/controlled_work_mappings/cw1.yml
```diff
@@ -0,0 +1,495 @@
...
+  attribute: partner_investments
+
+# Valuables (non-SMOD)
+- name: undefined_42
+  type: text
+  source: from_attribute
+  attribute: valuables
+  skip_if: valuables_in_dispute?
+
+# Valuables (non-SMOD)
```
HettieS, line 265: Partner valuables - just spotted this whilst doing a bit of copy pasta

patrick-laa: Good spot! Fixed.

app/lib/controlled_work_mappings/cw1.yml
```diff
@@ -0,0 +1,495 @@
...
+  skip_unless: partner_income_relevant?
+
+# Standard allowance for employment expense / Client
+- name: undefined_64
+  type: text
+  source: from_attribute
+  attribute: client_employment_deduction
+  skip_unless: client_income_relevant?
+
+# Standard allowance for employment expense / Client
```
HettieS, line 419: partner, not client

patrick-laa: 🙏 thanks!

app/lib/controlled_work_mappings/cw1.yml
```diff
@@ -0,0 +1,495 @@
...
+  type: text
+  source: from_cfe_payload
+  cfe_payload_location: result_summary.partner_disposable_income.total_outgoings_and_allowances
+  skip_unless: partner_income_relevant?
+
+# Total monthly disposable income / Client
+- name: undefined_73
+  type: text
+  source: from_cfe_payload
+  cfe_payload_location: result_summary.disposable_income.total_disposable_income
```
HettieS, line 480: just had a thought on this value based on some user research I was just in - it came up that minus values are not allowed on the CW forms (sounds obvious I know)

It's possible that these values can be a minus if the outgoings are somehow greater than the income. I wonder if we need to protect against that? Maybe we need a method here, so if the value is negative, return 0? What do you think?

PR #576, opened 2023-04-13
EL-828: Use correct values for pensioner disregard table when there is a partner

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-828)

## What changed and why

- We weren't including the pensioner disregard values applied to partner capital in the pensioner disregard table
- Also we were using `total_capital` rather than `total_capital_with_smod` in the first line, which was sometimes showing incorrect values (because SMOD has already been deducted by the time this table is displayed)

## Guidance to review

I have asked the CFE team if they can do the maths for us to save us doing `+`, and they will do this as a low-priority thing. In the meantime we will do the `+`.

## Checklist

### DIFF
--- a/app/models/calculation_result.rb
+++ b/app/models/calculation_result.rb
@@ -172,9 +172,10 @@ class CalculationResult
   end
 
   def pensioner_disregard_rows
-    total_capital = api_response.dig(:result_summary, :capital, :total_capital) +
-      api_response.dig(:result_summary, :partner_capital, :total_capital)
-    disregarded = api_response.dig(:result_summary, :capital, :pensioner_disregard_applied)
+    total_capital = api_response.dig(:result_summary, :capital, :total_capital_with_smod) +
+      api_response.dig(:result_summary, :partner_capital, :total_capital_with_smod)
+    disregarded = api_response.dig(:result_summary, :capital, :pensioner_disregard_applied) +
+      api_response.dig(:result_summary, :partner_capital, :pensioner_disregard_applied)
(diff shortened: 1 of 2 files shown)


HettieS (commented):
Looks good, although wondering if it might be nice to add some tests in `calculation_result_spec.rb` to check the rows appear as expected when there is/isn't partner pensioner disregard applied. Or add to the view spec!

patrick-laa:
@HettieS good point. I had to adjust the CalculationResult class to make it more forgiving of incomplete data so that I could write tests that only specified bits of the payload I cared about, because it turns out factories that produce deeply nested hashes are a bit of a pig to work with, but we now have more tests of the fixed functionality.

HettieS (approved):
Nice - tests look good. I've approved although I'm not too sure how I feel about how we've had to engineer the code to make the tests pass with the nil guarding. It feels like the code should represent the reality of the service/data and we should try and improve the test setup instead to best mimic that. I always feel like `&` also kind of obscures the intention of the code. But will leave with you to decide :)

patrick-laa:
@HettieS ugh when you put it like that I can't argue with it. I'll tighten up the api result factory to make overriding bits and pieces less painful so we don't need those guards...

HettieS:
> @HettieS ugh when you put it like that I can't argue with it. I'll tighten up the api result factory to make overriding bits and pieces less painful so we don't need those guards...

I'm really sorry 🥲 😂

PR #585, opened 2023-04-18
EL-833: Check your answers matter type text

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-833)

## What changed and why

If a check answers select field has an `alt_attribute` key, use that to construct the I18n key for the value text. That way we can use both the level of help and the proceeding type to determine what text to show for the 'What type of matter is this?' check answers field.

## Checklist


HettieS (approved):
👍

PR #595, opened 2023-04-19
EL-800 fill in CW1/2MH form

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-800)

## What changed and why

Added support for CW1_and_2MH form
Added new mapping file
Updated tests to check for sample of fields on new form

## Checklist


app/lib/controlled_work_mappings/cw1_and_2.yml
```diff
@@ -0,0 +1,361 @@
+# This file defines the mappings used to prepopulate a CW1/2MH PDF form with the answers given by a user to CCQ's questions
+# combined with the output from CFE.
+
+# The mappings conform to the format described in ControlledWorkDocumentValueMappingService
+
+
+# "The client has a partner whose means are to be aggregated:"
+- name: CheckBox17
```
HettieS, line 8: is the boolean implementation correct? Just checking as I had something similar on my form, where the yes and no checkboxes were two separate checkboxes rather than a boolean. I implemented them as two checkboxes rather than one boolean as I couldn't easily see how to combine them.

willc-work: Thanks Hettie, fixed now if you can take another look

HettieS: looks great! We've both created a `def no_partner` so can get rid of one depending on which form gets merged first 👍

MazOneTwoOne: Thanks both. I'll do something similar.

PR #597, opened 2023-04-20
[EL-798] Pre fill CW2 IMM form

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-798)

## What changed and why

We need to pre-fill CW forms to make things easier for case workers. This PR adds the CW2 IMM form and relevant mapping file.

## Guidance to review

Sense check the fields are correct, test the export. I didn't use the conditional `skip_if` SMOD boolean methods in this form as this form doesn't ask about SMOD assets - but there might be a case for having them in the cw2.yml doc. I can add them in if we feel they are needed. My thinking was the case worker should only select the CW2IMM form if there is no SMOD - so we don't need to worry about checking for it. But I could be wrong!

I'd like to write a few more tests as well so will continue to work on that. 

## Checklist


willc-work (approved):

app/lib/controlled_work_mappings/cw1.yml
```diff
@@ -159,14 +159,16 @@
 - name: Other property 1_2
   type: text
   source: from_attribute
-  attribute: additional_non_smod_properties_value
+  attribute: additional_properties_value
```
patrick-laa, line 162: This makes me uncomfortable. I don't want the value of additional properties here, I want the value of additional _non-smod_ properties here. Like how further up in the file I want the value of additional _smod_ properties.

app/lib/controlled_work_mappings/cw1.yml
```diff
@@ -159,14 +159,16 @@
...
+  form_type: cw1
```
patrick-laa, line 163: This feels redundant. We know the form type is cw1 _because this file is called cw1.yml_

app/models/controlled_work_document_content.rb
```diff
@@ -13,13 +13,16 @@
...
+
   def from_cfe_payload(path)
     path_parts = path.split(".").map { _1.to_i.to_s == _1 ? _1.to_i : _1 }
     value = session_data.dig("api_response", *path_parts)
     format(value)
   end
 
-  def from_attribute(attribute)
+  def from_attribute(attribute, form_type = nil)
+    @form_type = form_type
```
patrick-laa, line 25: Be wary of setting a permanent attribute on a model from a temporary method. In future, when new methods are added, we may not remember that depending on whether `from_attribute` had been called before or after it, `@form_type` may or may not be set and maybe this will affect how that new method operates.

app/models/controlled_work_document_content.rb
```diff
@@ -154,15 +170,23 @@ def combined_childcare_costs
...
   end
 
-  def additional_non_smod_properties
-    @additional_non_smod_properties ||= [
-      (session_data.dig("api_response", "assessment", "capital", "capital_items", "properties", "additional_properties") unless additional_property_in_dispute?),
-      session_data.dig("api_response", "assessment", "partner_capital", "capital_items", "properties", "additional_properties"),
-    ].flatten.compact
+  def additional_properties
+    case form_type
+    when "cw1"
```
patrick-laa, line 179: I really don't think this model, which is just a wrapper for some session data and until now knew nothing about CW forms, should have to know about the logic associated with specific forms. A method that returns different output depending on who is calling it is a bit iffy anyway, and this would mean committing to updating this method for every new CW form we add.

Hitherto the yaml file has been the only place that knows about the fields of any given CW form. I'd be strongly inclined to keep it that way.

HettieS:
@patrick-laa thanks for your feedback! Makes sense. I'll decouple from the forms and do the checks based on whether we want smod or not (or if we are smod agnostic) and then those specific methods can be used in the relevant form. I'll set the modifier in an initialiser and hopefully making those changes will address your comments. Hopefully getting my head round this finally.

app/models/controlled_work_document_content.rb
```diff
@@ -64,33 +66,33 @@ def main_home_percentage_owned
     end
   end
 
-  def additional_non_smod_properties_value
-    additional_non_smod_properties_sum "value"
+  def additional_properties_value(*modifier)
```
patrick-laa, line 69: I don't think you need the `*` here and below - these methods will only ever be called with 1 argument and possibly should explode if we ever try to call them with more.

app/models/controlled_work_document_content.rb
```diff
@@ -154,15 +164,24 @@ def combined_childcare_costs
...
 
-  def additional_non_smod_properties
-    @additional_non_smod_properties ||= [
-      (session_data.dig("api_response", "assessment", "capital", "capital_items", "properties", "additional_properties") unless additional_property_in_dispute?),
-      session_data.dig("api_response", "assessment", "partner_capital", "capital_items", "properties", "additional_properties"),
-    ].flatten.compact
+  def additional_properties(property_type = "all")
+    raise "Invalid modifier in YML" unless VALID_PROPERTY_TYPES.include?(property_type)
+
+    @additional_properties ||= if property_type.include?("non_smod")
```
patrick-laa, line 174: Given that we're now passing in a property_type, we shouldn't memoize (i.e. use `@additional_properties ||= `) any more. Because if first I called `additional_properties("all")` it would assign an array of all additional properties to `@additional_properties`, and if I then called `additional_properties("non_smod")`, all additional properties would still be returned to me.

patrick-laa: Also, arguably the check we want on the right hand side is `if property_type == "non_smod"` because if, as an absurd example, we added `smod_and_non_smod` to VALID_PROPERTY_TYPES, this "include" would break.

patrick-laa (commented):
Some tiny tweak suggestions, but this is looking really solid!

app/models/controlled_work_document_content.rb
HettieS: @patrick-laa I ended up adding the `*` as I haven't actually passed in a modifier in the `cw2.yml`. My thinking was, we set a default of `property_type = "all"` so it wasn't necessary to set that in the `cw2.yml` as well - but maybe we should. I didn't want to put `modifier: all` in a bunch of places, but maybe that's the better thing to do. What do you think?

app/models/controlled_work_document_content.rb
HettieS: will sort these two things out now :)

app/models/controlled_work_document_content.rb
patrick-laa: @HettieS ah, I see. I agree, we don't want to explicitly add "modifier" lines in the YAML. So if `additional_properties_value` is either going to be called with one or zero arguments, my instinct would be to set the method to `def additional_properties_value(property_type = "all")`, but TBH I do now see why you used `*` and if you wanted to stick with that I think it would be fine.

app/models/controlled_work_document_content.rb
```diff
@@ -154,15 +164,24 @@ def combined_childcare_costs
...
 
-  def additional_non_smod_properties
-    @additional_non_smod_properties ||= [
-      (session_data.dig("api_response", "assessment", "capital", "capital_items", "properties", "additional_properties") unless additional_property_in_dispute?),
-      session_data.dig("api_response", "assessment", "partner_capital", "capital_items", "properties", "additional_properties"),
-    ].flatten.compact
+  def additional_properties(property_type = "all")
+    raise "Invalid modifier in YML" unless VALID_PROPERTY_TYPES.include?(property_type)
+
+    @additional_properties = if property_type == "non_smod"
```
patrick-laa, line 174: You could delete `@additional_properties =` entirely - the variable doesn't get read anywhere, so there's nothing gained by setting it here.

patrick-laa (approved):
Looking good to me!

PR #598, opened 2023-04-20
EL-801 add support for populating civmeans7 form

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-801)

## What changed and why

Add mapping for civ means 7 form
add test for content
add pdf

## Checklist


app/lib/controlled_work_mappings/civ_means_7.yml
```diff
@@ -0,0 +1,487 @@
+# This file defines the mappings used to prepopulate a CIVMEANS7 PDF form with the answers given by a user to CCQ's questions
+# combined with the output from CFE.
+
+# The mappings conform to the format described in ControlledWorkDocumentValueMappingService
+
+# "The client has a partner whose means are to be aggregated:"
+- name: CheckBox3
```
HettieS, line 7: think this needs changing to a checkbox with just a `checked_value` attribute, and also another one probably needs adding for the 'no' option

app/lib/controlled_work_mappings/civ_means_7.yml
```diff
@@ -0,0 +1,487 @@
...
+  type: boolean_radio
+  source: from_attribute
+  attribute: partner
+  yes_value: "1"
+  no_value: "Off"
+
+# "The client's assets (held in sole name or jointly held) have been claimed by the opponent:"
+- name: CheckBox2
```
HettieS, line 15: same as above I think

HettieS (changes requested):
just noticed the same issue where we need handle fields as checkboxes and not boolean radios

PR #617, opened 2023-04-28
EL-820: Feature flags endpoint

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-820)

## What changed and why

New endpoint that allows us to see at a glance which feature flags are enabled.

## Checklist

### DIFF
--- a/app/views/feature_flags/index.html.slim
+++ b/app/views/feature_flags/index.html.slim
@@ -1,0 +1,32 @@
+.govuk-grid-column-two-thirds
+  h1.govuk-heading-xl = t(".feature_flags")
+  = govuk_table do |table|
+    ruby:
+      table.caption(size: "m", text: t(".time_based"))
+      table.body do |body|
+        body.row do |row|
+          row.cell(text: t(".flag"), header: true)
+          row.cell(text: t(".enabled"), header: true)
+        end
+        FeatureFlags.time_dependant.each do |flag|
+          body.row do |row|
+            row.cell(text: flag)
+            row.cell(text: FeatureFlags.enabled?(flag) ? t("generic.yes_choice") : t("generic.no_choice"))
+          end
+        end
+      end
+  = govuk_table do |table|
+    ruby:
+      table.caption(size: "m", text: t(".static"))
+      table.body do |body|
+        body.row do |row|
+          row.cell(text: t(".flag"), header: true)
+          row.cell(text: t(".enabled"), header: true)
+        end
+        FeatureFlags.static.each do |flag|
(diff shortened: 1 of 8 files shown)


spec/features/feature_flags_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe "Feature flags" do
+  before do
+    allow(FeatureFlags).to receive(:static).and_return(%i[static_flag])
+    allow(FeatureFlags).to receive(:time_dependant).and_return(%i[time_dependant_flag])
+    allow(FeatureFlags).to receive(:enabled?).with(:static_flag).and_return(true)
+    allow(FeatureFlags).to receive(:enabled?).with(:time_dependant_flag).and_return(false)
+  end
+
+  scenario "I accept cookies via the banner" do
```
HettieS, line 11: is this scenario text correct?

HettieS (approved):
Looks good, one small comment!

spec/features/feature_flags_spec.rb
patrick-laa: Not even remotely - that'll teach me to copy-paste without thinking.

app/views/feature_flags/index.html.slim
```diff
@@ -0,0 +1,32 @@
...
+          row.cell(text: t(".flag"), header: true)
+          row.cell(text: t(".enabled"), header: true)
+        end
+        FeatureFlags.static.each do |flag|
+          body.row do |row|
+            row.cell(text: flag)
+            row.cell(text: FeatureFlags.enabled?(flag) ? t("generic.yes_choice") : t("generic.no_choice"))
+          end
+        end
+      end
```
MazOneTwoOne, line 32: Not a blocker just a question - I know this isn't meant for the public, but am I right in saying this would be difficult for screen readers? We haven't specified table headings, body and row as per component guidelines.

patrick-laa: It's a good question. I wasn't sure so I added a spec to `spec/system/accessibility_spec.rb`, which does automated accessibility best-practice checking. It didn't complain about this page which makes me think that those additional fields, while good practice strictly speaking, aren't necessary to make the page accessible for screen readers. The automated check isn't perfect, so it could have missed something tho.

MazOneTwoOne: Fair enough and glad it's had some automated checks on it at least - here's the [suggested standards](https://www.w3.org/WAI/tutorials/tables/one-header/), I checked against.

PR #620, opened 2023-04-28
EL-862: Hide hint if feature flag enabled

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-862)

## What changed and why

Hide the inaccurate hint text if the flow is altered.

## Checklist

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -82,6 +82,12 @@ RSpec.configure do |config|
     Flipper.disable(:cw_forms)
   end
 
+  config.around(:each, :household_section_flag) do |example|
+    Flipper.enable(:household_section)
+    example.run
+    Flipper.disable(:household_section)
+  end
+
   config.before(:suite) do
     DatabaseCleaner.clean_with :truncation
   end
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -86,6 +86,12 @@
     Flipper.disable(:cw_forms)
   end
 
+  config.around(:each, :household_section_flag) do |example|
```
HettieS, line 89: a question on this - it looks like we turn the flag on for each test that is run, then back off. So I'm confused why `applicant_form_spec.rb:40` is passing 🤔

HettieS (approved):
Approved, with a question

spec/rails_helper.rb
patrick-laa: @HettieS we only turn the `household_section` feature flag on for tests that have the `household_section_flag` _test_ flag. `household_section_flag` is _not_ specified for `applicant_form_spec.rb:40` (either the "it" block itself, or a parent "context" or "describe" block). Whereas it _is_ set on applicant_form_spec.rb:44

PR #629, opened 2023-05-03
EL-883 CW form refactor

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-883)

## Checklist


patrick-laa (commented):
This is a huge improvement! A few comments on model methods, but overall this is putting the codebase into a much more manageable state.

spec/services/controlled_work_document_value_mapping_service_spec.rb
```diff
@@ -57,9 +71,159 @@
       mappings = YAML.load_file(Rails.root.join("app/lib/controlled_work_mappings/cw2.yml")).map(&:with_indifferent_access)
       result = described_class.call(session_data, mappings)
       representative_sample = {
+        "CheckBox13" => 1, # Has a partner
         "CheckBox69" => 1, # Not passporting
-        "CheckBox64" => 1, # Not asylum supported
+        "CheckBox64" => nil, # Asylum support not given
```
patrick-laa, line 76: Ooh, I was wrong! In this scenario I'm pretty sure we would want the "Asylum support?" question on the PDF to be filled in as "No". Now, technically we haven't _asked_ them the asylum support question, so `asylum_support` is nil, so `no_asylum_support?` is now returning nil. But I'm pretty sure for the purposes of the CW2 form we don't want to leave that question blank. So I think we should revert the change I asked for in `no_asylum_support?`

HettieS: Cool I'll put it back! Yeah, in some cases it feels like the expected behaviour is a bit up for grabs, so I wasn't sure which would be best in the end 🤷‍♀️ I'll revert to how it was as I think that makes sense to.

patrick-laa (approved):

PR #665, opened 2023-05-15
EL-861: Housing costs screen

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-861)

## What changed and why

- New page for both client and partner housing costs
- New page for Housing Costs in client only flow 
- Removing Housing Benefit question
- Removing ‘Housing Payments’ from both client outgoings and partner outgoings 

## Guidance to review

- Existing client and partner housing cost questions (housing payments and housing benefit) have been replaced with new screens for certificated and controlled flows.

## Checklist


app/forms/housing_costs_form.rb
```diff
@@ -0,0 +1,38 @@
+class HousingCostsForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include SessionPersistable
+  include NumberValidatable
+
+  ATTRIBUTES = %i[housing_payments housing_payments_frequency housing_benefit_value housing_benefit_frequency].freeze
+  HOUSING_PAYMENT_FREQUENCIES = %w[every_week every_two_weeks every_four_weeks monthly total].freeze
```
patrick-laa, line 8: Rather than repeat this constant here, it might be better to delete this and instead reference OutgoingsForm::VALID_FREQUENCIES. Likewise with MortgageOrLoanPaymentForm::HOUSING_PAYMENT_FREQUENCIES

app/forms/housing_costs_form.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  attribute :housing_payments, :gbp
+  validates :housing_payments, numericality: { greater_than_or_equal_to: 0, allow_nil: true }, presence: true
+
+  attribute :housing_payments_frequency, :string
+  validates :housing_payments_frequency,
+            presence: true,
+            inclusion: { in: HOUSING_PAYMENT_FREQUENCIES, allow_nil: false },
+            if: -> { send(:housing_payments_frequency).to_i.positive? }
```
patrick-laa, line 17: Here and in other places you don't need to do `send(:housing_payments_frequency)`, as you can simply do `housing_payments_frequency`

app/helpers/housing_costs_helper.rb
```diff
@@ -0,0 +1,8 @@
+module HousingCostsHelper
+  def outgoings_links(level_of_help)
```
patrick-laa: As mentioned in Slack, this name clashes with the method defined in OutgoingsHelper. Each method name in `helpers` needs to be unique.

app/helpers/mortgage_or_loan_payment_helper.rb
```diff
@@ -0,0 +1,8 @@
+module MortgageOrLoanPaymentHelper
+  def outgoings_links(level_of_help)
```
patrick-laa: Given that the text and link defined here are identical to the ones defined in the HousingCostsHelper method, you might as well delete this and use the other helper method in both screens.

app/lib/steps/property_section.rb
```diff
@@ -3,10 +3,11 @@ class PropertySection
     PROPERTY_STEPS = %i[property property_entry].freeze
     ADDITIONAL_PROPERTY_STEPS = %i[additional_property additional_property_details].freeze
     ADDITIONAL_PARTNER_PROPERTY_STEPS = %i[partner_additional_property partner_additional_property_details].freeze
+    HOUSING_COSTS = %i[housing_costs mortgage_or_loan_payment].freeze
```
patrick-laa, line 6: For total consistency this constant name should end `_STEPS`

app/views/estimate_flow/forms/_housing_costs.html.slim
```diff
@@ -0,0 +1,59 @@
...
+                                        @form.housing_payment_frequencies,
+                                        :first,
+                                        :last,
+                                        legend: -> { tag.p(t("estimate_flow.#{i18n_key}.frequency")) }
+  h2.govuk-heading-m class="govuk-!-margin-bottom-2"
+     = t("estimate_flow.#{i18n_key}.housing_benefits.label")
+  p.govuk-caption-m class="govuk-!-margin-bottom-4"
+    - t("estimate_flow.#{i18n_key}.housing_benefits.hint_paragraph").each
+      p.govuk-caption-m class="govuk-!-margin-bottom-2" = _1
+  ul.govuk-list.govuk-list--bullet class="govuk-hint govuk-!-margin-top-0 govuk-!-margin-bottom-0"
```
patrick-laa, line 35: Ok, this section needs tightening up. What you want is something like:

```
ul.govuk-list.govuk-list--bullet
  li = t("estimate_flow.#{i18n_key}.housing_benefits.hint_bullet_1")
  - if @check.partner
    li = t("estimate_flow.#{i18n_key}.housing_benefits.hint_bullet_2_partner")
  li = t("estimate_flow.#{i18n_key}.housing_benefits.hint_bullet_3")
```
To do this, you'll need to change the yaml for the bullets from:
```
hint_bullet_1:
  - Blah
 ```
to:
```
hint_bullet_1: Blah
 ```

Because each bullet is just one bullet, not an array of bullets.

app/views/estimate_flow/housing_costs.html.slim
```diff
@@ -0,0 +1,10 @@
+- content_for :page_title
+  = t(".heading")
+- content_for :back do
+  = back_link(:housing_costs, @check, @back_buttons_invoke_browser_back_behaviour)
+.govuk-grid-column-two-thirds
+  = render "estimate_flow/forms/housing_costs", i18n_key: "housing_costs"
```
patrick-laa, line 6: I don't think there's actually any benefit having the form in its own partial here. We introduced that pattern where 2 views needed to render the same form (i.e. client and partner views that basically show the same thing). In this case, I think it would make life simpler if you deleted this line, and instead copied and pasted the contents of `estimate_flow/forms/_housing_costs.html.slim` into this file, indenting appropriately. Then everything we need would be in one place.

config/locales/en.yml
```diff
@@ -807,6 +807,57 @@ en:
...
+    housing_costs:
+      heading: Housing costs
+      heading_hint: 
+        - "Enter the housing costs for the home the client lives in."
+      heading_hint_partner: 
+        - "Enter the housing costs for the home the client lives in."
+        - "If the client and partner live separately, do not include the partner’s housing costs."
+      housing_payments:
+        label: Housing payments
+        hint_paragraph:
```
patrick-laa, line 819: I'd rename this and similar ones below to `hint_paragraphs` with an `s`, to indicate that it's not a single string but rather an array of strings

spec/services/cfe/regular_transactions_payload_service_spec.rb
```diff
@@ -100,5 +100,19 @@
...
+
+    context "when the 'household flow' feature flag is active", :household_section_flag do
+      let(:session_data) do
+        {
+          "housing_payments_value" => nil,
+          "housing_payments_frequency" => nil,
+        }
+      end
+
+      it "does not send the housing payments deductions, as they are in another screen" do
```
patrick-laa, line 112: I think this test is redundant, and it's my fault for sending you banking up the wrong tree. The issue is that actually RegularTransactionsPayloadService doesn't _care_ which screen housing payments deductions were entered in, because the way you've written this code, they always get stored in the same place in the session. So the reason in this test that nothing is added to the payload is nothing to do with the feature flag, and everything to do with the face you've said that the session contains nil values.

Given that this class is unaffected by the new screens, this test is just a bit confusing and redundant. We can remove it.

spec/support/flow_helpers.rb
```diff
@@ -125,6 +125,22 @@ def fill_in_property_entry_screen(screen_name: :property_entry, form_name: :clie
   click_on "Save and continue"
 end
 
+def fill_in_housing_costs_screen(screen_name: :housing_costs, form_name: :housing_costs)
```
patrick-laa, line 128: The `screen_name` and `form_name` variables are redundant - again, it's a pattern that only gets used when we reuse the same method for both client and partner versions of a screen. So you can delete them both, and just hard-code `confirm_screen :housing_costs`, `fill_in "housing_costs_form[housing_payments]", with: "0"` etc, because there won't be a scenario where we want to use a _different_ screen name or form name here.

patrick-laa (commented):
Lots of little things, but overall this is looking really solid!

app/views/estimate_flow/housing_costs.html.slim
MazOneTwoOne: I guess we'd do the same for the `app/views/estimate_flow/_mortgage_or_loan_payment.html.slim` partial. I'll amalgamate this one too.

app/views/estimate_flow/housing_costs.html.slim
```diff
@@ -0,0 +1,65 @@
...
+                                          :last,
+                                          legend: -> { tag.p(t("estimate_flow.housing_costs.frequency")) }
+    h2.govuk-heading-m class="govuk-!-margin-bottom-2"
+      = t("estimate_flow.housing_costs.housing_benefits.label")
+    p.govuk-caption-m class="govuk-!-margin-bottom-4"
+      - t("estimate_flow.housing_costs.housing_benefits.hint_paragraphs").each
+        p.govuk-caption-m class="govuk-!-margin-bottom-2" = _1
+    ul.govuk-list.govuk-list--bullet class="govuk-hint govuk-!-margin-top-0 govuk-!-margin-bottom-0"
+      li = t("estimate_flow.housing_costs.housing_benefits.hint_bullet_1")
+    ul.govuk-list.govuk-list--bullet class="govuk-hint govuk-!-margin-top-1 govuk-!-margin-bottom-0"
```
patrick-laa, line 42: This still isn't quite right. We want one list with 3 items, not 3 lists each with 1 item.

I also think there's too much manual tidging of spacing with `govuk-!-margin` classes.  This is a standard page that should use standard typography in a standard way. If we're having to apply a bunch of manual spacing fixes that either means the design is mis-using the design standards or that we're mis-using the standard components. I think the fixes are as follows:

We shouldn't be using `p.govuk-caption-m` on this screen. We should be using `p.govuk-hint`. The only caption is in the header, which is dealt with in the `shared/heading` partial. If you swap those out, you should find you don't need to do manual spacing adjustment of the `p` elements.

Likewise the `li` elements should be `li.govuk-hint`s and the `ul` should _not_ have the govuk-hint class. That should fix the spacing issues of the bullets

My guess is that the `h2` elements don't need the margin-bottom adjustment classes at all. The designs look pretty much like they don't need it.

patrick-laa (commented):
Some tiny tweaks needed to one view, I think, but we're on the brink!

app/views/estimate_flow/housing_costs.html.slim
```diff
@@ -0,0 +1,70 @@
...
+                                          :first,
+                                          :last,
+                                          legend: -> { tag.p(t("estimate_flow.housing_costs.frequency")) }
+    h2.govuk-heading-m
+      = t("estimate_flow.housing_costs.housing_benefits.label")
+    p.govuk-hint
+      - t("estimate_flow.housing_costs.housing_benefits.hint_paragraphs").each
+        p.govuk-hint = _1
+
+    - if @check.partner
```
patrick-laa, line 41: Minor quibble: The `if` block here and the `unless` block have almost the same content. You could just do:

```slim
    ul.govuk-list.govuk-list--bullet
      li.govuk-hint = t("estimate_flow.housing_costs.housing_benefits.hint_bullet_1")
      - if @check.partner
        li.govuk-hint = t("estimate_flow.housing_costs.housing_benefits.hint_bullet_2_partner")
      li.govuk-hint = t("estimate_flow.housing_costs.housing_benefits.hint_bullet_3")

patrick-laa (approved):
One small suggestion, but I think we're ready for UAT!

PR #676, opened 2023-05-18
Utility to detect stale i18n keys

patrick-laa:
Having poked around at our `en.yml` file for #675 I noticed that there are a few obsolete keys in there. ~I'll raise a ticket to clear them out, but in the meantime~ I've built a little utility to help us detect obsolete keys to make that task easier. Here's a sample of its output:

Screenshot 2023-05-18 at 10 48 11

Update: I have also now cleared out the obsolete keys, and fixed a couple of small bugs where we were wrongly _not_ using a key that we should have been.

### DIFF
--- a/spec/unused_i18n_key_checker.rb
+++ b/spec/unused_i18n_key_checker.rb
@@ -6,18 +6,18 @@ class UnusedI18nKeyChecker
     @used_keys << key.to_s unless @used_keys.include?(key.to_s)
   end
 
-  def self.check_unused_keys(ignore_stems: [])
+  def self.check_unused_keys(ignore: [])
     return unless ENV["CHECK_UNUSED_KEYS"]
 
     mappings = YAML.load_file(Rails.root.join("config/locales/en.yml"))
     defined_keys = define_keys(mappings["en"])
     not_ignored_defined_keys = defined_keys.reject do |defined_key|
-      ignore_stems.any? { defined_key.starts_with?(_1) }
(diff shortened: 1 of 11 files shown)


spec/unused_i18n_key_checker.rb
```diff
@@ -0,0 +1,35 @@
...
+    @used_keys << key.to_s unless @used_keys.include?(key.to_s)
+  end
+
+  def self.check_unused_keys(ignore_stems: [])
+    return unless ENV["CHECK_UNUSED_KEYS"]
+
+    mappings = YAML.load_file(Rails.root.join("config/locales/en.yml"))
+    defined_keys = define_keys(mappings["en"])
+    not_ignored_defined_keys = defined_keys.reject do |defined_key|
+      ignore_stems.any? { defined_key.starts_with?(_1) }
```
davidread, line 13: A typical i18n stemmer like Snowball will convert "liking" to the stem "like". So `starts_with` wouldn't include it. That may be helpful or maybe I've simply misunderstood what's going on here....!

patrick-laa: Ah, different sort of "stem" here. This isn't to do with the text content but rather the key. E.g. a typical key might be "activemodel.errors.models.level_of_help_form.attributes.level_of_help.blank", and this line means that if we've passed in `ignore_stems: ["activemodel.error"]` as a parameter, that key would be ignored because it starts with an ignorable stem.

willc-work (approved):
Looks great.

PR #691, opened 2023-05-24
EL-870 Finalise Household flow

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-870)

Updated the flow to match the miro board
https://miro.com/app/board/uXjVPBBAi10=/?moveToWidget=3458764552902974344&cot=14

## Guidance to review
The steps/sections that we use could do with a refactor as the new household flow groups things in a different way. I propose to create a ticket for refactoring the sections when we remove the household_flow feature flag.

## Checklist

### DIFF
--- a/app/lib/steps/applicant_details_section.rb
+++ b/app/lib/steps/applicant_details_section.rb
@@ -2,7 +2,7 @@ module Steps
   class ApplicantDetailsSection
     class << self
       def all_steps
-        %i[applicant dependant_details]
+        %i[applicant dependant_details partner_details]
       end
 
       def all_steps_for_current_feature_flags
@@ -18,11 +18,15 @@ module Steps
       end
 
       def groups(session_data)
-        [:applicant, dependant_details(session_data)].compact
+        [:applicant, dependant_details(session_data), partner_details(session_data)].compact
       end
 
       def dependant_details(session_data)
         :dependant_details unless Steps::Logic.passported?(session_data)
+      end
+
+      def partner_details(session_data)
+        :partner_details if Steps::Logic.passported?(session_data) && FeatureFlags.enabled?(:household_section)
       end
     end
   end
--- a/app/lib/steps/applicant_details_section.rb
+++ b/app/lib/steps/applicant_details_section.rb
@@ -26,6 +26,8 @@ module Steps
       end
 
       def partner_details(session_data)
+        return unless Steps::Logic.partner?(session_data)
+
         :partner_details if Steps::Logic.passported?(session_data) && FeatureFlags.enabled?(:household_section)
       end
     end
--- a/app/lib/steps/capital_section.rb
+++ b/app/lib/steps/capital_section.rb
@@ -11,9 +11,9 @@ module Steps
 
       def all_steps_for_current_feature_flags
         if FeatureFlags.enabled?(:household_section)
-          %i[assets partner_assets vehicle vehicles_details assets].freeze
+          %i[assets partner_assets vehicle vehicles_details].freeze
         else
-          (PROPERTY_STEPS + %i[vehicle vehicle_details] + TAIL_STEPS).freeze
+          (PROPERTY_STEPS + %i[vehicle vehicle_details assets]).freeze
         end
       end
 
@@ -26,7 +26,7 @@ module Steps
         else
           [property_steps(session_data),
            vehicle_steps(session_data),
-           Steps::Group.new(*TAIL_STEPS)].compact
+           Steps::Group.new(:assets)].compact
         end
       end
 
--- a/app/lib/steps/capital_section.rb
+++ b/app/lib/steps/capital_section.rb
@@ -31,8 +31,6 @@ module Steps
       end
 
       def property_steps(session_data)
-        return if FeatureFlags.enabled?(:household_section)
-
         Steps::Group.new(*(Steps::Logic.owns_property?(session_data) ? PROPERTY_STEPS : %i[property]))
       end
 
--- a/app/lib/steps/capital_section.rb
+++ b/app/lib/steps/capital_section.rb
@@ -2,11 +2,11 @@ module Steps
   class CapitalSection
     PROPERTY_STEPS = %i[property property_entry].freeze
     VEHICLE_STEPS = %i[vehicle vehicle_details vehicles_details].freeze
-    TAIL_STEPS = %i[assets partner_assets].freeze
+    ASSET_STEPS = %i[assets partner_assets].freeze
 
     class << self
       def all_steps
-        (PROPERTY_STEPS + VEHICLE_STEPS + TAIL_STEPS).freeze
+        (PROPERTY_STEPS + VEHICLE_STEPS + ASSET_STEPS).freeze
       end
 
       def all_steps_for_current_feature_flags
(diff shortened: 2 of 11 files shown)


patrick-laa:
I'm going to pair-review this with @MazOneTwoOne tomorrow.

app/lib/steps/helper.rb
```diff
@@ -49,14 +49,24 @@ def step_groups_for(session_data)
...
+           PropertySection]
+        else
+          [CaseDetailsSection,
+           ApplicantDetailsSection,
+           IncomeSection,
+           CapitalSection,
+           PartnerSection,
+           PartnerIncomeSection,
+           PartnerCapitalSection,
+           PropertySection]
```
patrick-laa, line 68: PropertySection is always empty when the feature flag is disabled, so this can be removed.

app/lib/steps/partner_capital_section.rb
```diff
@@ -2,19 +2,15 @@ module Steps
   class PartnerCapitalSection
     PROPERTY_STEPS = %i[partner_property partner_property_entry].freeze
     VEHICLE_STEPS = %i[partner_vehicle partner_vehicle_details].freeze
-    TAIL_STEPS = %i[partner_assets].freeze
+    TAIL_STEPS = %i[partner_details partner_assets].freeze
```
patrick-laa, line 5: I think this is a typo - when the flag is enabled, PartnerCapitalSection isn't used, when it's not enabled, partner_details are provided by the PartnerSection.

app/lib/steps/capital_section.rb
```diff
@@ -2,7 +2,7 @@ module Steps
   class CapitalSection
     PROPERTY_STEPS = %i[property property_entry].freeze
     VEHICLE_STEPS = %i[vehicle vehicle_details vehicles_details].freeze
-    TAIL_STEPS = %i[assets].freeze
+    TAIL_STEPS = %i[assets partner_assets].freeze
```
patrick-laa, line 5: Would it make sense to rename this ASSET_STEPS, given that in the new flow it doesn't appear on the tail end, it's at the start, and we don't use this constant in the old flow at all?

app/lib/steps/capital_section.rb
```diff
@@ -11,23 +11,26 @@ def all_steps
...
       end
 
       def grouped_steps_for(session_data)
         return [] if Steps::Logic.asylum_supported?(session_data)
 
-        [property_steps(session_data),
-         vehicle_steps(session_data),
-         Steps::Group.new(*TAIL_STEPS)].compact
+        if FeatureFlags.enabled?(:household_section)
+          [Steps::Group.new(*TAIL_STEPS),
```
patrick-laa, line 24: I don't think we want these screens in the same Group - at the moment, if you change client assets from the check answers screen, you're taken to the partner assets screen next before you get back to the check answers screen, which seems weird to me. I think they each belong in their own group.

app/lib/steps/applicant_details_section.rb
```diff
@@ -18,12 +18,16 @@ def grouped_steps_for(session_data)
...
-        [:applicant, dependant_details(session_data)].compact
+        [:applicant, dependant_details(session_data), partner_details(session_data)].compact
       end
 
       def dependant_details(session_data)
         :dependant_details unless Steps::Logic.passported?(session_data)
       end
+
+      def partner_details(session_data)
+        :partner_details if Steps::Logic.passported?(session_data) && FeatureFlags.enabled?(:household_section)
```
patrick-laa, line 29: If the client doesn't have a partner this method should return nil - otherwise we see the partner details screen even when there is no partner.

app/lib/steps/capital_section.rb
patrick-laa: Also, if the client has no partner, we don't want to include partner_assets at all.

app/lib/steps/outgoings_section.rb
```diff
@@ -0,0 +1,22 @@
...
+      def all_steps_for_current_feature_flags
+        all_steps
+      end
+
+      def grouped_steps_for(session_data)
+        if Steps::Logic.passported?(session_data) || Steps::Logic.asylum_supported?(session_data)
+          []
+        else
+          [Steps::Group.new(:outgoings),
+           Steps::Group.new(:partner_outgoings)].compact
```
patrick-laa, line 17: We don't want this screen if the client has no partner

spec/support/flow_helpers.rb
```diff
@@ -157,6 +157,16 @@ def fill_in_vehicle_details_screen(screen_name: :vehicle_details, form_name: :cl
   click_on "Save and continue"
 end
 
+def fill_in_vehicles_details_screen(screen_name: :vehicles_details, vehicle_finance: "0")
+  confirm_screen screen_name
```
patrick-laa, line 161: We don't need `screen_name` to be passed in as an argument - it's always going to be "vehicles_details". The only reason it's an argument in some methods in this file is when we use the same method to fill out both the client and partner versions of a screen, which have different screen names.

spec/flows/household_flow_spec.rb
```diff
@@ -7,24 +7,36 @@
     fill_in_applicant_screen(partner: "Yes")
     fill_in_forms_until(:outgoings)
```
patrick-laa, line 8: Given that this spec is now the official definition of the expected household flow, I think there's a case for replacing `fill_in_forms_until` with an explicit list of `fill_in_x_screen`, `fill_in_y_screen`. Quite a lot has changed between the applicant screen and the outgoings screen and it would be good to list out what that new flow looks like.

spec/flows/household_flow_spec.rb
```diff
@@ -7,24 +7,36 @@
...
+    fill_in_property_screen(choice: "Yes, owned outright")
+    fill_in_property_entry_screen
+    fill_in_additional_property_screen(choice: "Yes, owned outright")
+    fill_in_additional_property_details_screen
+    fill_in_partner_additional_property_screen(choice: "Yes, owned outright")
+    fill_in_partner_additional_property_details_screen
+    confirm_screen("check_answers")
+  end
+
+  it "includes partner_details in the flow when applicant is passported" do
```
patrick-laa, line 27: Given that EL-873 is a 0-pointer, I think we need to verify as part of this ticket that the household flow works when there is no partner. It would be great to have tests here that runs through the screens that appear in non-partner scenarios. (Maybe both passported and non-passported?)

spec/flows/household_flow_spec.rb
```diff
@@ -7,24 +7,36 @@
...
-    confirm_screen("vehicle")
-    fill_in_forms_until(:partner_details)
-    fill_in_partner_details_screen
-    # partner dependants and housing benefits screens are skipped
-    confirm_screen("partner_benefits")
     fill_in_forms_until(:partner_outgoings)
     fill_in_partner_outgoings_screen
-    # partner property screen is skipped
-    # partner vehicle screen is skipped
+    confirm_screen("assets")
```
patrick-laa, line 12: We don't technically need a `confirm_screen(X)` if immediately afterwards we have a `fill_in_X_screen` call, because the latter always includes a call to the former. (I know we were already doing it in a couple of places, but I just spotted that it's a bit redundant)

patrick-laa (commented):
Overall shape of it looks good - just a few quibbles, mostly with the client-only flow.

app/lib/steps/partner_capital_section.rb
willc-work: yes sorry hangover from when I thought details should be bewteen assets pages

patrick-laa (approved):

PR #692, opened 2023-05-24
EL-903 add details component to property page

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-903)

add new details component to the household property page when spag flag is enabled

## Checklist

### DIFF
--- a/app/views/estimate_flow/household_flow/_property.html.slim
+++ b/app/views/estimate_flow/household_flow/_property.html.slim
@@ -9,7 +9,7 @@
     = form.govuk_error_summary t("generic.error_summary_title")
 
     = render "shared/household_tag"
-    - if FeatureFlags.enabled?(:special_applicant_groups)
+    - if FeatureFlags.enabled?(:special_applicant_groups) && @check.partner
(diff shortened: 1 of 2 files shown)


app/views/estimate_flow/household_flow/_property.html.slim
```diff
@@ -7,9 +7,18 @@
 .govuk-grid-column-two-thirds
   = form_for(@form, url: wizard_path, method: :put) do |form|
     = form.govuk_error_summary t("generic.error_summary_title")
-    = form.govuk_collection_radio_buttons :property_owned, property_options, :first, :last,
-                                          legend: { text: t(legend_key), size: "xl", tag: "h1" },
-                                          hint: { text: t("estimate_flow.property.hint") }
+
+    - if FeatureFlags.enabled?(:special_applicant_groups)
```
patrick-laa, line 11: I think this only applies if the flag is enabled _and_ the client has a partner. The "legacy hint" will continue to be the thing to show if we render this screen and the client has no partner.

PR #693, opened 2023-05-24
EL-901 & EL-902: Special applicants assets

patrick-laa:
[Jira ticket 1](https://dsdmoj.atlassian.net/browse/EL-901)
[Jira ticket 2](https://dsdmoj.atlassian.net/browse/EL-902)

## What changed and why

New help text for special applicant groups on household flow assets pages

## Checklist

### DIFF
--- a/app/helpers/assets_helper.rb
+++ b/app/helpers/assets_helper.rb
@@ -1,19 +1,36 @@
 module AssetsHelper
   def assets_links(level_of_help, smod_applicable)
-    first_links = {
+    links = {
       t("estimate_flow.assets.guidance.#{level_of_help}.text") => t("estimate_flow.assets.guidance.#{level_of_help}.link"),
       t("estimate_flow.assets.other_property_guidance.text") => t("estimate_flow.assets.other_property_guidance.#{level_of_help}.link"),
     }
-    links = if level_of_help == "controlled"
-              first_links
-            else
-              first_links.merge({
-                t("estimate_flow.assets.disregarded_capital_guidance.text") => t("estimate_flow.assets.disregarded_capital_guidance.link"),
-              })
-            end
+
+    if level_of_help == "certificated"
+      links[t("estimate_flow.assets.disregarded_capital_guidance.text")] = t("estimate_flow.assets.disregarded_capital_guidance.link")
+
+      if FeatureFlags.enabled?(:special_applicant_groups)
+        links[t("estimate_flow.assets.prisoner_guidance.text")] = t("estimate_flow.assets.prisoner_guidance.link")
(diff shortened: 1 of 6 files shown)


app/views/estimate_flow/household_flow/_partner_assets.html.slim
```diff
@@ -8,9 +8,6 @@
            show_smod_content: false,
            hints: t("estimate_flow.partner_assets.hints")
 
-  - translation_text = @check.controlled? ? "controlled" : "certificated"
 = render "shared/question_sidebar", \
```
MazOneTwoOne, line 11: Just a question, not a blocker. Why would it not be like this following the pattern in similar partials? 

```
= render "shared/question_sidebar",
    level_of_help: @check.level_of_help,
    links: partner_assets_links(@check)

```

app/helpers/assets_helper.rb
```diff
@@ -1,19 +1,36 @@
...
+        links[t("estimate_flow.assets.bankrupt_guidance.text")] = t("estimate_flow.assets.bankrupt_guidance.link")
+      end
+    end
 
     return links unless smod_applicable
 
     links.merge({ t("generic.smod.guidance.text") => t("generic.smod.guidance.#{level_of_help}.link") })
   end
+
+  def partner_assets_links(check)
```
MazOneTwoOne, line 22: Just a question, not a blocker.

Why would we not set `@check.level_of_help` in the '_partner_assets.html.slim' partial? That way we could get rid this line:

`translation_text = check.controlled? ? "controlled" : "certificated"`

app/views/estimate_flow/household_flow/_partner_assets.html.slim
patrick-laa: I'm not sure I understand the question - what we're doing _is_ that, isn't it?

app/helpers/assets_helper.rb
patrick-laa: Again, not sure I follow. Are you suggesting that we pass in the level of help as an argument here instead of the whole check? Either way, we need _some_ way of knowing which translation text to use. Or are you suggesting we use `@check.level_of_help` instead of `translation_text`, since they end up as the same thing?

For the latter, it's a hangover from back when `level_of_help` could be nil (specifically when the level of help _page_ was feature flagged). So we could drop that now.

app/views/estimate_flow/household_flow/_partner_assets.html.slim
MazOneTwoOne: The is an errand '\' in there.

app/helpers/assets_helper.rb
MazOneTwoOne: That we use  `@check.level_of_help` instead of  `translation_text` - thanks for explaining why we're not following that now

PR #697, opened 2023-05-26
EL-930: Add a user agent string

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-930)

## What changed and why

Send a user agent header to CFE on every request

## Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,7 +85,8 @@ RUN apk add --update --no-cache tzdata && \
     echo "Europe/London" > /etc/timezone
 
 # libpq: required to run postgres
-RUN apk add --no-cache libpq postgresql-client
+# git: used by the app to report its version to CFE
+RUN apk add --no-cache libpq postgresql-client git
 
 # Install Chromium and Puppeteer for PDF generation
 # Installs latest Chromium package available on Alpine (Chromium 108)
--- a/helm_deploy/laa-estimate-eligibility/values/production.yaml
+++ b/helm_deploy/laa-estimate-eligibility/values/production.yaml
@@ -51,7 +51,7 @@ googleTagManager:
 
 cfe:
   host: http://cfe-civil.cfe-civil-production.svc.cluster.local
-  environment_name: uat
+  environment_name: production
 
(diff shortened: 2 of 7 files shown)


helm_deploy/laa-estimate-eligibility/values/production.yaml
```diff
@@ -51,6 +51,7 @@ googleTagManager:
 
 cfe:
   host: http://cfe-civil.cfe-civil-production.svc.cluster.local
+  environment_name: uat
```
willc-work, line 54: should this be production?

patrick-laa: Whoops. Yup.

willc-work (approved):

Dockerfile
```diff
@@ -85,7 +85,8 @@ RUN apk add --update --no-cache tzdata && \
     echo "Europe/London" > /etc/timezone
 
 # libpq: required to run postgres
-RUN apk add --no-cache libpq postgresql-client
+# git: used by the app to report its version to CFE
+RUN apk add --no-cache libpq postgresql-client git
```
davidread, line 89: 1. Have you tested this works in a Docker-based environment? I'm surprised git works inside the Docker image, because `.git` is in the .dockerignore, so that dir shouldn't get copied to the Docker image, so git shouldn't work at all.
2. It's not ideal for us to add git to CCQ's attack surface (if someone got inside. Although you've already got chromium 🤯)
3. It's not ideal to run git every time the app starts up

It would be better to do that task at build time i.e. in the Dockerfile. You could add a line just before `43 COPY . .` something like:
```
git rev-parse --short HEAD > ./VERSION
```
and in your app read the file.

patrick-laa: 🤦 I think I got a bit demob happy on my lat day before a break and rushed this. I'll fix it up when I'm back, or if you want to repurpose EL-933 as a 'fix the implementation' ticket and stick it in the backlog, someone else might be able to get to it next week.

PR #722, opened 2023-06-07
EL-906 Add aria labels

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-906)

Add aria-labels to some input fields to allow for proper screen reader announcements for what should happen with the input field

## Checklist

### DIFF
--- a/app/views/shared/_money_input.html.slim
+++ b/app/views/shared/_money_input.html.slim
@@ -1,7 +1,12 @@
-- size ||= nil
+ruby:
+  size ||= nil
+  set_aria ||= false
+  level ||= nil
+  checktype = level == "controlled" ? "controlled_" : nil
 = form.govuk_text_field field,
                         width:,
                         label: { text: label_text, size: },
                         prefix_text: BuildEstimatesHelper::POUND,
                         value: decimal_as_money_string(form, field),
-                        inputmode: "decimal"
+                        inputmode: "decimal",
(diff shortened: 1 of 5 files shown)


app/views/shared/_money_input.html.slim
```diff
@@ -1,7 +1,12 @@
-- size ||= nil
+ruby:
+  size ||= nil
+  set_aria ||= nil
```
patrick-laa, line 3: Super minor pedantry, but we treat `set_aria` as a boolean, so maybe it should default to `false` rather than `nil`. Makes _no_ practical difference, so feel free to ignore!

config/locales/en.yml
```diff
@@ -565,6 +565,20 @@ en:
...
+      pension_value: Enter amount of pension
+      controlled_student_finance_value: Enter amount of student finance
+      student_finance_value: Enter amount of student finance
+      other_value: Enter amount of income received from other sources in the last 3 months
+      controlled_other_value: Enter amount of income received from other sources in the last month
+      childcare_payments_value: Enter amount of childcare payments
+      maintenance_payments_value: Enter amount of maintenance payments from a former partner
+      legal_aid_payments_value: Enter amount paid towards legal aid in a criminal case
+      housing_payments_value: Enter amount of housing payments
+      housing_benefit_value: Enter the amount of housing benefit
```
patrick-laa, line 581: I think there's a missing change on the `housing_costs` screen, because at the moment the 
 `housing_payments` and `housing_benefit_value` fields on those screens aren't having these labels applied.

PR #728, opened 2023-06-08
EL-942 Client employment status screen

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-942)

## What changed and why

We are adding new employment screens to enable a better journey for users with self employed clients.

This PR:
* Hides the employment question on the Applicant/client details section behind the self employed feature flag
* Adds the new employment status screen which renders content conditionally according to level of help
* Updates the check answers page accordingly

This screen will be followed by amendments to the current employment screen in EL-943.

This PR does not make any changes to the payload we send to CFE. I believe that is being done in EL-943 but could be wrong!

## Guidance to review

* Check my logic around the feature flag makes sense for the different combinations of household flag, self employed flag, controlled and certificated journeys
* This PR might need a rejig in places after we remove the household flow flag

## Checklist


app/lib/check_answers_fields.yml
```diff
@@ -56,6 +56,7 @@ sections:
           type: boolean
         - attribute: employment_status
           type: select
+          skip_unless: self_employed_flag_disabled?
```
patrick-laa, line 59: There's sort of a double negative here (`unless` + `disabled`), which makes this a bit hard to read. You _could_ change this to `skip_if: self_employed_flag_enabled?`, if you were prepared to update `CheckAnswers::SectionListerService#build_field` with a line like:

 ```
return if field_data[:skip_if].present? && @model.send(field_data[:skip_if])
```

Very possibly overkill though!

File: app/lib/steps/income_section.rb [hunk 2]
```diff
@@ -35,6 +40,12 @@ def employment_steps(session_data)
         Steps::Group.new(:employment) if Steps::Logic.employed?(session_data)
       end
 
+      def employment_status_step
```
patrick-laa, line 43: I think we want `employment_status` in the same `Steps::Group` as `employment` (the Group system determines whether you go on a multi-screen loop when you click a Change link on the check answers page. But equally, that might be something to sort out in EL-943 when we wire up the new employment screen, so maybe we can ignore that for now.

app/lib/steps/logic.rb
```diff
@@ -43,7 +43,9 @@ def owns_vehicle?(session_data)
       def employed?(session_data)
         return false if asylum_supported?(session_data) || passported?(session_data)
 
-        ApplicantForm::EMPLOYED_STATUSES.map(&:to_s).include? session_data["employment_status"]
+        employment_status_form = FeatureFlags.enabled?(:self_employed) ? EmploymentStatusForm::EMPLOYED_STATUSES : ApplicantForm::EMPLOYED_STATUSES
```
patrick-laa, line 46: This variable should perhaps be called `employment_statuses` as it's an array of strings rather than a form.

spec/support/flow_helpers.rb
```diff
@@ -37,6 +37,15 @@ def fill_in_applicant_screen(choices = {})
   click_on "Save and continue"
 end
 
+def fill_in_applicant_screen_no_employment_question(choices = {})
```
patrick-laa, line 40: What we've done elsewhere in this file is to check if a field exists before trying to fill it in. I.e. instead of adding this new method, update fill_in_applicant_screen to say

```
hoose choices.fetch(:employed, "Unemployed"), name: "applicant_form[employment_status]" if page.text.include?("Whatever the employment question is")
```

patrick-laa (approved):
This looks really solid to me. I've got some very minor suggestions, but nothing that would require more than trivial changes. So to keep things moving in your absence I'm going to move this on to UAT.

app/lib/steps/income_section.rb
patrick-laa: Turns out I'm wrong about this anyway - Kate has confirmed that she doesn't want the two screens to be grouped in this way (and we may eventually remove this whole grouping system on all sections)

PR #759, opened 2023-06-23
EL-960: Add a content security policy

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-960)

## What changed and why

- Added a content security policy using nonces
- Every script loaded now uses a nonce
- Configured reporting to Sentry

## Checklist

### DIFF
--- a/helm_deploy/laa-estimate-eligibility/values/production.yaml
+++ b/helm_deploy/laa-estimate-eligibility/values/production.yaml
@@ -67,3 +67,6 @@ geckoboard:
   validationsDataset: "ccq_validations_production"
   journeysDataset: "ccq_journeys_production"
   enabled: enabled
+
+sentry:
+  cspReportEndpoint: https://o345774.ingest.sentry.io/api/4504177640275968/security/?sentry_key=217c7d14dd2e4cfa8dc62924a1bbd237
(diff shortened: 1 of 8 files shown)


helm_deploy/laa-estimate-eligibility/values/production.yaml
```diff
@@ -66,3 +66,6 @@ geckoboard:
   validationsDataset: "ccq_validations_production"
   journeysDataset: "ccq_journeys_production"
   enabled: enabled
+
+sentry:
+  cspReportEndpoint: https://o345774.ingest.sentry.io/api/4504177640275968/security/?sentry_key=217c7d14dd2e4cfa8dc62924a1bbd237
```
davidread, line 71: it looks fishy to have this production key in public, rather than in a secret. if this is ok to be public, please add a comment in the file (and same for staging.yaml)

patrick-laa: Good point! I had to double check to reassure myself that these aren't secret (because they'll appear in the response header of every request to our service), so if we can save someone else that same stomach lurch with a comment then that's a good thing. Will do.

willc-work (approved):

PR #768, opened 2023-06-27
EL-965: short lived credentials for ECR

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-965)

## What changed and why

Change how we tag images so we can differentiate between branch images and production images in our new lifecycle policy
Stop using a cron job to clear up old ECR images (we use a lifecycle policy instead)
Switch to using short-lived ECR credentials when we build and push

## Guidance to review

The lifecycle policy is being set up in the cloud-platform-environments repo (PR to follow)

## Checklist


.circleci/config.yml
```diff
@@ -199,21 +200,23 @@ jobs:
...
-          template: basic_fail_1
-
+      - aws-cli/setup:
+          role_arn: $ECR_ROLE_TO_ASSUME
+          region: $ECR_REGION
+      - run: |
+          aws ecr get-login-password --region $ECR_REGION | docker login --username AWS --password-stdin ${AWS_ECR_REGISTRY_ID}.dkr.ecr.${ECR_REGION}.amazonaws.com
+      - aws-ecr/build-image:
+          push-image: true
+          tag: << parameters.image_type >>-${CIRCLE_SHA1}
```
exonian, line 217: What could the `aws-ecr/build-and-push-image` job not handle, that we need to go back to rolling our own?
Can this concatenation not be done in the workflow definition, or is it something else?

bin/deploy
```diff
@@ -14,26 +14,26 @@ deploy_dynamic() {
...
                           --values ./helm_deploy/laa-estimate-eligibility/values/$NAMESPACE.yaml \
                           --set image.repository="$ECR_ENDPOINT" \
-                          --set image.tag="${CIRCLE_SHA1}"
+                          --set image.tag="production-$CIRCLE_SHA1"
 }
 
 
-if [[ "$NAMESPACE" == "uat" ]]; then
-  deploy_dynamic
+if [[ "$NAMESPACE" == "branch" ]]; then
```
exonian, line 35: I think "namespace" becomes a confusing variable name if we change its usage to be directly the deployment type and not [analogous to] the k8s namespace

.circleci/config.yml
```diff
@@ -260,32 +263,52 @@ workflows:
...
+          image_type: "production"
           requires:
             - install_dependencies
           filters:
             branches:
               only:
                 - main
       - deploy:
-          name: Deploy UAT
+          name: Deploy branch UAT
```
exonian, line 285: This repetition is distressing but I can't think of a way around it :(

bin/deploy
```diff
@@ -14,26 +14,26 @@ deploy_dynamic() {
...
 
-deploy_static() {
+deploy_standard() {
   helm upgrade check-client-qualifies ./helm_deploy/laa-estimate-eligibility/. \
                           --install --wait \
                           --namespace=${K8S_NAMESPACE} \
                           --values ./helm_deploy/laa-estimate-eligibility/values/$NAMESPACE.yaml \
                           --set image.repository="$ECR_ENDPOINT" \
-                          --set image.tag="${CIRCLE_SHA1}"
+                          --set image.tag="production-$CIRCLE_SHA1"
```
exonian, line 31: Oh so "production" in these docker images is not the same as "production" the namespace, right? It's used by prod, staging and main-uat?

That is as valid as it is confusing 😅 
Perhaps we might consider naming them for their longevity/stability to avoid confusion of the two senses of "production"?

bin/deploy
patrick-laa: You're right, I'm overloading this variable. Will fix.

.circleci/config.yml
patrick-laa: Ooh, I wonder if I can use the `$CIRCLE_BRANCH` variable instead of two branch filters...

.circleci/config.yml
patrick-laa: The issue, as I understand it, is that aws-ecr doesn't know how to use OIDC short-lived credentials, so we need to define a job that uses aws-cli to handle authentication for us. At least, that's my read of the [guidance](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/deploying-an-app/container-repositories/deprecating-long-lived-credentials.html#migrating-to-short-lived-credentials-for-circleci)

exonian: Oh, you're absolutely right! How disappointing! Right you are, then :)

PR #770, opened 2023-06-27
EL-961 Derive feature flags from session data, then env vars

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-961)

## What changed and why

For some types of feature flags, we will now derive the value of these from the session_data on the check object. This is so that flags can be backwards compatible, and eliminate errors associated with a flag being switched on when a user check is still underway. 

Relevant flag values i.e. those changing content/flow that are not policy specific, are stored on the session_data at the beginning of a check. These are then used to determine the flag value throughout the check.

I believe time-based flags are policy specific so these are not stored on the session_data. And not all static flags would affect the user flow (cw form flag for example). So I think we need to think about whether the flag is a global flag or a session flag on a case by case basis.

There are now three arguments that can be passed into `def enabled?` which are the flag name, `session_data` (if applicable) and a keyword argument of `without_session_data`. Although it isn't technically required, always pass in `session_data` when you can. In places where you cannot pass it in, pass in `without_session_data: true`. Usage is:

`FeatureFlags.enabled?(:flag, session_data)`
`FeatureFlags.enabled?(:flag, without_session_data: true)`

## Guidance to review

Check the places where I have passed in `session_data` to `FeatureFlags.enabled?` and make sure it makes sense
Make sure the README makes sense!

I'm not sure if `without_session_data` is the best naming so open to any other suggestions/implementation!

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,11 @@ For "static" feature flags we set the flag values in env vars.
 To add a new feature flag, set a `"#{flag_name.upcase}_FEATURE_FLAG"` env var with value `"ENABLED"` in all environments where you want the flag enabled.
 Then add `flag_name` to the list of flags  in `app/lib/feature_flags.rb`.
 
-To use the feature flag in your code, just call `FeatureFlags.enabled?(:flag_name)`.
+When adding a `flag_name` to the list of static flags, you will need to decide if this is a `"global"` flag i.e. always taken from the env var and not the session, or a `"session"` flag i.e. taken from the `session_data` of the check. 
+
+We introduced this as a way of making our feature flags 'backwards compatible' - if a user check is underway during the switch-on of a flag, their user journey will not be affected and they can continue through their check with the flag value that was set at the beginning of their journey.
+
+To use the feature flag in your code, just call `FeatureFlags.enabled?(:flag_name)`. There is an optional `session_data` argument which you must use if you wish to derive the flag value from the `session_data`. Usage is `FeatureFlags.enabled?(:flag_name, session_data)`.
 
 In tests, you can temporarily enable a feature flag by setting the ENV value.
 However, flags are _not_ reset between specs, so to avoid polluting other tests use an `around` block and change the ENV value back once the test has run.
--- a/README.md
+++ b/README.md
@@ -148,9 +148,9 @@ Then add `flag_name` to the list of flags  in `app/lib/feature_flags.rb`.
 
 When adding a `flag_name` to the list of static flags, you will need to decide if this is a `"global"` flag i.e. always taken from the env var and not the session, or a `"session"` flag i.e. taken from the `session_data` of the check. 
 
-We introduced this as a way of making our feature flags 'backwards compatible' - if a user check is underway during the switch-on of a flag, their user journey will not be affected and they can continue through their check with the flag value that was set at the beginning of their journey.
+We introduced this as a way of making our feature flags 'backwards compatible' - if a user check is underway during the switch-on of a flag, their user journey will not be affected by any flag-related changes. This is because we use their `session_data` to determine the value of the flag that was set at the start of their check. 
 
-To use the feature flag in your code, just call `FeatureFlags.enabled?(:flag_name)`. There is an optional `session_data` argument which you must use if you wish to derive the flag value from the `session_data`. Usage is `FeatureFlags.enabled?(:flag_name, session_data)`.
+To use the feature flag in your code, call `FeatureFlags.enabled?(:flag_name, session_data)`. For cases where you are not able to pass in `session_data` e.g. on the start page, call `FeatureFlags.enabled?(:flag_name, without_session_data: true)`.
 
 In tests, you can temporarily enable a feature flag by setting the ENV value.
 However, flags are _not_ reset between specs, so to avoid polluting other tests use an `around` block and change the ENV value back once the test has run.
--- a/app/controllers/estimates_controller.rb
+++ b/app/controllers/estimates_controller.rb
@@ -66,7 +66,7 @@ private
   def load_session_derived_flags
     feature_flags = {}
     FeatureFlags::STATIC_FLAGS.select { |_, v| v == "session" }.map do |flag|
-      feature_flags[flag.first.to_s] = FeatureFlags.enabled?(flag.first)
+      feature_flags[flag.first.to_s] = FeatureFlags.enabled?(flag.first, without_session_data: true)
     end
 
     feature_flags
--- a/app/controllers/estimates_controller.rb
+++ b/app/controllers/estimates_controller.rb
@@ -3,7 +3,7 @@ class EstimatesController < ApplicationController
 
   def new
     new_assessment_code = SecureRandom.uuid
-    session[assessment_id(new_assessment_code)] = {}
+    session[assessment_id(new_assessment_code)] = { "feature_flags": load_session_derived_flags }
     redirect_to estimate_build_estimate_path new_assessment_code, Steps::Helper.first_step
   end
 
@@ -61,5 +61,14 @@ private
 
   def track_completed_journey(calculation_result)
     JourneyLoggerService.call(assessment_id, calculation_result, @check, cookies)
+  end
+
+  def load_session_derived_flags
+    feature_flags = {}
+    FeatureFlags::STATIC_FLAGS.select { |_, v| v == "session" }.map do |flag|
+      feature_flags[flag.first.to_s] = FeatureFlags.enabled?(flag.first)
+    end
+
+    feature_flags
   end
 end
--- a/app/controllers/estimates_controller.rb
+++ b/app/controllers/estimates_controller.rb
@@ -3,7 +3,7 @@ class EstimatesController < ApplicationController
 
   def new
     new_assessment_code = SecureRandom.uuid
-    session[assessment_id(new_assessment_code)] = { "feature_flags": load_session_derived_flags }
+    session[assessment_id(new_assessment_code)] = { "feature_flags" => load_session_derived_flags }
     redirect_to estimate_build_estimate_path new_assessment_code, Steps::Helper.first_step
   end
 
--- a/app/lib/feature_flags.rb
+++ b/app/lib/feature_flags.rb
@@ -3,10 +3,17 @@ class FeatureFlags
     example_2125_flag: { from: "2125-01-01", public: false },
   }.freeze
 
-  STATIC_FLAGS = %i[sentry cw_forms special_applicant_groups self_employed public_beta].freeze
+  # the values of some feature flags will come from the session and not the env variables.
+  # "global" - feature flag value should be derived from the env variable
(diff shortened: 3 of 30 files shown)


spec/models/feature_flags_spec.rb
```diff
@@ -13,6 +13,30 @@
...
+        ENV["SENTRY_FEATURE_FLAG"] = "enabled"
+        each.run
+        ENV["SENTRY_FEATURE_FLAG"] = "disabled"
+      end
+
+      it "defaults to global flag when the flag does not exist in the session_data" do
+        expect(described_class.enabled?(:sentry, session_data)).to eq true
+      end
+
+      context "when flag is specified in the session_data" do
```
willc-work, line 30: would this test be more appropriate if it used a session based flag? Sentry is a global flag that will always look to the env_var and not the session data.
However if we are using sentry flag because we dont want to have to update this test when the more ephemeral flags are removed, im fine with that.

willc-work (approved):
one query on a test, otherwise looks great

README.md
```diff
@@ -144,7 +144,11 @@ For "static" feature flags we set the flag values in env vars.
 To add a new feature flag, set a `"#{flag_name.upcase}_FEATURE_FLAG"` env var with value `"ENABLED"` in all environments where you want the flag enabled.
 Then add `flag_name` to the list of flags  in `app/lib/feature_flags.rb`.
 
-To use the feature flag in your code, just call `FeatureFlags.enabled?(:flag_name)`.
+When adding a `flag_name` to the list of static flags, you will need to decide if this is a `"global"` flag i.e. always taken from the env var and not the session, or a `"session"` flag i.e. taken from the `session_data` of the check. 
+
+We introduced this as a way of making our feature flags 'backwards compatible' - if a user check is underway during the switch-on of a flag, their user journey will not be affected and they can continue through their check with the flag value that was set at the beginning of their journey.
+
+To use the feature flag in your code, just call `FeatureFlags.enabled?(:flag_name)`. There is an optional `session_data` argument which you must use if you wish to derive the flag value from the `session_data`. Usage is `FeatureFlags.enabled?(:flag_name, session_data)`.
```
exonian, line 151: If I've understood the code correctly (adding the flag value to the session data is only done for session flags, you can look for a global flag on session data first you just won't find it) I wonder if it would be easier to comprehend if we said that `session_data` arg wasn't optional at all?

My concern is that someone reading the code might think that the passing of `session_data` to the function call _does_ something and thus is a bit of a trap (and obviously if you don't pass it to one that's a session flag you don't get the correct behaviour, which again could be a trickier one to debug).

spec/models/feature_flags_spec.rb
exonian: > that will always look to the env_var and not the session data

Don't we always look in the session data first? (and for global flags the difference is it normally wouldn't have been put there?)
I think this test does a good job of simulating a session flag at lookup time. I do think we could maybe do with a test of what flags get set in the session data on check creation, though

HettieS: @willc-work yes, the reason I went for the sentry flag (even though it technically wouldn't appear in the session data) is due to it being unlikely that we would need to update the test - that flag should be constant and other flags are temporary and get deleted over time. But I'm definitely open to suggestions of a better way to handle that!

@exonian good shout, I'll see if I can create a test to check which flags get set in the check creation

app/lib/feature_flags.rb
```diff
@@ -3,21 +3,28 @@ class FeatureFlags
...
-  STATIC_FLAGS = %i[sentry cw_forms special_applicant_groups self_employed].freeze
+  # the values of some feature flags will come from the session and not the env variables.
+  # "global" - feature flag value should be derived from the env variable
+  # "session" - feature flag value should be derived from the session_data of the check
+  STATIC_FLAGS = { sentry: "global", cw_forms: "global", special_applicant_groups: "session", self_employed: "session" }.freeze
 
   class << self
-    def enabled?(flag)
+    def enabled?(flag, session_data = nil)
+      if session_data && session_data["feature_flags"]&.key?(flag.to_s)
```
exonian, line 13: I wonder if we should have the same checking the key is a recognised one here as we do for global ones?

README.md
HettieS: Yes I was also not sure on the optional arg and agree with you - the reason it is optional currently, is that when the application loads, we have flags set on the start page that need to determine what to show, before a check has been created. So we can't use a required check argument with `def enabled?`

The other option that @patrick-laa and I discussed was creating two methods instead, something like `def globally_enabled?` and `def session_enabled?` which would have relevant code and required arguments, and you'd have to pick which one you're using depending on the use case. What do you think? Happy to look at implementing that if you think it is better but we wondered if it might be overkill. You're right though, it would be easy not to pass in session_data when it is needed.

exonian: Ooh yes interesting, I hadn't thought of that edge case! I wonder if we could keep it technically optional but conventionally required ie we say "if you have a session with a check in, pass it, always"?

The other option, with different lookup methods, sounds appealing but does take us further from the simple, agnostic lookup. Maybe that's a good thing? Being explicit at lookup time. Might make the behaviour easier to understand, as it's closer to where it's used.

This is absolutely not my ultimate call though, so I'll just leave it at highlighting the risks I see and my thoughts from looking at the code in isolation, and totally trust you folk to implement what's best given that info 👍

patrick-laa: How about there are 3 arguments to enabled - `flag_name` (mandatory), `session_data` (can be nil), and a keyword argument `i_understand_that_i_should_always_be_providing_session_data_except_in_very_particular_circumstances` which defaults to false. So this is valid:

```ruby
FeatureFlags.enabled?(:foo, session_data)
```

And so is this:

```ruby
FeatureFlags.enabled?(:foo, nil, i_understand_that_i_should_always_be_providing_session_data_except_in_very_particular_circumstances: true)
```

But this will raise an exception:

```ruby
FeatureFlags.enabled?(:foo, nil)
```

That way it's very hard for someone to accidentally do the wrong thing and not notice.

app/controllers/estimates_controller.rb
```diff
@@ -73,4 +73,13 @@ def load_check
   def track_completed_journey(calculation_result)
     JourneyLoggerService.call(assessment_id, calculation_result, @check, cookies)
   end
+
+  def load_session_derived_flags
+    feature_flags = {}
+    FeatureFlags::STATIC_FLAGS.select { |_k, v| v == "session" }.map do |flag|
```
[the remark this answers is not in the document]
HettieS: yep, good idea to be consistent if this is how we do it in the majority of cases in the codebase - will change

app/lib/feature_flags.rb
HettieS: I might not be thinking this through correctly, but I'm not sure how that would work - the 'flag unrecognised' fallback is still the final fallback in this code. If the flag isn't found in the `session_data`, the code will fall back to the `STATIC_FLAGS`, then the `ENABLED_AFTER_DATE` flags and then finally will fall back to the error if the flag can't be found in any of those places. But I might be misunderstanding!

README.md
HettieS: @patrick-laa I quite like this idea of passing in the additional argument. It makes it very hard to get it wrong without realising. Will work with that and see what it looks like.

spec/features/feature_flags_spec.rb
```diff
@@ -45,19 +45,33 @@
...
       visit edit_feature_flag_path("sentry")
       choose "No"
       click_on "Save and continue"
       expect(page).to have_content "sentryNo"
-      expect(FeatureFlags.enabled?(:sentry)).to eq false
+      expect(FeatureFlags.enabled?(:sentry, without_session_data: true)).to eq false
+    end
+  end
+
+  context "when setting feature flags in the session" do
```
HettieS, line 64: wasn't sure where to put this test so ended up putting it in this file - also I'm not too sure about creating the example flag purely for testing. But seems to be the safest way to get this test passing without having to depend on temporary flags. It means though that there will always be a `feature_flags => { example => false }` on our session data

README.md
HettieS: I've implemented the above now and updated the README so I think this is ready for another review if possible!

README.md
```diff
@@ -146,7 +146,11 @@ For "static" feature flags we set the flag values in env vars.
...
 
-To use the feature flag in your code, just call `FeatureFlags.enabled?(:flag_name)`.
+When adding a `flag_name` to the list of static flags, you will need to decide if this is a `"global"` flag i.e. always taken from the env var and not the session, or a `"session"` flag i.e. taken from the `session_data` of the check. 
+
+We introduced this as a way of making our feature flags 'backwards compatible' - if a user check is underway during the switch-on of a flag, their user journey will not be affected by any flag-related changes. This is because we use their `session_data` to determine the value of the flag that was set at the start of their check. 
+
+To use the feature flag in your code, you will need to call `def enabled?`. In order to check if a flag is enabled on a global level, you will need to call `FeatureFlags.enabled?(:flag_name, without_session_data: true)`. In order to use the flag value set in the session, call `FeatureFlags.enabled?(:flag_name, session_data)`. `session_data` is an optional argument. If you do not wish to pass in `session_data` to determine the flag value, you must confirm this by passing in `without_session_data: true`.
```
exonian, line 153: The bit about passing in `without_session_data: true` for global checks is confusing to me. From the discussion, I would expect that only to be used when you _couldn't_ pass session data, which to me is a bit different.

README.md
exonian: Nice approach

README.md
HettieS: I think I have written this README in a confusing way. So yes I guess the only time you wouldn't pass session_data is when you don't have it, e.g. on the start pages. So in those cases you'd have `FeatureFlags.enabled?(:flag, without_session_data: true)` - if you just pass `FeatureFlags.enabled?(:flag)` you'll get an error.

On reflection I have written something really confusing and we should just say 'always pass `session_data` if you can'.

app/lib/feature_flags.rb
exonian: So I think there's two different things:

One is where the code asks for a flag, and it's not in session data, so it falls back as you say and we get the goodness like error handling and defaults.

The other was that I saw `if STATIC_FLAGS.key?(flag)` and interpreted it as us wanting to only look for allow-listed flag names in the env, and therefore figured we'd want to only look in the session for allow-listed feature flag names. But now looking at it more I think maybe that's not the reasoning at all, so it's fine.

README.md
HettieS: I guess I was thinking there might be cases where you know you only care about a global flag value, which won't be in the `session_data` so there isn't any point to passing that in. But that probably just makes things less clear!

HettieS: I'll reword this 😅

HettieS: @exonian have updated the README, let me know if it now sounds clearer!

patrick-laa (approved):
Looks good to me!

PR #772, opened 2023-06-29
EL-988: pdf metadata

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-988)

## What changed and why

Use exiftool to modify the metadata of generated pdfs to ensure the language is set for accessibility purposes

## Guidance to review

I tried to use PDFTK but its metadata reporting/modification doesn't appear to include language.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,6 @@ see https://stackoverflow.com/questions/71262775/how-do-i-ensure-assets-are-pres
   - postgres
   - yarn
   - pdftk
-  - exiftool
 
 ## Setting up the app
 
@@ -39,12 +38,6 @@ You will be prompted on the command line to start the server with something like
 
 ```
 brew services start postgresql@14
-```
-
-Next install exiftool:
-
-```
-brew install exiftool
 ```
 
 You will also need pdftk. There is a [Mac installer](https://www.pdflabs.com/tools/pdftk-the-pdf-toolkit/pdftk_server-2.02-mac_osx-10.11-setup.pkg) for convenience.
--- a/README.md
+++ b/README.md
@@ -20,8 +20,8 @@ see https://stackoverflow.com/questions/71262775/how-do-i-ensure-assets-are-pres
 - System dependencies
   - postgres
   - yarn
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -21,6 +21,7 @@ see https://stackoverflow.com/questions/71262775/how-do-i-ensure-assets-are-pres
   - postgres
   - yarn
   - pdftk [Mac installer](https://www.pdflabs.com/tools/pdftk-the-pdf-toolkit/pdftk_server-2.02-mac_osx-10.11-setup.pkg)
+  - exiftool (Available with `brew install exiftool`)
```
MazOneTwoOne, line 24: Do we want to add how to install `pdftk` & `exiftool` in the 'Setting up the app' section, as we have with `postgres` & `yarn`?

patrick-laa: Good point. I've rejigged the readme to make it more consistent.

app/controllers/estimates_controller.rb
```diff
@@ -39,25 +39,11 @@ def download
...
-      display_url: request.url.split("/estimates").first,
-      execute_script: "document.querySelectorAll('button').forEach(el => el.style.display = 'none')",
-    }
-
-    pdf = Grover.new(html, **grover_options).to_pdf
-
-    send_data pdf,
-              filename: "#{I18n.t('generic.download_name')} - #{Time.zone.now.strftime('%Y-%m-%d %H.%M.%S')}.pdf",
-              type: "application/pdf"
+    PdfService.with_pdf_data_from_html_string(html, request.url.split("/estimates").first) do |pdf_data|
```
MazOneTwoOne, line 42: Is `request.original_url` a more up to date version of `request.url`?

MazOneTwoOne (approved):
Just the one comment, happy to approve.

app/controllers/estimates_controller.rb
patrick-laa: So `url` uses `fullpath` and `original_url` uses `original_fullpath`, and as far as I can tell, sometimes `fullpath` can be modified by Rails internals while `original_fullpath` is always left untouched. It doesn't really matter in this case because we're ignoring the path bit. So in fact we can just call `request.base_url` (which both `url` and `original_url` use), since that's the only bit we want anyway!

PR #773, opened 2023-06-29
BUG: Main UAT deployments failing

patrick-laa:
## What changed and why

Previously, both branch deployments and main UAT deployments used the same `helm upgrade` command that explicitly set some dynamic values with `--set` arguments. However, now main UAT deployments use the `helm upgrade` command that staging and production deployments use, which assumes those values will be available in the values yaml file. So to get it working, we need the right values in the UAT values yaml file (which will still be overridden by the dynamic values provided on branch deployments).

Added viable default values to uat.yaml

## Checklist


exonian (approved):
Seems reasonable

PR #775, opened 2023-06-29
EL-984 Pensioner disregard alert

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-984)

## What changed and why

Add an alert for pensioner disregard on the CW form selection page. This is because it is not obvious from the form that the calculation has been taken into account.

## Checklist


patrick-laa (approved):
Looks good to me. On balance I think it's the right thing not to DRY up the banner partial here with the change log banners, as this gives us flexibility at the cost of very little duplication.

PR #785, opened 2023-07-03
Self-employed testing - do not merge

patrick-laa:


patrick-laa:
Closing now self-employed is in production

PR #787, opened 2023-07-04
EL-985: Information not announced and not focusable on iOS

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-985)

## What changed and why

There are 2 parts on this ticket:
1. Making the figures announce-able after the '£' sign
2. [Making Information focusable on mobile phones/tablets](https://www.youtube.com/watch?v=qDm7GiKra28)

Part 1 is done, but unable to come up with a sensible implementation for Part 2 (see below).

## Guidance to review

Part 2 findings and approaches:

- When generating a PDF with puppeteer/grover, ios screen readers have trouble focussing on html elements which aren't `<ul>, <ol>` or `<tables>` 
   - this affects mobile and tablet users who are using 'Voiceover'
   - ascertained this by removing all classes and text, and generating a PDF with several html elements, to see which were focusable.

- In this implementation have changed the `<p> /<h2>` elements to `<ul>/<ol>` (overriding the gov.uk component style) which then makes element focusable. However, this current implementation will make `<p>` elements announced as list elements.
   - this creates a copy of the exact same page one for screen view and one for PDF view as removing of styles did not work for screen view.
   - very open to opinions on how to apply global stylings as coed is not ver DRY - have tried several approaches (reusable method, adding to stylesheet) 
   - **TO NOTE:** when generating a PDF from an Asylum eligibility check the ‘Your answers’ section which uses `<dl>/<dt>` tags are no longer focusable - but for all other scenarios are. Can change layout to use `<ul>` elements, but want to check this approach isn’t too crazy, before attempting. 

- Have also attempted to see there if there are any obvious changes which could be made on the _app/views/layouts/print_application.html.slim_ file. Nothing obvious 
   - this included setting doctype html & messing with character encoding. 

## Checklist


app/views/estimates/_capital_table.html.slim
```diff
@@ -32,10 +32,10 @@
         - vehicle.each do |key, value|
           - body.with_row do |row|
             - row.with_cell(header: true, text: t("estimates.show.vehicle.#{key}"))
-            - row.with_cell(text: value, numeric: true)
+            - row.with_cell(text: value, html_attributes: { class: "govuk-!-text-align-right" })
```
patrick-laa, line 35: Ok, so basically the cells with numbers in, by removing `numeric: true` we're causing them _not_ to have the class `govuk-table__cell--numeric` applied, and instead we apply `govuk-!-text-align-right`. I believe you that that works, but I don't know _why_ it would work. I guess the original class applies `font-variant-numeric: tabular-nums` and `font-feature-settings: "tnum" 1;`. Those are unusual - could they be having an odd interaction with puppeteer when generating the PDF?

We probably don't _need_ to know the answer, but at the very least we should probably document somewhere why we're making this change. As a comment, or in the readme. Or maybe best to create a reusable method called `def pdf_friendly_numeric_table_cell(row, value, additional_classes = [])` and call that every time we want to put a number in a table cell?

MazOneTwoOne: Sorry @patrick-laa not quite following the reusable method. 

- Placed this in the `def pdf_friendly_numeric_table_cell(row, value, additional_classes = []) end` in the _app/helpers/results_helper.rb_ file.


- Calling it like so:
```
- body.with_row do |row|
  ruby:
    text = header_with_hint("estimates.show.client_income_types.#{income_type}", "estimates.show.client_income_hints.#{income_type}")
  - row.with_cell(header: true, text:)
  - pdf_friendly_numeric_table_cell(row.with_cell, text: money_amount, html_attributes: { class: "govuk-!-text-align-right" })

```

patrick-laa: @MazOneTwoOne so I'm thinking it would be:

```slim
- pdf_friendly_numeric_table_cell(row, money_amount)
```

So the thing that calls the method doesn't need to know anything about CSS classes or what specifically it has to _do_ to make it pdf-friendly.

If there were other CSS classes you needed in other places you could pass them in, e.g:

```slim
- pdf_friendly_numeric_table(row, @model.household_vehicle_assessed_value, ["govuk-!-font-weight-bold"])
```

patrick-laa:
Ok, this is audacious, and I like that. I'm still on the fence about whether this is The Way, but if it is, there are some tweaks to make:

First, you need to decide whether or not the "print" view and the "pdf" view should contain any shared elements. At the moment it looks like you've got separate-but-parallel partials... but if the pdf isn't using any of the "print" stuff, then you'll want to undo all the change you've made to files like `app/views/estimates/_capital_table.html.slim` - there's no point making them "pdf-friendly" as you're not using them in the PDF.

Second, when it comes to the ways in which you're messing with HTML elements to make them come out right in the PDF. There's a lot of copy-pasting going on. For example, for the H2s you're going through and adding inline styles `list-style-type: none; margin: 0;` everywhere. Which is going to be a right pain if you ever need to slightly tweak that style. So step one might be to create a CSS class:

```
.pdf-friendly-h2 {
  list-style-type: none;
  margin: 0;
}
```

And then you can just apply that style.  Likewise, instead of manually putting an `li` element inside the h2, it would be tidier if you had a helper method that maybe looked like `def pdf_friendly_h2(text)`, and that method could worry about creating the HTML structure you need, so you don't need to keep on manually adding `li` elements everywhere.

Which brings me on to the third point: By duplicating all the views, you're opening yourself up to a lot of pain the next time we want to slightly change the results page. Because suddenly we have to make every change in 2 places, and worse, if we forget, we might not notice for _months_, because we don't often check the PDF. So the question is, can we avoid that?

Well, what if in the `download` method we set a variable `@is_pdf = true`? And then what if we used the same views for download as for print, but in those views, every time we want an H2, we call a helper, passing in `@is_pdf` as an argument? E.g. `= result_page_h2(t("estimates.show.vehicles", number: count), is_pdf: @is_pdf)`

Then inside `def result_page_h2(text, is_pdf: false)` we do something like:

```
if is_pdf
  render "shared/pdf_accessible_h2", text:
else
  render "shared/h2", text:
end
```

Finally we create a partial called `_pdf_accessible_h2.html.slim` that looks something like:

```
  h2.pdf-friendly-h2
    li = text
```

That _might_ be a bridge too far, as we'd need to make sure it worked not just for all `h2`s but also something similar for all `p`s as well. But if we can pull it off, that gives us something self-documenting and flexible. If it doesn't work, points 1 and 2 still stand, I think.

patrick-laa:
Also, I tried out the PDF generated by this branch on https://check.axes4.com/en/ and it looks like the number of errors reported there has gone up. I haven't figured out yet how to get more specific information from it, but we should tread carefully here to make sure we don't fix one problem by creating several others.

MazOneTwoOne:
> Ok, this is audacious, and I like that. I'm still on the fence about whether this is The Way, but if it is, there are some tweaks to make:
> 
> First, you need to decide whether or not the "print" view and the "pdf" view should contain any shared elements. At the moment it looks like you've got separate-but-parallel partials... but if the pdf isn't using any of the "print" stuff, then you'll want to undo all the change you've made to files like `app/views/estimates/_capital_table.html.slim` - there's no point making them "pdf-friendly" as you're not using them in the PDF.
> 
> Second, when it comes to the ways in which you're messing with HTML elements to make them come out right in the PDF. There's a lot of copy-pasting going on. For example, for the H2s you're going through and adding inline styles `list-style-type: none; margin: 0;` everywhere. Which is going to be a right pain if you ever need to slightly tweak that style. So step one might be to create a CSS class:
> 
> ```
> .pdf-friendly-h2 {
>   list-style-type: none;
>   margin: 0;
> }
> ```
> 
> And then you can just apply that style. Likewise, instead of manually putting an `li` element inside the h2, it would be tidier if you had a helper method that maybe looked like `def pdf_friendly_h2(text)`, and that method could worry about creating the HTML structure you need, so you don't need to keep on manually adding `li` elements everywhere.
> 
> Which brings me on to the third point: By duplicating all the views, you're opening yourself up to a lot of pain the next time we want to slightly change the results page. Because suddenly we have to make every change in 2 places, and worse, if we forget, we might not notice for _months_, because we don't often check the PDF. So the question is, can we avoid that?
> 
> Well, what if in the `download` method we set a variable `@is_pdf = true`? And then what if we used the same views for download as for print, but in those views, every time we want an H2, we call a helper, passing in `@is_pdf` as an argument? E.g. `= result_page_h2(t("estimates.show.vehicles", number: count), is_pdf: @is_pdf)`
> 
> Then inside `def result_page_h2(text, is_pdf: false)` we do something like:
> 
> ```
> if is_pdf
>   render "shared/pdf_accessible_h2", text:
> else
>   render "shared/h2", text:
> end
> ```
> 
> Finally we create a partial called `_pdf_accessible_h2.html.slim` that looks something like:
> 
> ```
>   h2.pdf-friendly-h2
>     li = text
> ```
> 
> That _might_ be a bridge too far, as we'd need to make sure it worked not just for all `h2`s but also something similar for all `p`s as well. But if we can pull it off, that gives us something self-documenting and flexible. If it doesn't work, points 1 and 2 still stand, I think.


1. Have removed the need for a 'pdf_friendly_numerical_cell' method, by adding in these bits when the PDF is generated  by grover. Helps with reading out numbers in cells on iOS devices. 
```
    prefer_css_page_size: true,    
    print_background: true,
    vision_deficiency: "deuteranopia",
    media_features: [{ name: "prefers-color-scheme", value: "dark" }],
    wait_until: "domcontentloaded",
```

2. Agreed. Have made some more methods to do this. Wasn't able to add additional CSS classes to the gov.uk ones. Was easier to write a method with that info in. There were two files where I couldn't write a 'friendly pdf helper method' so did it the long way;

- app/views/shared/_heading.html.slim
- app/views/estimates/_capital_table.html.slim


3. Agreed. Have removed duplication of partials and PDF styles are shown when a PDF is made. 


4. **TO NOTE:** As stated in the 'guidance for review' when generating a PDF from an Asylum eligibility check the ‘Your answers’ section which uses `<dl>/<dt>` tags are no longer focusable - but for all other scenarios are. Not sure it's possible to change this gov.uk component to the `<ul>` structure.

MazOneTwoOne:
> Also, I tried out the PDF generated by this branch on https://check.axes4.com/en/ and it looks like the number of errors reported there has gone up. I haven't figured out yet how to get more specific information from it, but we should tread carefully here to make sure we don't fix one problem by creating several others.

Agreed. Maybe this exercise gives us more evidence that we need another approach to generating PDFs or another pair of eyes is needed on this ticket?

patrick-laa (commented):
I think we're very nearly there. This feels like a proportionate approach that's fairly easy to understand in code. Just a few tweaks to tidy it up left.

app/helpers/results_helper.rb
```diff
@@ -6,4 +6,62 @@ def header_with_hint(header_key, hint_key, hint_args: {})
                 class: "govuk-hint govuk-!-margin-top-1 govuk-!-margin-bottom-0")
     ).html_safe
   end
+
+  def pdf_friendly_h1(text, additional_style, is_pdf)
+    if is_pdf
+      "<h1 class=\"govuk-heading-xl #{additional_style}\"><li>#{text}</li></h1>"
```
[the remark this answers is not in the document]
patrick-laa: Oops, this was a hangover from a review I started and then we spoke about directly. Pls ignore this thread!

patrick-laa (approved):

PR #795, opened 2023-07-07
EL-986 make pdf titles single blocks

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-986)
as covered in the comments, the problem is related to ligatures (@exonian I believe you called this as a potential problem early on, bit I evidently missed that). The css fix to the print page prevents the split on `fi `character pairs

removed the partial being used to display the service name as a H1 and added a H2 directly to the print page. 

as per:
https://usability.yale.edu/web-accessibility/articles/headings
https://www.w3.org/WAI/tutorials/page-structure/headings/#main-heading-after-navigation
 there is no problem with a H2 appearing before a H1 and the H1 should be the most important part, which for us is the result.

The issue with the H2 being split  (1.4.1) cannot be exactly replicated. The fix for 1.3.1 also prevents splits on this H2 but the PDF splits anything that crosses more than 1 line. For spans this isn't really an issue but for headings this will result in each line being announced as a heading. The fix for this in this case is to prevent wrapping. It doesnt cause any issues with the rest of the display, although the heading looks slightly smaller on the page.

## Checklist

### DIFF
--- a/app/assets/stylesheets/print.scss
+++ b/app/assets/stylesheets/print.scss
@@ -33,3 +33,9 @@
                               // regardless of class-based button styles
   }
 }
+// Grover/pupetteer has a bug https://github.com/puppeteer/puppeteer/issues/4125
+// this ensures that the bug does not cause the titles in H1 and H2 elements to
+// split on the characters 'fi'
+.no-ligatures{
+  font-variant-ligatures: none;
+}
--- a/app/services/pdf_service.rb
+++ b/app/services/pdf_service.rb
@@ -9,6 +9,10 @@ class PdfService
     },
     prefer_css_page_size: true,
     print_background: true,
+    viewport: {
+      width: 2400,
+      height: 4800,
+    },
     emulate_media: "screen",
     launch_args: ["--font-render-hinting=medium", "--no-sandbox", "--force-renderer-accessibility"],
     execute_script: "document.querySelectorAll('button').forEach(el => el.style.display = 'none')",
--- a/app/views/estimates/print.html.slim
+++ b/app/views/estimates/print.html.slim
@@ -3,7 +3,7 @@
     = t("estimates.show.not_likely_to_qualify")
   - else
     = t("estimates.show.likely_to_qualify")
-.govuk-grid-column-full
+.govuk-grid-column-full.no-ligatures
   .gem-c-organisation-logo
     .gem-c-organisation-logo__container
       = pdf_friendly_logo(@is_pdf)
@@ -12,7 +12,7 @@
 
   button.govuk-button.govuk-button--secondary data-module="govuk-button" data-trigger="print" = t(".print_this_page")
 
-  = render "shared/heading", header_text: t("service.name")
+  h2.govuk-heading-l = t("service.name")
   = render "result_panel_content"
 
   - if @model.level_of_help == "certificated"


patrick-laa:
@willc-work have you verified that this fixes the accessibility issue? It looked from the ticket like the splits being reported in the accessibility audit didn't necessarily line up with visible line breaks, so I wasn't sure if fixing the line breaks would fix the a11y problem.

app/views/estimates/print.html.slim
```diff
@@ -44,7 +44,8 @@
     = render "evidence"
 
   - if @model.calculated?(:gross_income) || @model.calculated?(:disposable_income) || @model.calculated?(:capital)
-    h2.govuk-heading-m = t("estimates.show.#{@model.level_of_help}_how_calculated")
+    h2.govuk-heading-m style="white-space: nowrap;"
```
patrick-laa, line 47: FWIW this will raise a CSP error on the print preview page, as inline styles are no longer allowed.

app/services/pdf_service.rb
```diff
@@ -10,6 +10,8 @@ class PdfService
     emulate_media: "screen",
     launch_args: ["--font-render-hinting=medium", "--no-sandbox"],
     execute_script: "document.querySelectorAll('button').forEach(el => el.style.display = 'none')",
+    media_features: [{ name: "prefers-color-scheme", value: "dark" }],
```
patrick-laa, line 13: Is this related to the fix or does it have some other effect?

willc-work: using just `scale:` changes the scale but not the splitting of the title into multiple blocks (even at 0.5/1 where the text fits comfortably on the line). adding `media_features` prevents the title being split into multiple sections when the text will fit on a single line, or if it doesn't fir on a single line, each line is its own block (as opposed to the current situation where it is chunked up and split mid-word)

willc-work:
@patrick-laa the ticket indicates that the issue is that the title is split into 3 sections, each of which are given a H1, so there is a single heading but the screenreader splits it into 3 and announces each one e.g
 "Heading 1 check if your client quali"
"Heading 1 fies for" ....
I dont have access to NVDA and Voiceover doesnt seem to like PDFs. So my assumption is that this fixes the issue with the first title because now the PDF has a single span containing the title, so should be read out as "Heading 1 check if your client qualifies for legal aid"

patrick-laa:
@willc-work the screenshot suggests that the split is "Check if your client quali" "fi" "es for legal aid". If this split was due to line breaks I'd be surprised, simply because I don't think any linebreak algorithm would break "qualifies" into 3 like that. So I'm not sure that line breaks are the source of the problem.

patrick-laa:
@MazOneTwoOne do you now have working Adobe? If so, can you download a PDF from https://el-986-pdf-a11y-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk/ and confirm that the reported bizarre "fi" split etc is fixed?
image-20230622-101259

willc-work:
@patrick-laa if I only have the line:
`media_features: [{ name: "prefers-color-scheme", value: "dark" }]`
and remove `scale` , this fixes the weird line break, so that line breaks are applied at the end of the line. But I am not sure if this will fix the screenreader issue as it might still read out `Heading 1` before each line.

MazOneTwoOne:
> @MazOneTwoOne do you now have working Adobe? If so, can you download a PDF from https://el-986-pdf-a11y-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk/ and confirm that the reported bizarre "fi" split etc is fixed? image-20230622-101259

MazOneTwoOne:
> @MazOneTwoOne do you now have working Adobe? If so, can you download a PDF from https://el-986-pdf-a11y-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk/ and confirm that the reported bizarre "fi" split etc is fixed? image-20230622-101259

Hey Both,

Still don't have Adobe Pro (for accessibility checks/tree) or NVDA (Windows screen reader) to verify this. 

What I did notice, is that these changes have also fixed iOS screen readers, being able to announce numbers but added in another issue with iOS screen readers. 

- If the table goes over the the page then the iOS screen reader can announce the rest of the table on the other page when opening with Adobe.

willc-work:
@patrick-laa ok I have signed up for a 14 day free trial at https://assistivlabs.com/ and done some more testing. Even though I am able to get the title into a single span locally, when I try it on UAT it splits it again into 3 sections it also does this on the second heading mentioned on the ticket. 

As noted the split is really weird, but in both sections `fi` is set on its own. Is it possible that whatever is splitting the words is reading `fi` as an end statement as in LINUX if statement?
https://www.thegeekstuff.com/2010/06/bash-if-statement-examples/

patrick-laa:
@willc-work looks like someone else experienced 'fi' weirdness - https://github.com/puppeteer/puppeteer/issues/4125#issuecomment-489107275, and suggested `font-variant-ligatures: none;` in CSS would fix. If this is our issue, props to @exonian who suggested it was a ligature issue ages ago.

Although that would only explain the h1 issue, not the h2 issue (and no idea why the h2, which also has a "fi", doesn't also have a split there)

app/assets/stylesheets/print.scss
```diff
@@ -32,4 +32,10 @@
     display: none !important; // In no circumstances do we want buttons to appear when the page is printed,
                               // regardless of class-based button styles
   }
-}
\ No newline at end of file
+}
+// Grover/pupetteer has a bug https://github.com/puppeteer/puppeteer/issues/4125
+// this ensures that the bug does not cause the titles in H1 and H2 elements to
+// split on the characters 'fi'
+* {
```
patrick-laa, line 39: We might want to make it so that this file is only included in the print layout, because at the moment this will affect every element on every screen

patrick-laa (approved):
OK cool, so if this solves the `fi` split mentioned in 1.3.1 of the report that's great. Do we think that the h2 issue they mentioned 1.4.1, which is a split without a 'fi', is completely separate?

Also, they say in the 1.3.1 bit that it would be better if there was only one h1 element. With this fix we've got it down from 4 to 2. But there actually are 2 h1 elements on the page - both "Check if your client qualifies...." and "Your client is likely to qualify for...". Would it be a good idea, as a bonus thing, to fix it so that the first one is an h1 and the second is an h2?

willc-work:
The h2 issue is the same issue isn't it? The report says it is split in 3 parts but I am actually seeing it split into 4 parts, with the 2 instances of 'fi' that occur in 'fi'nacially and certi'fi'cated causing the splits. This fix works on that sentence too.

re: the the multiple h1's I will go back and have a look at that.

patrick-laa:
@willc-work the problem with a `no-wrap` is that this CSS also controls the "print" view, and it means on the print preview screen, and if you try to print, the text runs off the side of the page:
Screenshot 2023-07-13 at 09 44 33

willc-work:
@patrick-laa the issue with the no-wrap appears to be browser specific. I have added in a scale: attribute to grover_options so that the page is scaled in browsers that dont automatically do this so that all of the wording fits on the line and is printed as exopected

app/services/pdf_service.rb
```diff
@@ -10,6 +10,7 @@ class PdfService
     emulate_media: "screen",
     launch_args: ["--font-render-hinting=medium", "--no-sandbox"],
     execute_script: "document.querySelectorAll('button').forEach(el => el.style.display = 'none')",
+    scale: 0.75 / 1.0,
```
patrick-laa, line 13: This doesn't fix the issue with trying to print the page on Chrome that I screenshotted before (because printing doesn't involve PdfService at all).

I've got another theory. The line breaks were occurring because when grover/puppeteer "looks at" the page, it's using a small viewport, narrow enough that the h2 line was being broken up. But it didn't actually insert visible linebreaks at those points, they only appeared in the underlying structure exposed by the accessibility tool. So we can infer that: _the size of the viewport affects accessibility data without affecting the look of the page_.

So how about you try removing the "no-wrap" CSS (to fix the print view in Chrome), but then setting the grover viewport to be super wide (the default is only 640px - https://github.com/Studiosity/grover#configuration) so that when it's viewing the page the header doesn't have a linebreak in it?

willc-work: Happy to give this a try, but the scale solution does fix the issue from your screenshot.
Screenshot 2023-07-14 at 19 06 41

patrick-laa: I'm afraid it doesn't fix the issue - your screenshot above is of something different.. I'm talking about printing the results using the 'Print' button on the results page, NOT printing the PDF. As I say, this is completely unaffected by how you scale the PDF, because there is no PDF involved.

willc-work:
@patrick-laa  apologies, you are correct, for some reason i cant use the print from page button on firefox or chrome either locally or on UAT.

app/views/estimates/print.html.slim
```diff
@@ -44,7 +44,7 @@
     = render "evidence"
 
   - if @model.calculated?(:gross_income) || @model.calculated?(:disposable_income) || @model.calculated?(:capital)
-    h2.govuk-heading-m = t("estimates.show.#{@model.level_of_help}_how_calculated")
+    h2.govuk-heading-m.no-wrap = t("estimates.show.#{@model.level_of_help}_how_calculated")
```
patrick-laa, line 47: Because this class is still here, this text still falls off the screen in non-PDF print view on Chrome. I was hoping this wouldn't be needed any more.

app/services/pdf_service.rb
```diff
@@ -7,6 +7,10 @@ class PdfService
       left: "1cm",
       right: "1cm",
     },
+    viewport: {
+      width: 800,
```
patrick-laa, line 11: Has this had any effect, or am I barking up the wrong tree? If it works the way I hoped it works, changing the width from 640 to 800 wouldn't get rid of all breakpoints in the accessibility metadata, but we'd maybe see them shift along a bit (without affecting how the PDF actually looked). Then it would just be a case of finding a width that was enough to remove them entirely.

patrick-laa (approved):

PR #807, opened 2023-07-11
EL-1030: The liveness check checks the database is accessible

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1030)

## What changed and why

This way, for UAT branches, if the database pod gets cycled and the schema gets lost, the Rails pods will stop being 'live' and new pods will be brought online, which will cause migrations to run and the schema to be re-built.

## Checklist

### DIFF
--- a/app/controllers/status_controller.rb
+++ b/app/controllers/status_controller.rb
@@ -1,10 +1,16 @@
 class StatusController < ApplicationController
+  # This is used by the liveness check, used to see if a pod needs replacing
   def index
-    render json: { alive: true }
+    if HealthCheckService.call(check_cfe: false)
+      render json: { alive: true }
+    else
+      render json: { alive: false }, status: :service_unavailable
+    end
   end
 
+  # This is used by the readiness check, used to see if a pod is ready to start receiving traffic
   def health
-    if HealthCheckService.call
+    if HealthCheckService.call(check_cfe: true)
       render json: { healthy: true }
     else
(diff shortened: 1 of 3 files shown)


app/services/health_check_service.rb
```diff
@@ -1,6 +1,6 @@
 class HealthCheckService
-  def self.call
-    database_healthy? && short_term_persistence_healthy? && cfe_healthy?
+  def self.call(check_cfe:)
+    database_healthy? && short_term_persistence_healthy? && (!check_cfe || cfe_healthy?)
```
exonian, line 3: This expression is awfully clever and I had to have a good sip of coffee and a proper think to reason the logic through... It scares me.

app/controllers/status_controller.rb
```diff
@@ -1,10 +1,16 @@
 class StatusController < ApplicationController
+  # This is used by the liveness check, used to see if a pod needs replacing
```
exonian, line 2: I wonder if we could or should rename any of this stuff (stuff being method names and paths) to make it better fit our mental models

exonian: Oh the other PR gets rid of one half anyway so that's fine

app/services/health_check_service.rb
patrick-laa: Ok, I've split it up into two more digestible expressions.

willc-work (approved):

PR #812, opened 2023-07-12
EL-1036 - Use Jira/CircleCI integration

exonian:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1036)

## What changed and why

The jira/notify step has been added to the deploy jobs. This is very much like slack/notify.

## Guidance to review

It has been added to the deploy jobs using `post-steps` where it is invoked in the workflow, as this is the neatest way of passing it parameters that we don't otherwise have available (like display-case environment name, and potentially environment type in future) and avoids having to pass these through the deploy job definition purely to pass into this command.

It has been added to the build job simply as a step, as I don't think it needs any parameters.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -240,7 +240,8 @@ jobs:
           channel: eligibility-alerts-<< parameters.environment >>
           event: fail
           template: basic_fail_1
-      - jira/notify
+      - jira/notify:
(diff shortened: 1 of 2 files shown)


patrick-laa (approved):
Bonus points for alphabetising the orbs

willc-work (approved):

PR #817, opened 2023-07-14
EL-1047: CSP-friendly approach to JS back button

patrick-laa:

## What changed and why

No more inline JS

## Checklist

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -39,14 +39,14 @@ module ApplicationHelper
   end
 
   def back_link(step, check, mimic_browser_back)
-    link = if mimic_browser_back
-             "javascript:history.back()"
-           elsif (previous_step = Steps::Helper.previous_step_for(check.session_data, step))
+    previous_step = Steps::Helper.previous_step_for(check.session_data, step)
+
+    link = if previous_step
              estimate_build_estimate_path(params[:estimate_id], previous_step)
            else
              provider_users_path
            end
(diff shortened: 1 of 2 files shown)


app/helpers/application_helper.rb
```diff
@@ -39,14 +39,14 @@ def integer_field_value(form, field)
...
-           elsif (previous_step = Steps::Helper.previous_step_for(check.session_data, step))
+    previous_step = Steps::Helper.previous_step_for(check.session_data, step)
+
+    link = if previous_step
              estimate_build_estimate_path(params[:estimate_id], previous_step)
            else
              provider_users_path
            end
-    link_to t("generic.back"), link, class: "govuk-back-link"
+    link_to t("generic.back"), link, class: "govuk-back-link", data: { behaviour: ("browser-back" if mimic_browser_back) }
```
MazOneTwoOne, line 49: Question - why do we want the behaviour for the users to be, for them to go back into the previous steps of the estimate flow and not just to the 'Check Your Answers' page?

I'm finding that If I go too far back, the 'Save and Continue' button doesn't let me go to the 'Check Your Answers' page, and expects me to go through the whole flow.

MazOneTwoOne: It was exactly this:

> let me guess: You're running these tests on localhost, you're running rails server with bundle exec rails s rather than bin/dev, and you haven't run yarn build recently? Because it sounds to me like the new javascript simply isn't running when you test this out.

I thought it was odd, when testing it out.

MazOneTwoOne (approved):

PR #824, opened 2023-07-17
EL-1022: Redirect to primary host system

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1022)

## What changed and why

Make it so that the start page and the new estimate path both redirect to the primary host if one is set, so that existing checks can be completed on an old host but new ones always end up on the new one.

## Checklist


willc-work (approved):
looks good, should we have a ticket to remove this at some future point as it will not be required once the old URL is turned off?

PR #826, opened 2023-07-17
EL-1037: Add new CW1 and 2  mapping for MTR phase 1

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1037)

## What changed and why

- Mechanism to choose different PDF/mapping based on presence of feature flag.
- Add the new CW1&2 pdf and mapping file

## Checklist


spec/forms/controlled_work_document_selection_spec.rb
```diff
@@ -65,4 +65,14 @@
...
     expect(page).to have_content "What level of help does your client need?"
   end
+
+  context "when MTR phase 1 is in effect" do
+    before { travel_to Date.new(2023, 8, 4) }
+
+    it "downloads an updated PDF" do
+      choose "CW1&2 - mental health"
+      click_on "Download the pre-populated form"
+      expect(page.response_headers["Content-Type"]).to eq("application/pdf")
```
willc-work, line 75: i might be missing something here, how do we know this is the new file?

willc-work (approved):
one query, not a blocker

spec/forms/controlled_work_document_selection_spec.rb
patrick-laa: Good point. Mostly this spec was testing "when we load an MTR phase 1 PDF, nothing explodes" rather than "when we're after 3 August, we load an MTR phase 1 PDF". I've added a new expectation to specify that the correct mapping file gets retrieved.

PR #834, opened 2023-07-19
EL-1051 show production specific robots.txt file

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1051)

Robots.txt has its own controller and displays a robots.txt file depending on feature flag
noindex is removed when feature flag is enabled
canonical link added to each page when the primary host exists

## Checklist


app/views/layouts/application.html.slim
```diff
@@ -12,7 +12,17 @@ html dir="ltr" lang="en-GB" class="govuk-template"
...
     = tag.meta property: "og:image", content: asset_path("images/govuk-opengraph-image.png")
     = tag.meta name: "theme-color", content: "#0b0c0c"
-    = tag.meta name: "robots", content: "noindex"
+    ruby:
+      index_prod = FeatureFlags.enabled?(:index_production, without_session_data: true) ? "" : "noindex"
+    = tag.meta name: "robots", content: index_prod.to_s
+
+    ruby:
+      canoncical_link = if ENV["PRIMARY_HOST"].present?
+                          "https://check-your-client-qualifies-for-legal-aid.service.gov.uk#{url_for(only_path: true, protocol: 'https')}"
```
patrick-laa, line 21: Rather than hardcode the URL here, would it make sense to use `"https://#{ENV["PRIMARY_HOST"]}"`?

app/views/layouts/application.html.slim
```diff
@@ -12,7 +12,17 @@ html dir="ltr" lang="en-GB" class="govuk-template"
...
+                        end
+      use_canionical = ENV["PRIMARY_HOST"].present? ? "canonical" : nil
```
patrick-laa, line 23: Rather than always create a `link` element, and half the time have it say `<link rel="" href="" />`, would it make sense to put a single `- if  ENV["PRIMARY_HOST"].present?` and indent the logic to construct the link element inside that?

app/views/layouts/application.html.slim
```diff
@@ -12,7 +12,17 @@ html dir="ltr" lang="en-GB" class="govuk-template"
...
+
+    link[rel="#{use_canionical}" href='<%= #{canoncical_link} %>']
```
patrick-laa, line 25: That `<%=` is ERB syntax that won't be correctly rendered here. You'll end up with HTML that looks like:

```html
 <link rel="canonical" href="<%= https://check-your-client-qualifies-for-legal-aid.service.gov.uk/ %">
 ```
 
 And that's not a valid href value.

File: app/views/layouts/application.html.slim (hunk 1 above)
Comment (patrick-laa) [reply app/views/layouts/application.html.slim:21]: Also the `protocol: https` bit is redundant, I think, as the `url_for` isn't building the bit with the protocol in it.

File: app/views/layouts/application.html.slim [hunk 4]
```diff
@@ -12,7 +12,12 @@ html dir="ltr" lang="en-GB" class="govuk-template"
     = tag.meta name: "viewport", content: "width=device-width, initial-scale=1"
     = tag.meta property: "og:image", content: asset_path("images/govuk-opengraph-image.png")
     = tag.meta name: "theme-color", content: "#0b0c0c"
-    = tag.meta name: "robots", content: "noindex"
+    ruby:
+      index_prod = FeatureFlags.enabled?(:index_production, without_session_data: true) ? "" : "noindex"
```
patrick-laa, line 16: If we do want Google to index the site, we'll have a directive that looks like: `<meta name="robots" content="">`. I'm not sure it's [strictly valid](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag#directives) to have a robots directive with blank content. To be on the safe side, maybe we should have it so that the "content" value if we do want indexing is `"all"` rather than `""`.

patrick-laa (approved):
One thought, but looks good.

PR #854, opened 2023-07-27
EL-1058: Replace irrelevant CW zeroes with blanks

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1058)

## What changed and why

* Refactor ControlledWorkDocumentContent model to split methods across multiple modules. This is a cosmetic change, as all the methods end up on the same object, and in time it might be nice to change that, but this split makes it easier to read and follow between all the different methods.
* Adjust some of the CW form disposable income fields so that if they're not relevant to the check, they are left blank rather than being filled with `0`.

## Checklist


app/models/concerns/outgoings_summarisable.rb
```diff
@@ -0,0 +1,93 @@
...
+  def dependants_allowance_over_16
+    from_cfe_payload("result_summary.disposable_income.dependant_allowance_over_16") if client_income_relevant? && adult_dependants
+  end
+
+  def client_tax_and_national_insurance
+    tax_and_national_insurance if client_income_relevant? && employed?
+  end
+
+  def partner_tax_and_national_insurance
+    tax_and_national_insurance "partner_" if partner_income_relevant? && partner_employed?
```
willc-work, line 35: might be good to add parentheses around the paramater here for consistency

app/models/concerns/outgoings_summarisable.rb
```diff
@@ -0,0 +1,93 @@
...
+private
+
+  def net_housing_costs(summary_section_prefix = "")
+    session_data.dig("api_response", "result_summary", "#{summary_section_prefix}disposable_income", "net_housing_costs")
+  end
+
+  def tax_and_national_insurance(summary_section_prefix = "")
+    tax = session_data.dig("api_response", "result_summary", "#{summary_section_prefix}disposable_income", "employment_income", "tax") || 0
+    national_insurance = session_data.dig("api_response", "result_summary", "#{summary_section_prefix}disposable_income", "employment_income", "national_insurance") || 0
+    -1 * (tax + national_insurance)
```
willc-work, line 86: I think future devs would appreciate a comment here explaining that the -1 multiplier is to present these as negative figures as they are shown as deductions on the CW forms.

PR #863, opened 2023-08-02
EL-1069: Banner, for public beta URLs

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1069)

## What changed and why

Banner notification to be added to old private beta domain. So that users understand that the link is no longer active.

## Checklist


app/helpers/application_helper.rb
```diff
@@ -54,4 +54,8 @@ def enable_google_analytics(cookies)
       cookies[CookiesController::COOKIE_CHOICE_NAME] == "accepted" &&
       !cookies[CookiesController::NO_ANALYTICS_MODE]
   end
+
+  def beta_pilot_url?
+    request.base_url.include?("cloud-platform")
```
patrick-laa, line 59: This isn't an ideal test, because it's almost impossible to replicate in development and test environments. It's also reliant on something that happens to be the case (the environments we want to show this on _happen to_ have "cloud-platform" in the URL and the environments we don't _happen_ not to), but it would be better to rely on something that's a bit more definite. I suggest that we want to show this at times where the following applies:

- We have defined a "primary" host for this environment
- This environment is _not_ the primary host.

If you look in ApplicationController#redirect_to_primary_host you'll see a similar check defined. I think we could copy that here, so this method would become `def using_non_primary_url?` or something like that.

app/views/cookies/show.html.slim
```diff
@@ -1,107 +1,111 @@
...
-      td.govuk-table__cell
-        = t(".one_year")
-    tr.govuk-table__row
-      th.govuk-table__header[scope="row"]
-        = LaaEstimateFinancialEligibilityForLegalAid::SESSION_COOKIE_NAME
-      td.govuk-table__cell
-        = t(".stores_user_provided_data")
-      td.govuk-table__cell
-        = t(".until_browser_closed")
+.govuk-grid-row
```
patrick-laa, line 1: You don't need this outermost div, as a div with this class is already defined in `application.html.slim` on line 82

app/views/cookies/show.html.slim
```diff
@@ -1,107 +1,111 @@
...
+  .govuk-grid-column-full
+    - content_for(:page_title)
```
patrick-laa, line 3: For consistency, this `content_for` block and its contents should probably sit outside the grid-column class, like in e.g. privacies/show.html.slim

app/views/layouts/application.html.slim
```diff
@@ -48,6 +48,36 @@ html dir="ltr" lang="en-GB" class="govuk-template"
         div role="navigation"
           = yield :back
 
+      - if beta_pilot_url? && FeatureFlags.enabled?(:public_beta, without_session_data: true)
+       - if controller_name == "estimates"
```
patrick-laa, line 52: This is another not very ideal check. It just happens to be the case that all `estimates` pages want full width banners and all other pages want 2-thirds width, but checking the controller name isn't very easy to read, there are lots of structural refactors we could make where this will break, and when a different page wants a different width we won't thank ourselves for not having made it easy for ourselves. Also, this is hugely repetitive - we repeat 12 lines of code verbatim thanks to this `if` statement.

Here's what I suggest. As per the guidance, this component should always appear directly above the h1, right? So let's insert the banner into `shared/_heading.html.slim`. On 2-thirds-width pages that will already be rendered inside `govuk-grid-column-two-thirds` divs, so it will always be the width we want.

The only gotcha is that we don't always use the `shared/heading` component yet - most notably on single-question pages like `provider_users/show`, where we build the h1 element from the legend of the field component. In those cases, I think the solution is now to render that component in all cases (it's just that on screens like that we _wouldn't_ pass in `header_text` as a param). That would also tidy up the way that in a lot of cases we're manually rendering `shared/household_tag` and `shared/partner_tag`, while in some cases we're relying on `shared/heading` to do it for us. If we call `shared/heading` on every screen, we can _stop_ manually rendering those tags. So then `heading` becomes more useful and our screens all become more consistent.

config/locales/en.yml
```diff
@@ -404,6 +404,8 @@ en:
     beta: Beta
     beta_message_html: This is a new service - your <a href="%{user_survey_link}" target="_blank">feedback</a> will help us improve it
     beta_message_short: This is a new service
+    beta_message_pilot_heading: This page is no longer active as the pilot phase is complete.
+    beta_message_pilot_body_html: Thank you for taking part. </br> <a href="https://check-your-client-qualifies-for-legal-aid.service.gov.uk" target="_blank">Check if your client qualifies for legal aid</a> is now live and available to all users.
```
patrick-laa, line 408: Also, rather than hard-code this URL here, I'd use `ENV["PRIMARY_HOST"]` passed in as a variable

app/views/cookies/show.html.slim
MazOneTwoOne: By doing this I no longer need the outermost div - `.govuk-grid-row`.

Before this, the new banner was sitting on top of the existing text on the 'cookies' page.

config/locales/en.yml
MazOneTwoOne: What it the syntax? I've arrived at this:

```
public_beta_link: <%= ENV["PRIMARY_HOST"] %>
beta_message_pilot_body_html: Thank you for taking part. </br> <a href="%{public_beta_link}" target="_blank">Check if your client qualifies for legal aid</a> is now live and available to all users.

```

app/helpers/application_helper.rb
```diff
@@ -55,6 +55,10 @@ def enable_google_analytics(cookies)
       !cookies[CookiesController::NO_ANALYTICS_MODE]
   end
 
+  def using_non_primary_url?
+    ENV["PRIMARY_HOST"].blank? || ENV["PRIMARY_HOST"] == request.host
```
patrick-laa, line 59: This looks backwards. Reading your code here in plain English, it seems to be saying:

"We are using a non-primary URL if either there is no primary host defined, or if there is a primary host and we are currently on it"

I think it should say:

"We are using a non-primary URL if there _is_ a primary host defined and we are _not_ on it."

patrick-laa (commented):
All looking very tidy, just one bit of `if` logic that needs looking at.

app/helpers/application_helper.rb
MazOneTwoOne: I think I've updated that now - I was getting confused when I lifted from the ApplicationController .

patrick-laa: It's still not quite there - because we only want to show the banner if PRIMARY_HOST _is_ defined. At the moment, if it's not defined, this check is effectively `nil != request.host`. Since request.host is always not nil, this will mean that if PRIMARY_HOST is not defined, this method will always return true.

MazOneTwoOne: like this?

```
  def using_non_primary_url?
    ENV["PRIMARY_HOST"].nil? && ENV["PRIMARY_HOST"] != request.host
  end
```

app/helpers/application_helper.rb
```diff
@@ -55,6 +55,10 @@ def enable_google_analytics(cookies)
       !cookies[CookiesController::NO_ANALYTICS_MODE]
   end
 
+  def using_non_primary_url?
+    ENV["PRIMARY_HOST"].blank? && ENV["PRIMARY_HOST"] != request.host
```
patrick-laa, line 59: This says "if PRIMARY_HOST is blank and is different to the host we're currently on, then we're using a non-primary URL." I don't think that's right.

MazOneTwoOne: This checks if there is a primary host defined (returns out if not), and then, if we are not on it. (_I think 😬_). 

```
  def using_non_primary_url?
    return unless ENV["PRIMARY_HOST"].nil?

    ENV["PRIMARY_HOST"] != request.host
  end
```

app/helpers/application_helper.rb
```diff
@@ -55,6 +55,12 @@ def enable_google_analytics(cookies)
       !cookies[CookiesController::NO_ANALYTICS_MODE]
   end
 
+  def using_non_primary_url?
+    return unless ENV["PRIMARY_HOST"].nil?
```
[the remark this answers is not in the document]
MazOneTwoOne: ok...(this is painful).. so I think this will 

- return true, if the url you are on is _not_ the PRIMARY_HOST, when that value is defined. 
- return false, if the url you are on _is_  the PRIMARY_HOST, when that value is defined. 

```
  def using_non_primary_url?
    request.host != ENV["PRIMARY_HOST"] if ENV["PRIMARY_HOST"].present?
  end
```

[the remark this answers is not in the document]
patrick-laa: So we're _nearly_ there.

`x if y` is the same as:

```
if y
  x
end
```

This is an if statement without an `else` clause, so it doesn't specify what to return if `y` is not true. As such, if `y` is true, that expression will return the value of `x`, but if `y` is not true the expression will return `nil`.

So in this case if PRIMARY_HOST is not set, this method will return `nil`, when ideally we'd like it to return `false`. This is effectively the same thing (because nil is what's called "falsey", so behaves quite a lot _like_ false), but to really tighten this up, I think we're better off ensuring it always returns either `true` or `false`.

I think we're best off using an "and" statement. i.e. we want this method to say "(we know we are using a non-primary URL when) PRIMARY_HOST is present and the request host is different to the PRIMARY_HOST".

To achieve that, I'd recommend:

```ruby
def using_non_primary_url?
   ENV["PRIMARY_HOST"].present? && request.host != ENV["PRIMARY_HOST"]
end
```

PR #865, opened 2023-08-04
EL-993: CCQ Service Unavailable Page

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-993)

## What changed and why

If the CCQ service is unavailable (e.g. down for maintenance) or if CFE is not available the cloud platform error page is shown. We want control of the error message shown.

## Guidance to review

My only comment on this, is why can't we class this 'service unavailable' page with a [503 error code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503), instead of replacing the 500 page? 

We could make the 'service unavailable' page more bespoke in the future, if needed and keep the 500 page as a generic page? 

## Checklist


app/controllers/application_controller.rb
```diff
@@ -58,4 +58,10 @@ def redirect_to_primary_host
 
     redirect_to [request.protocol, ENV["PRIMARY_HOST"], request.fullpath].join, allow_other_host: true
   end
+
+  def check_maintenance_mode
+    if FeatureFlags.enabled?(:maintenance_mode, without_session_data: true)
+      redirect_to "/500"
```
patrick-laa, line 64: If the site is in maintenance mode and I visit `check-client-blah...gov.uk/privacy`, this will show me the error page but it will do it by sending me to `check-client-blah...gov.uk/500`. Which is a bit weird, because this isn't a "500" situation - 500 means there's been an error. And it's also annoying to be redirected, because if it's just a temporary blip I'd like to be able to reload the page to see if it's fixed, and reloading won't change the URL back to `.../privacy`.

So I'd recommend that instead of a redirect, we do a render, i.e. (I think) `render file: "public/500.html", status: : service_unavailable, layout: false`

helm_deploy/laa-estimate-eligibility/values/production.yaml
```diff
@@ -61,6 +61,7 @@ featureFlags:
   overrides: NOT_ENABLED
   publicBeta: NOT_ENABLED
   indexProduction: NOT_ENABLED
+  maintenance_mode: NOT_ENABLED
```
patrick-laa, line 64: Ok so this is a good guess, but for consistency, you want to use camelCase not snake_case here. i.e. `maintenanceMode`.

Now, how does that get translated into an appropriate environment variable? Well, for that you need to _also_ add a couple of lines to `helm_deploy/laa-estimate-eligibility/templates/_envs.tpl`, round about line 73

spec/controllers/application_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe ApplicationController, type: :controller do
```
patrick-laa, line 3: As this is a user-facing feature, it would be better to have a user-centred spec. What I mean by this is that in general, it's good for our tests to describe what happens in terms of what the users experience, because that's a better way of "proving" that the site does what we expect. So we want our test to basically say:

> when the site is in maintenance mode, if I try to visit the start page I should see a message saying "Sorry there is a problem with the service"

The closest we can get to that in RSpec would be a feature spec (i.e. a new file `spec/features/maintenance_mode_spec.rb`, where we ideally switch on the feature flag in the same way we do for other tests (i.e. setting a tag like `:maintenance_mode_flag` in rails_helper.rb) that has a spec that says things like `visit root_path` and `expect(page).to have_text "Sorry there is a problem with the service")`

The spec you have here is fine, but it just checks that if maintenance mode is on, calling the `check_maintenance_mode` method will call the `redirect_to` method, and there could be some bugs that mean that, even though that is true, the end result of showing the correct error message to the user is not achieved.

app/controllers/application_controller.rb
MazOneTwoOne: Had to go with this:

`render file: "public/500_Service_Unavailable.html", status: :service_unavailable, layout: "layouts/500"`

- If I set `layout: false`, we get none of the CSS stylings. 
- If I set `layout: true` , we get two headers and footers. 
- Therefore created another 500 page and added in a specific layout for it.

MazOneTwoOne: Ignore this comment - we (as per slack message) updated CSP settings.

patrick-laa (approved):

patrick-laa:
Looks good. For UAT we can use our feature flag override functionality to turn on maintenance mode (although thinking about it, once it's switched on, we won't have access to the feature flag override screens to turn it off again! That's not a problem, though, I don't think - we can us the command line to remove the override if we need to).

PR #874, opened 2023-08-08
EL-1070: Move change logs to new Updates page

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1070)

## What changed and why

Removed custom JS/CSS to toggle change log on home page
Added new updates page that lists all change logs

## Checklist


app/views/change_logs/_log_item.html.slim
```diff
@@ -1,3 +1,4 @@
-.change-log-item data-change-log-role="item"
-  h2.govuk-heading-s = Date.parse(change_log[:change_on]).strftime("%-d %B %Y")
+.change-log-item class="govuk-!-margin-bottom-7"
```
MazOneTwoOne, line 1: Did we not get rid of this `change-log-item` class?

app/views/change_logs/history_texts/_dependant_allowance.html.slim
```diff
@@ -1,3 +1,6 @@
+h3.govuk-heading-s
+  = t("change_logs.dependant_allowance.history.header")
+  .govuk-tag class="govuk-!-margin-left-3" = t("change_logs.tags.mtr")
```
MazOneTwoOne, line 3: Adding in the tags, now seem so simple and obvious. 

I thought we were going to something more involved, like building a helper which would take arguments [title] and [tag] and from that we would build this bit.

app/views/start/index.html.slim
```diff
@@ -73,10 +73,5 @@
     hr
```
MazOneTwoOne, line 73: Do we still need the `- if ChangeLogs.occurred.any?` as I assume we will always show this bit of text?

MazOneTwoOne (approved):
Just some small comments but happy to approve.

One overall minor comment (which I might be overthinking), should we have a think about how we are using 'change-log' going forward? If we are introducing an 'Updates' page should we refer to things as the 'update-logs' and 'update-item'?

app/views/start/index.html.slim
patrick-laa: That's a fair point - that was put in when we first built this logic and we didn't have any change logs, but we'll never now _not_ have any. I'll get rid of it.

app/views/change_logs/_log_item.html.slim
patrick-laa: Good spot! Yes, that's now entirely redundant. I'll remove it.

app/views/change_logs/history_texts/_dependant_allowance.html.slim
patrick-laa: I had thought about building something more structured, but since we'll be replacing all these partials with something dynamic and database-y in the near future, I thought better to leave it simple for now, since it's all pretty manageable as-is.

patrick-laa:
@MazOneTwoOne re your point about naming, I'm inclined to keep calling change logs "change logs" because soon we'll have _two_ sorts of content on the Updates page: change logs and bug notifications. So I think it's fair to say a change log is a type of update item, but it's not the _only_ type of update item.

MazOneTwoOne:
> @MazOneTwoOne re your point about naming, I'm inclined to keep calling change logs "change logs" because soon we'll have _two_ sorts of content on the Updates page: change logs and bug notifications. So I think it's fair to say a change log is a type of update item, but it's not the _only_ type of update item.

When you put it like that, it makes sense. The 'Updates' page will have 'change-log' items & 'bug-log' items.

app/views/start/index.html.slim
patrick-laa: Oh, I'm wrong: a bunch of the tests "time travel" to points before we had any change logs, and then break when the start page is viewed.

MazOneTwoOne: Oh ok! 
Do we then,
- leave a comment in for future devs?
or
- rewrite the tests?

patrick-laa: I'm inclined to leave it for now, on the grounds that we'll have to rewrite this code anyway when Change Logs become database objects, and when they are it'll be much more obvious that we shouldn't assume that they will exist.

PR #876, opened 2023-08-08
EL-1060: Site-wide basic auth

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1060)

## What changed and why

Use basic authentication everywhere if it's switched on in config
Except robots and status controllers
Use this _instead_ of feature flag override basic auth
Leave higher-security Blazer basic auth untouched.

## Checklist

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,7 +1,7 @@
 class ApplicationController < ActionController::Base
   BROWSER_ID_COOKIE = :browser_id
   default_form_builder GOVUKDesignSystemFormBuilder::FormBuilder
-  before_action :force_setting_of_session_cookie, :check_maintenance_mode
+  before_action :force_setting_of_session_cookie, :authenticate, :check_maintenance_mode
 
   class MissingSessionError < StandardError; end
 
@@ -63,5 +63,16 @@ private
     if FeatureFlags.enabled?(:maintenance_mode, without_session_data: true)
       render file: "public/500.html", layout: false
     end
+  end
+
+  def authenticate
+    return unless FeatureFlags.enabled?(:basic_authentication, without_session_data: true)
+
+    return true if authenticate_with_http_basic do |username, password|
+      username == "ccq" && password == ENV["BASIC_AUTH_PASSWORD"]
+    end
+
+    request_http_basic_authentication
+    false
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -68,11 +68,8 @@ private
   def authenticate
     return unless FeatureFlags.enabled?(:basic_authentication, without_session_data: true)
 
-    return true if authenticate_with_http_basic do |username, password|
+    authenticate_or_request_with_http_basic("Check if your client qualifies for legal aid") do |username, password|
       username == "ccq" && password == ENV["BASIC_AUTH_PASSWORD"]
     end
-
-    request_http_basic_authentication
-    false
   end
 end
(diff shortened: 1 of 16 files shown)


app/controllers/application_controller.rb
```diff
@@ -58,4 +58,15 @@ def redirect_to_primary_host
 
     redirect_to [request.protocol, ENV["PRIMARY_HOST"], request.fullpath].join, allow_other_host: true
   end
+
+  def authenticate
+    return unless FeatureFlags.enabled?(:basic_authentication, without_session_data: true)
+
+    return true if authenticate_with_http_basic do |username, password|
```
MazOneTwoOne, line 65: So reading [this ](https://stackoverflow.com/questions/5101020/in-ruby-on-rails-what-does-authenticate-with-http-basic-do) we can call `authenticate_or_request_with_http_basic` here without the need for `request_http_basic_authentication` - no idea if it'll work but thought it was worth a mention.

app/controllers/robots_controller.rb
```diff
@@ -1,4 +1,6 @@
 class RobotsController < ApplicationController
+  skip_before_action :authenticate, only: :index
```
MazOneTwoOne, line 2: I hadn't realised how simple this was to implement. I thought we'd have to control it on the ApplicationController. 

question - why do we need to specify to only apply it to `index`, when it's the only thing there?

helm_deploy/laa-estimate-eligibility/templates/_envs.tpl
```diff
@@ -110,11 +112,11 @@ env:
       secretKeyRef:
         name: kube-secrets
         key: blazer-database-password
-  - name: FEATURE_FLAGS_PASSWORD
+  - name: BASIC_AUTH_PASSWORD
     valueFrom:
       secretKeyRef:
         name: kube-secrets
-        key: feature-flags-password
+        key: basic-auth-password
```
MazOneTwoOne, line 119: question - this stuff is going "over my head". Am I to understand that here we are updating where we now point to where our password values are kept???

spec/features/basic_auth_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+RSpec.describe "Basic auth" do
+  around do |example|
+    ENV["BASIC_AUTH_PASSWORD"] = "password"
```
MazOneTwoOne, line 5: question - do we need to specify that we have a our username in these scenarios?

MazOneTwoOne (approved):
Approving with comments to check my understanding. 

I couldn't actually sign-in once I enabled the feature flag but I'm sure, I was doing something wrong. 

p.s - love the password name :D

helm_deploy/laa-estimate-eligibility/templates/_envs.tpl
patrick-laa: So there is a box of secrets, and it's called `kube-secrets`, and we used to pull a secret called `feature-flags-password` out of it and use it to set an environment variable called `FEATURE_FLAGS_PASSWORD`. Now we don't need that any more, so we've deleted it. We now pull a secret called `basic-auth-password` out of the box of secrets, and use it to populate an environment variable called `BASIC_AUTH_PASSWORD`.

app/controllers/robots_controller.rb
patrick-laa: That's just a failsafe. The default should be for all pages to be authenticated, and so if we ever add any new actions to RobotsController we want to make sure they're authenticated unless we specifically add code to say that they aren't.

app/controllers/application_controller.rb
patrick-laa: Good spot! Yes, that looks like it rolls both things into one, which is much neater.

PR #883, opened 2023-08-10
EL-1093: Avoid repeatedly setting up a suggestions component

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1093)

## What changed and why

The suggestions components were getting re-initialized multiple times, leading to multiple suggestions boxes. Now they are only initialized when they are first added to the DOM.

## Checklist


app/javascript/add-another.js
```diff
@@ -39,12 +39,14 @@ const addAnother = (addAnotherContainer) => {
...
   addAnotherContainer.querySelector('[data-add-another-role="sectionList"]').append(newSection);
-  setUpSection(newSection, counter)
+  setUpSection(newSection, counter, { setUpSuggestions: true })
 };
 
-const setUpSection = (section, counter) => {
+const setUpSection = (section, counter, options) => {
   setUpRemoveButton(section);
-  setUpSuggestions(section);
+  if (options && options.setUpSuggestions) {
```
MazOneTwoOne, line 47: question - not sure I'm following on this. 

- So now, every time we use the 'add-another' pattern we say if it includes a 'suggestions list'. 
- If it does we call the `setUpSuggestions` function and pass the `section` as an argument. 

I don't quite follow how this prevents more that one 'suggestions list' showing up?

MazOneTwoOne (approved):
Happy to approve. 

Have a clarifying question to check my understanding.

app/javascript/add-another.js
patrick-laa: In application.js, on page load, we say "for every `govuk-input` field, initialize the javascript that gets the suggestions list ready". But the problem with that is that when we add another, on the benefits screen we _create new_ govuk-input fields, that haven't been initialized.

So previously, every time we made a change to an add another screen, we re-ran the logic that initializes the javascript for suggestions lists for every `govuk-input` on the screen. But it turns out, initializing the javascript multiple times causes multiple suggestions lists to be readied.

So _now_ we still initialise JS for all `govuk-input`s that are on the page when it loads in application.js, and we've change the logic here so that it only runs `govuk-input` initialization on sections that have just been added to the page (by clicking the 'Add another' button), so for each `govuk-input` the javascript only gets initialized once.

PR #885, opened 2023-08-10
EL-1054 part 2: Pull 90-day-completion rate

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1054)

## What changed and why

After some discussion, we would like to display in Geckoboard the completion rate not for all time but for the last 90 days. 

## Checklist


app/services/metrics/for_key_metric_dashboard_service.rb
```diff
@@ -134,7 +134,9 @@ def checks_completed(range = nil)
       relevant_events(range).where(event_type: "page_view", page: "view_results").count("DISTINCT assessment_code")
     end
 
-    def completion_rate(range = nil)
+    # The 'all time' completion rate is actually the last-90-day completion rate, as for a % rate, ancient historical
+    # data is less relevant.
+    def completion_rate(range = 90.days.ago.beginning_of_day..Date.current)
```
MazOneTwoOne, line 139: question - do we want to return `nil` if this validates as true `return if started.zero? ` or do we want to return `0`?

MazOneTwoOne (approved):
Approving with a question.

app/services/metrics/for_key_metric_dashboard_service.rb
patrick-laa: I think it should be nil. If no checks have been started, then the completion _rate_ isn't zero. I'd argue there is no completion rate - in the absence of any checks, it doesn't make sense as a concept. Zero_ means "none of the checks that were started were completed", while nil means "this metric doesn't apply".

MazOneTwoOne: that makes sense.

PR #897, opened 2023-08-16
EL-1086: Rework results page for eligible controlled checks

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1086)

## What changed and why

Make the CW form CTA more prominent when present.

## Checklist

### DIFF
--- a/app/helpers/results_helper.rb
+++ b/app/helpers/results_helper.rb
@@ -48,7 +48,7 @@ module ResultsHelper
     end
   end
 
-  def pdf_friendly_p_element(text, is_pdf, additional_class)
+  def pdf_friendly_p_element(text, is_pdf, additional_class = nil)
     if is_pdf
       tag.ul(class: "govuk-list #{additional_class}", style: "list-style-type: none; margin: 0; padding-bottom: 10px; font-variant-ligatures: none;") do
         tag.li(text)
--- a/app/views/estimates/_controlled_work_next_steps.html.slim
+++ b/app/views/estimates/_controlled_work_next_steps.html.slim
@@ -1,13 +1,17 @@
-- if @check.partner
-  = pdf_friendly_paragraphs(t("estimates.show.controlled_cw_next_steps.intro_paragraphs_partner"), @is_pdf)
-- else
-  = pdf_friendly_paragraphs(t("estimates.show.controlled_cw_next_steps.intro_paragraphs"), @is_pdf)
+= pdf_friendly_p_element(t("estimates.show.controlled_cw_next_steps.intro_paragraph#{'_partner' if @check.partner}"), @is_pdf)
+= yield
+= pdf_friendly_p_element(t("estimates.show.controlled_cw_next_steps.second_paragraph"), @is_pdf)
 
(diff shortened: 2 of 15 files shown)


app/views/estimates/show.html.slim
```diff
@@ -53,12 +46,16 @@
           = render "capital_table", caption_size: "m"
 
   .govuk-button-group
-    - if @model.level_of_help == "controlled" && @model.decision == "eligible"
-      = link_to t(".continue_to_cw"), new_estimate_controlled_work_document_selection_path(params[:id]), class: "govuk-button", data: { module: "govuk-button" }
-    - else
+    - if @model.level_of_help == "certificated" || @model.decision != "eligible"
```
MazOneTwoOne, line 49: question - so EL-1097 is trying to rename & re-order the CTA's on this page. Would it be best to merge that one first then come back to this?

app/helpers/results_helper.rb
```diff
@@ -48,7 +48,7 @@ def pdf_friendly_h3(text, is_pdf)
     end
   end
 
-  def pdf_friendly_p_element(text, is_pdf, additional_class)
+  def pdf_friendly_p_element(text, is_pdf, additional_class = nil)
```
MazOneTwoOne, line 51: question - this makes sense. Would we want to get rid of the `nil` in all the place we are calling `pdf_friendly_p_element` or shoudl we leave them there?

app/views/estimates/_controlled_work_next_steps.html.slim
```diff
@@ -1,14 +1,18 @@
-- if @check.partner
-  = pdf_friendly_paragraphs(t("estimates.show.controlled_cw_next_steps.intro_paragraphs_partner"), @is_pdf)
-- else
-  = pdf_friendly_paragraphs(t("estimates.show.controlled_cw_next_steps.intro_paragraphs"), @is_pdf)
+= pdf_friendly_p_element(t("estimates.show.controlled_cw_next_steps.intro_paragraph#{'_partner' if @check.partner}"), @is_pdf)
+= yield
```
MazOneTwoOne, line 2: question - so do we `yield` back to `estimates/show.html.slim` & `estimates/print.html.slim` just to show the 'Evidence you might need panel'??

app/views/estimates/show.html.slim
patrick-laa: It could work either way - we'll either have to merge that first and rebase this, or vice versa (and it might end up being vice versa given how long it is before Will's back). But it's only a couple of lines of code so it should be fine either way.

app/helpers/results_helper.rb
patrick-laa: Ah, yep I hadn't spotted that we were calling this with a bunch of explicit `nil`s. That's not ideal, so I've added a tidy-up commit to stop doing that.

app/views/estimates/_controlled_work_next_steps.html.slim
patrick-laa: Exactly - the show and print screens handle the "Evidence you might need" panel pretty differently. Rather than try to resolve it in here, we `yield` back to them at this point so they can put whatever they want in.

MazOneTwoOne (approved):

PR #899, opened 2023-08-16
EL-1097 update content and order of buttons on result page

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1097)

## What changed and why

Buttons moved and priorities changed, labels changed

## Checklist


app/helpers/results_helper.rb
```diff
@@ -36,16 +36,8 @@ def pdf_friendly_h2(text, size, is_pdf, additional_class = nil)
     end
   end
 
-  def pdf_friendly_h3(text, is_pdf)
```
willc-work, line 39: @MazOneTwoOne Could you have a look at this. I was getting failing code coverage for if part of this method. From what I can see we use it only in app/views/estimates/_controlled_work_next_steps.html.slim

But I cannot work out what combination of attributes a case needs and what action has to be undertaken to generate a PDF that uses this styling. Hence I have removed it to see if anything breaks and it doesnt appear to have had any impact. I would welcome your thoughts on it though.

patrick-laa: @willc-work this gets used when generating the PDF of the results page for a controlled work check. The reason you got a code coverage error is that you've changed the `downloads_spec` tests so that now you're not generating a download for a controlled work check.

And the reason you've ended up doing that is it looks like you've accidentally removed the 'Download' option on the results screen for eligible controlled checks.

willc-work: thanks @patrick-laa

MazOneTwoOne (approved):

app/views/estimates/show.html.slim
```diff
@@ -55,10 +55,13 @@
   .govuk-button-group
     - if @model.level_of_help == "controlled" && @model.decision == "eligible"
       = link_to t(".continue_to_cw"), new_estimate_controlled_work_document_selection_path(params[:id]), class: "govuk-button", data: { module: "govuk-button" }
-    - else
-      = link_to t(".start_another_check"), new_estimate_path, class: "govuk-button", data: { module: "govuk-button" }
+    - elsif @model.level_of_help == "certificated" || @model.level_of_help == "controlled" && @model.decision == "ineligible"
```
patrick-laa, line 58: This could just be `- else`, as we want this branch to run in every situation other than the one mentioned in the `if` clause.

PR #900, opened 2023-08-16
EL-1018: Update CW Form template used in pre-population

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1018)

## What changed and why

Updating CW Form PDFs that are populated by CCQ.

## Guidance to review

Removed some tests, which talk about, when the 'MTR Phase 1' feature flag is enabled.  Wasn't sure if I should also remove the mappings & PDF templates for 'MTR Phase 1' (and associated tests)??

## Checklist


patrick-laa (commented):
We should remove the things we're no longer using - that means removing PDFs that are no longer ever server, removing mappings that are no longer ever read, and removing redundant entries in the ` TEMPLATES ` hash.

app/lib/controlled_work_mappings/cw2_header.yml
```diff
@@ -0,0 +1,261 @@
+# This file defines the mappings used to prepopulate a CW2 IMM PDF form with the answers given by a user to CCQ's questions
```
patrick-laa, line 1: This alarms me - GitHub hasn't picked up that this file is just `cw2_mtr_phase_2.yml` renamed, although it has done for every other new yaml file. Are you confident that this is identical to the file you've deleted?

app/services/controlled_work_document_value_mapping_service.rb
```diff
@@ -25,7 +25,7 @@ def call(session_data, mappings)
     def convert(mapping, content)
       case mapping[:type]
       when "checkbox"
```
patrick-laa, line 27: so yeah, as per above, I'd change this to `always_checked_checkbox` or something now that its behaviour has changed.

patrick-laa (commented):
I think we're basically there, just a bit nervous with all these yaml files flying around.

app/lib/controlled_work_mappings/cw2_header.yml
MazOneTwoOne: I can explain - it was my 'neatness' itch, that I had to scratch. 

This **was** copied from the previous `cw2_mtr_phase_1.yml` file. I just indented mappings to the left, to match all the other mappings.

PR #904, opened 2023-08-17
EL-1099: Standardise numbering of "add another" sections

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1099)

## What changed and why

All 'add another' sections have labels starting at 1. Remove JS logic to allow starting at a hidden zero.

## Checklist

### DIFF
--- a/app/javascript/add-another.js
+++ b/app/javascript/add-another.js
@@ -85,7 +85,7 @@ const setUpSections = (sectionList) => {
 const setNumbering = (section, counter) => {
   const counterElement = section.querySelector('[data-add-another-role="counter"]')
   if (counterElement) {
-    counterElement.innerHTML = counterElement.dataset.addAnotherNumberFrom === "zero" ? counter : counter + 1;
+    counterElement.innerHTML = counter + 1;
   }
   section.querySelectorAll('[data-add-another-dynamic-elements]').forEach((element) => {
     element.dataset.addAnotherDynamicElements.split(",").forEach((pairString) => {
--- a/app/services/check_answers/section_lister_service.rb
+++ b/app/services/check_answers/section_lister_service.rb
@@ -80,16 +80,11 @@ module CheckAnswers
       return if field_data[:skip_unless].present? && !model.send(field_data[:skip_unless])
       return if field_data[:skip_if].present? && model.send(field_data[:skip_if])
 
-      label = if index
-                index.zero? ? "#{field_data.fetch(:attribute)}_first" : "#{field_data.fetch(:attribute)}_subsequent"
-              else
-                field_data.fetch(:attribute)
-              end
       addendum = "_partner" if @check.partner && field_data[:partner_dependant_wording]
 
       disputed = field_data[:disputed_if].present? && @check.smod_applicable? && model.send(field_data.fetch(:disputed_if))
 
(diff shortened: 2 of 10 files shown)


app/javascript/add-another.js
```diff
@@ -85,7 +85,7 @@ const setUpSections = (sectionList) => {
 const setNumbering = (section, counter) => {
   const counterElement = section.querySelector('[data-add-another-role="counter"]')
   if (counterElement) {
-    counterElement.innerHTML = counterElement.dataset.addAnotherNumberFrom === "zero" ? counter : counter + 1;
+    counterElement.innerHTML = counter + 1;
```
MazOneTwoOne, line 88: question - so now the result will still be `: counter 1`, `: counter 2`, `: counter 3` but simplified as there will no longer be `data-add-another-number-from="zero"`

MazOneTwoOne, line 88: question - so the result will still be `: counter 1`, `: counter 2`, `: counter 3`, but simplified as there is no longer this (HTML label?) `data-add-another-number-from="zero"`?

MazOneTwoOne (approved):
Approved with some questions.

app/javascript/add-another.js
patrick-laa: So previously, the default was for the sequence  of counters was `1`, `2`,  3`, but if `data-add-another-number-from="zero"` was set it would go `0` `1` `2` (this was used for additional bank accounts, where the first one wouldn't have a label, then the second would be called `additional bank account 1` and the third would be `additional bank account 2`).

Now, for bank accounts, as with everything else, the first is called `bank account 1`, the second is `bank account 2` the third is `bank account 3`. So now we always start from 1, and never start from `0`, so  we've removed the option to start from zero.

app/services/check_answers/section_lister_service.rb
```diff
@@ -80,16 +80,11 @@ def build_field(field_data, model, table_label, index: nil)
...
-                index.zero? ? "#{field_data.fetch(:attribute)}_first" : "#{field_data.fetch(:attribute)}_subsequent"
-              else
-                field_data.fetch(:attribute)
-              end
       addendum = "_partner" if @check.partner && field_data[:partner_dependant_wording]
 
       disputed = field_data[:disputed_if].present? && @check.smod_applicable? && model.send(field_data.fetch(:disputed_if))
 
-      Field.new(label: "#{table_label}_fields.#{label}#{addendum}",
+      Field.new(label: "#{table_label}_fields.#{field_data.fetch(:attribute)}#{addendum}",
```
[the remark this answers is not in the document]
patrick-laa: When the `field_data[:many]` attribute was set (only happened for bank accounts), we could have an arbitrary number of rows within a table (for all other "add another" patterns, we would add arbitrary numbers of tables, not rows within tables).  That attribute would trigger the `build_many_fields` method to run, which would then call `build_field` passing in an `index`. The old code would detect the presence of that index and if it was there, use `_first` and `_subsequent` to construct the labels (so the first one would be "Money in bank accounts" and the second onwards would be "Additional bank account X"). Now we don't need any of that, because the format for all rows is "Money in bank account X".

app/javascript/add-another.js
MazOneTwoOne: That makes sense now. Thank you.

PR #905, opened 2023-08-17
EL-1100: Add a sampling function

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1100)

## What changed and why

Ensure that we only send a small % of 'status check' traffic to Sentry for performance monitoring purposes.

## Checklist


config/initializers/sentry.rb
```diff
@@ -1,5 +1,5 @@
 sentry_dsn = Rails.configuration.sentry_dsn
-if %w[production].include?(Rails.env) && sentry_dsn.present? && ENV["SENTRY_FEATURE_FLAG"]&.casecmp("enabled")&.zero?
+if sentry_dsn.present? && ENV["SENTRY_FEATURE_FLAG"]&.casecmp("enabled")&.zero?
```
MazOneTwoOne, line 2: ok - this is how we are applying this to all our environments.

MazOneTwoOne (approved):

config/initializers/sentry.rb
patrick-laa: Yeah, so a gotcha to look out for is that _all_ our Cloud Platform environments (production, staging and UAT) run in Rails's "production mode" as opposed to its development or test modes. So the check on this line used to say "if we're running in production mode and we have the right environment variables'. It now just says 'if we have the right environment variables', because that way if we want to run the code on our local machines in development mode (e.g. to play with and tweak the sampling function) we can just set appropriate environment variables in `.env.development`.

MazOneTwoOne: So more unintuitive than I initial thought 😅

PR #908, opened 2023-08-18
EL-1102: Make migration rake task more robust

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1102)

## What changed and why

Allow migrations to recover from an error that sometimes happens when 2 pods try to run them at the same time.

## Checklist

### DIFF
--- a/lib/tasks/jobs/migrations.rake
+++ b/lib/tasks/jobs/migrations.rake
@@ -1,12 +1,16 @@
 namespace :db do
   namespace :migrations do
-    desc "Run db:migrate and retry if ActiveRecord::ConcurrentMigrationError encountered"
+    desc "Run db:migrate and retry if 2-pods-running-this-at-once issues encountered"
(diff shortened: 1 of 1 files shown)


lib/tasks/jobs/migrations.rake
```diff
@@ -1,12 +1,16 @@
 namespace :db do
   namespace :migrations do
```
MazOneTwoOne, line 2: question - why could we not make the namespace more descriptive like `:migrate_again_and_retry` or something along those lines. Maybe it's a bit too wordy.

patrick-laa: So the "namespace" is a bit generic, like "broadly what is this about". Then line 4, the task name, tries to be a bit descriptive. I thought that "run_with_retry" in the context of "migrations" captured what we wanted here, but yes, you could be as descriptive as you like.

PR #913, opened 2023-08-21
EL-1052: Remove WAF

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1052)

## What changed and why

The WAF in readonly mode isn't telling us anything useful.
The WAF in "intervention" mode would raise the possibility of false positives breaking the site for some users
Nothing the WAF picks up is stuff we think our site couldn't handle.

So we're going to remove the WAF.

## Checklist


helm_deploy/laa-estimate-eligibility/values/production.yaml
```diff
@@ -37,9 +37,6 @@ ingress:
   annotations:
     external-dns.alpha.kubernetes.io/set-identifier: "check-client-qualifies-laa-estimate-eligibility-laa-estimate-financial-eligibility-for-legal-aid-production-green"
     external-dns.alpha.kubernetes.io/aws-weight: "100"
-    nginx.ingress.kubernetes.io/enable-modsecurity: "true"
-    nginx.ingress.kubernetes.io/modsecurity-snippet: |
```
MazOneTwoOne, line 41: question - is this the nginx stuff that was discussed in 'Dev time', the "stuff" being that nginx is the controller which handles the firewall in our ingress? 

I'm assuming we're removing this part as we no longer need modsecuirty to be our WAF.

MazOneTwoOne (approved):

helm_deploy/laa-estimate-eligibility/values/production.yaml
patrick-laa: Exactly, we're removing the stuff that tells nginx "add modsecurity into the mix".

PR #925, opened 2023-08-30
EL-1115 add separate checks for readiness and liveness

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1115)

separated liveness and readiness checks during deploy process
delete pvc after merge or close PR

## Checklist

### DIFF
--- a/.github/actions/delete-uat-release/action.yml
+++ b/.github/actions/delete-uat-release/action.yml
@@ -74,6 +74,8 @@ runs:
         then
           helm delete $release_name
           echo "##[set-output name=message;]$(echo "Deleted UAT release ${release_name}!")"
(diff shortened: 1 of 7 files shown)


patrick-laa (commented):
Am I right that EL-1115 is going to get a part 2 pull request? By separating out the checks like this without making the database persistent, if the DB pod gets cycled and the database goes away, nothing will happen to cause the migrations to be re-applied. I wonder if it might be good to merge the change to keep the database persistent _before_ we merge this change, or at least at the same time, to avoid losing main UAT in the interim?

willc-work:
@patrick-laa thanks, it was not intentionally split, I just seemed to lose some commits. Added them back in now.

patrick-laa (approved):

PR #932, opened 2023-09-01
EL-1081: Improve wording of error messages in add another scenarios

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1081)

## What changed and why

Use different error messages in 'add another' scenarios if there are multiple items on the page.

## Checklist


app/forms/concerns/add_anotherable.rb
```diff
@@ -51,10 +51,22 @@ def blank_model
...
     items.each_with_index do |item, index|
-      item.errors.messages.each do |field, messages|
-        errors.add(:"items_#{index + 1}_#{field}", messages.first)
+      item.errors.each do |error|
+        errors.add(:"items_#{index + 1}_#{error.attribute}", error_message(error, items.length, item.class, index + 1))
       end
     end
   end
+
+  def error_message(error, number_of_items, model_class, position)
```
MazOneTwoOne, line 64: ok - so we'll have to have a `_when_many` error message any time we use the add another pattern. Is that worth mentioning in the `error_message` method as a comment??

MazOneTwoOne (approved):

app/forms/concerns/add_anotherable.rb
patrick-laa: So it's more that we'll have a `when_many` error message when we're (1) using the add another pattern and (2) there is more than 1 item added. I was hoping that the former was clear because this file is the AddAnotherable concern - so it's specifically the place for defining how 'add another' works, and that the latter was clear from the content of the method.

MazOneTwoOne: That's fair and not a big 'mental leap' to make/remember when using 'add another pattern'.

PR #943, opened 2023-09-06
EL-1132: Discount both k8s check endpoints from sentry profiling

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1132)

## What changed and why

Update the check to include "health" as well as "status", but also check that those are the whole of the URL, not some longer URL that just includes that word.

## Checklist

### DIFF
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -12,7 +12,7 @@ if sentry_dsn.present? && ENV["SENTRY_FEATURE_FLAG"]&.casecmp("enabled")&.zero?
     # (there are lots and lots of status checks, and we want to know if there are
     # problems but we don't want to clog up Sentry with data)
     config.traces_sampler = lambda do |sampling_context|
-      /\A\/(status|health)\Z/.match?(sampling_context[:env]["PATH_INFO"]) ? 0.01 : 0.1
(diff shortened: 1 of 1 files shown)


config/initializers/sentry.rb
```diff
@@ -12,7 +12,7 @@
     # (there are lots and lots of status checks, and we want to know if there are
     # problems but we don't want to clog up Sentry with data)
     config.traces_sampler = lambda do |sampling_context|
-      /status/.match?(sampling_context[:env]["PATH_INFO"]) ? 0.01 : 0.1
+      /\A\/(status|health)\Z/.match?(sampling_context[:env]["PATH_INFO"]) ? 0.01 : 0.1
```
MazOneTwoOne, line 15: question (not a blocker) - I think lower case 'z' would work, just as well. Why would we want a new line accepted in this regex?

> \z
> Matches the end of the string.
> \Z
> Matches the end of the string unless the string ends with a ``\n'', in which case it matches just before the ``\n''.

MazOneTwoOne (approved):

config/initializers/sentry.rb
patrick-laa: Ah! That's me not having great regex-fu. It makes no practical difference (there will never be a newline character in the path), but `\z` is more appropriate here. Good spot, I will fix.

PR #955, opened 2023-09-13
[EL-1136] Add event tracking for external CW form link

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1136)

## What changed and why

We want to be able to track when the external CW form link is clicked so that we can understand how our users are interacting with CW forms.

In this PR I have added in a redirect so we can trigger an event before sending the user on to the external link.

## Guidance to review

I'm not sure if this is the best way to go about this or if it is a bit hacky so good to know thoughts. Maybe Javascript is the way to go but I wasn't sure how to make that work with our existing analytics service. I also don't know how to write Javascript 🙃 

Please check that this makes sense, doesn't open us to any vulnerabilities with how I have set up the redirect. Please see if you think the URL makes sense/needs changing. 

Also with [EL-823](https://github.com/ministryofjustice/laa-estimate-financial-eligibility-for-legal-aid/pull/946) in UAT, that PR makes some changes to the controllers so this will need a rework once that gets merged. EL-823 also adds a `RedirectsController` so I'm thinking of moving the controller code into there, if we intend to keep that controller long-term.

This is a bit of a first pass so probably could do some refactoring as well. Once we are happy with implementation I'll write the accompanying Blazer query.

## Checklist


app/controllers/estimates_controller.rb
```diff
@@ -51,6 +51,11 @@ def download
     end
   end
 
+  def external_cw_form_redirect
```
patrick-laa, line 54: As this method is about sending the user to an external document, I think the natural place for it is in `DocumentsController`, which does exactly that. You could even reuse the `show` method, i.e. if the URL was `/documents/cw_forms?referrer=view_results&assesment_code=foo` then (so long as you added `cw_forms` to GuidanceLinkService), the link tracking would be handled for you automatically by the existing code. But equally it's not really a guidance link so you could argue that it's better to give it its own endpoint.

(BTW, so long as the analytics event you use starts with `click_`, the [existing link click Blazer query](https://check-your-client-qualifies-for-legal-aid.service.gov.uk/data/queries/24-guidance-link-clicks) will show up clicks alongside guidance link clicks, which is potentially quite useful

app/helpers/application_helper.rb
```diff
@@ -64,4 +64,8 @@ def using_non_primary_url?
   def survey_link
     t("service.public_beta_survey_link")
   end
+
+  def external_cw_form_link
```
patrick-laa, line 68: As this helper is a one-liner that's only used in one place, I think you could reasonably just move the content of this method into the controller. (If you're wondering how to access the `t` helper within a controller, the way we typically do it is by calling `I18n.t`, but now I wonder if `helpers.t` would also work)

config/routes.rb
```diff
@@ -33,6 +33,8 @@
   get "/auth/subdomain_redirect", to: "oauth_redirects#subdomain_redirect", as: :subdomain_redirect
   post "/auth/google/redirect", to: "oauth_redirects#google_redirect", as: :google_oauth_redirect
 
+  get "/external/:id", to: "estimates#external_cw_form_redirect", as: :external
```
patrick-laa, line 36: I'd prefer a slightly more expressive name - reading `external_path` in the view file confused me. Whereas `external_cw_form_redirect_path` would tell me everything I needed to know.

patrick-laa (approved):
I've added some suggestions for refactors, but equally I don't think there's anything actually wrong with your approach, so approving it if you want to keep it as is.

app/controllers/estimates_controller.rb
HettieS: I did wonder about just tapping into the existing stuff in DocumentsController. I was just a bit reluctant as I didn't think the external webpage was like a 'document' as such (but I guess it kind of is) and, as you've said, it's not a guidance link which is why I ended up making a new method. BUT it would be nice to reuse stuff so maybe I can play around with a bit of renaming.

HettieS: Also that's great news about the Blazer query 🙌

patrick-laa: Yeah, both DocumentsController and GuidanceLinkService are potentially badly named, because some of the links are to web pages with legislation, which are neither "documents" in the typical sense nor "guidance". So I'd definitely 👍 some renaming!

patrick-laa (approved):
This looks great - I heartily approve of the rename. I think reusing `document_link` works wel here.

PR #957, opened 2023-09-14
EL-1138: Show visual feedback when removing an 'add-another' instance

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1138)

## What changed and why

Show a little widget when an instance is removed to make it clearer that something has been taken away.

## Checklist

### DIFF
--- a/app/javascript/add-another.js
+++ b/app/javascript/add-another.js
@@ -71,10 +71,31 @@ const setUpSuggestions = (section) => {
 
 const remove = (section) => {
   const sectionList = section.closest('[data-add-another-role="sectionList"]')
+  showItemRemovedFeedback(section);
   section.remove();
   setUpSections(sectionList); // Some JS triggers, in particular radio conditional reveals, must be re-initialised
   setUpAddButton(sectionList.closest('[data-module="add-another"]'));
 }
+
+const showItemRemovedFeedback = (section) => {
+  const topLevelElement = section.closest('[data-module="add-another"]');
+  const feedback = document.createElement("div");
+  feedback.className = "add-another-removed-feedback";
+
+  const text = document.createElement("div");
+  text.className = "add-another-removed-feedback-text govuk-body";
+  text.innerHTML = topLevelElement.dataset.addAnotherRemovedFeedbackText;
+
+  const button = document.createElement("div")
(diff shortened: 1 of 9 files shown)


app/javascript/add-another.js
```diff
@@ -71,11 +71,32 @@ const setUpSuggestions = (section) => {
...
 }
 
+const showItemRemovedFeedback = (section) => {
+  const topLevelElement = section.closest('[data-module="add-another"]');
+  const feedback = document.createElement("div");
+  feedback.className = "add-another-removed-feedback";
+
+  const text = document.createElement("div");
+  text.className = "add-another-removed-feedback-text govuk-body";
+  text.innerHTML = topLevelElement.dataset.addAnotherRemovedFeedbackText;
```
MazOneTwoOne, line 87: ok - I know see that this bit `addAnotherRemovedFeedbackText`, is being told what text to show inside the feedback element box thing.

app/javascript/add-another.js
```diff
@@ -71,11 +71,32 @@ const setUpSuggestions = (section) => {
...
+
+  const button = document.createElement("div")
+  button.className = "add-another-removed-feedback-button govuk-body";
+  button.innerHTML = topLevelElement.dataset.addAnotherHideMessageText;
+
+  feedback.append(text);
+  feedback.append(button);
+  section.after(feedback);
+
+  button.addEventListener("click", () => { feedback.remove() });
```
MazOneTwoOne, line 97: question - how does this disappear when you click 'Save and continue'?

MazOneTwoOne (approved):
Just one question but happy to approve.

app/javascript/add-another.js
patrick-laa: So the 'Save and continue' button triggers a (non-AJAX) POST, where the browser says to our server "Here is the contents of the form, tell me what to display next". If the form is invalid, the server returns a new version of the page HTML  that's constructed from scratch by the server, so has no knowledge of any extra bits of HTML that were added to the previous page by the client-side javascript. The browser throws away the old page and shows the new page in its place, so those bits seem to disappear because they were in the old version but aren't in the new version.

PR #958, opened 2023-09-14
EL-1134 PDF populate with zeros

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1134)

## What changed and why

Show 0 in capital section and dont mask negative values

## Checklist

### DIFF
--- a/app/models/concerns/asset_summarisable.rb
+++ b/app/models/concerns/asset_summarisable.rb
@@ -5,11 +5,13 @@ module AssetSummarisable
   end
 
   def smod_savings
-    bank_accounts.select(&:account_in_dispute).sum { _1.amount.to_d } if bank_accounts&.select(&:account_in_dispute)&.any?
+    return unless any_smod_assets?
+
+    bank_accounts.select(&:account_in_dispute).any? ? bank_accounts.select(&:account_in_dispute).sum { _1.amount.to_d } : 0
   end
 
   def non_smod_client_savings
-    bank_accounts.reject(&:account_in_dispute).sum { _1.amount.to_d } if bank_accounts&.reject(&:account_in_dispute)&.any?
+    bank_accounts&.reject(&:account_in_dispute)&.any? ? bank_accounts.reject(&:account_in_dispute).sum { _1.amount.to_d } : 0
   end
 
   def partner_savings
@@ -18,20 +20,24 @@ module AssetSummarisable
 
   # Investments
   def smod_investments
-    investments if investments_in_dispute
+    if any_smod_assets?
+      investments_in_dispute ? investments : 0
+    end
   end
 
   def non_smod_client_investments
-    investments unless investments_in_dispute
+    investments_in_dispute ? 0 : investments || 0
   end
 
   # Valuables
   def smod_valuables
-    valuables if valuables_in_dispute
+    if any_smod_assets?
+      valuables_in_dispute ? valuables : 0
+    end
   end
 
   def non_smod_client_valuables
-    valuables unless valuables_in_dispute
+    valuables_in_dispute ? 0 : valuables || 0
   end
 
   # Totals
@@ -40,10 +46,10 @@ module AssetSummarisable
   end
 
   def combined_non_disputed_capital
-    from_cfe_payload("result_summary.capital.combined_non_disputed_capital") if client_capital_relevant?
+    from_cfe_payload("result_summary.capital.combined_non_disputed_capital") || 0
   end
 
   def combined_assessed_capital
-    from_cfe_payload("result_summary.capital.combined_assessed_capital") if client_capital_relevant?
+    from_cfe_payload("result_summary.capital.combined_assessed_capital") || 0
   end
 end
--- a/app/models/concerns/asset_summarisable.rb
+++ b/app/models/concerns/asset_summarisable.rb
@@ -46,10 +46,14 @@ module AssetSummarisable
   end
 
   def combined_non_disputed_capital
+    return if asylum_support?
+
     from_cfe_payload("result_summary.capital.combined_non_disputed_capital") || 0
   end
 
   def combined_assessed_capital
+    return if asylum_support?
+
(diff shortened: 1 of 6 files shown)


patrick-laa (commented):
So for me the main problem I have is still being clear on what the requirements are. I think what we agreed is something like: If a section is skipped, fill in no fields in that section. Otherwise, fill in all fields in that section (and treat the partner column like a separate section that's skipped if there's no partner).

If that's right, then I think this is a bit incomplete - I'm still seeing a lot of blank fields in the income section when the income section isn't skipped (see screenshot):
Screenshot 2023-09-14 at 16 13 06

And likewise I think if a client doesn't own their main home but the capital section isn't skipped, I don't think the home fields should be blank in the capital section (see screenshot:)

Screenshot 2023-09-14 at 16 13 25

Then when the client gets asylum support and all sections are skipped, I think all fields in those sections should be blank, but I'm now seeing some zeroes in the capital section (see screenshot:)

Screenshot 2023-09-14 at 16 14 12

And finally, when there are some SMOD assets, so part A is _not_ skipped, I think part A shouldn't have any blank fields, but I'm seeing a lot (see screenshot:)

Screenshot 2023-09-14 at 16 15 20

It might be that I'm missing the point here, so I'm writing this up as an overall comment before I start wrongly correcting specific lines of code.

app/models/concerns/property_summarisable.rb
```diff
@@ -2,24 +2,24 @@ module PropertySummarisable
   # MAIN HOME
   # Value
   def main_home_value
-    from_cfe_payload("assessment.capital.capital_items.properties.main_home.value") if owns_property?
+    from_cfe_payload("assessment.capital.capital_items.properties.main_home.value") || 0
```
patrick-laa, line 5: So it looks like the property fields all need an asylum support check, otherwise this fills in zeroes in a section that should be left blank.

app/models/concerns/property_summarisable.rb
```diff
@@ -2,24 +2,24 @@ module PropertySummarisable
...
   end
 
   def smod_main_home_value
     main_home_value if house_in_dispute
```
patrick-laa, line 9: This, and all the other SMOD property fields, need a check whether there are any SMOD assets. If there are any SMOD assets, then for example this field should return a zero if the main home isn't SMOD. Otherwise it should return nil

app/models/concerns/outgoings_summarisable.rb
```diff
@@ -1,27 +1,31 @@
 module OutgoingsSummarisable
   def client_mortgage
-    net_housing_costs if client_income_relevant? && property_owned_with_mortgage?
+    property_owned_with_mortgage? ? net_housing_costs : 0
```
patrick-laa, line 3: These outgoings fields are now being filled out both if the client gets asylum support and if a client gets passporting benefit, which they shouldn't be.

app/models/concerns/asset_summarisable.rb
```diff
@@ -18,20 +20,24 @@ def partner_savings
 
   # Investments
   def smod_investments
-    investments if investments_in_dispute
+    if any_smod_assets?
```
patrick-laa, line 23: It might be worth being consistent about how we do these checks - in `smod_savings` we're doing a `return unless`, here we're doing an `if` - for readability, since it's applying the same logic in every method, ideally we'd do it in the same way.

app/models/concerns/asset_summarisable.rb
```diff
@@ -18,20 +20,24 @@ def partner_savings
...
+      investments_in_dispute ? investments : 0
+    end
   end
 
   def non_smod_client_investments
-    investments unless investments_in_dispute
+    investments_in_dispute ? 0 : investments || 0
```
patrick-laa, line 29: This is still not taking into account asylum support.

app/models/concerns/outgoings_summarisable.rb
```diff
@@ -32,36 +44,56 @@ def not_all_child_dependants_are_over_16
...
+    return if asylum_support? || !client_income_relevant?
+
+    from_cfe_payload("result_summary.disposable_income.maintenance_allowance") || 0
   end
 
   def partner_maintenance_allowance
-    from_cfe_payload("result_summary.partner_disposable_income.maintenance_allowance") if partner_income_relevant?
+    return if asylum_support? || !client_income_relevant?
+
+    from_cfe_payload("result_summary.partner_disposable_income.maintenance_allowance") || 0 if partner
```
patrick-laa, line 90: So here we return nil if `asylum_support?` or `!client_income_relevant?` or (implicitly) `!partner`. But we already had a single convenience method that did all of this that we were previously using - `partner_income_relevant?`. So let's use that.

app/models/concerns/outgoings_summarisable.rb
```diff
@@ -1,27 +1,39 @@
...
   def client_mortgage
-    net_housing_costs if client_income_relevant? && property_owned_with_mortgage?
+    return if asylum_support? || !client_income_relevant?
+
+    property_owned_with_mortgage? ? net_housing_costs : 0
   end
 
   def client_rent
-    net_housing_costs if client_income_relevant? && !owns_property?
+    return if asylum_support? || !client_income_relevant?
```
patrick-laa, line 9: This is a bit redundant - `client_income_relevant?` already incorporates a check for `asylum_support`

patrick-laa (commented):
These are potentially notes to self, as it looks like I might be picking up these tickets, but these are the issues we need to work through, I think.

patrick-laa:
Closing this because I'm going in a slightly different direction in #975 . I think we've demonstrated here that trying to sort everything out with a million duplicated or nearly-duplicated `if` clauses is asking for trouble - if you forget one scenario where a section should be skipped, you have to go back and make changes to every `if` clause of every method that populates that section.

So instead I'm trying specifying the section of every field, and _first_ running some logic to see whether a given section should be skipped, and only then looking to populate that section.

PR #959, opened 2023-09-14
EL-1164: Use new partial to ensure link_errors always set

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-164)

## What changed and why

- Add a new partial that handles setting up radio buttons within a govuk_radio_button_fieldset, including a `link_errors`
- Ensure that this partial gets used in all cases where radio buttons are created
- (There's one exception, the old "employment" form, but that's never used and is being deleted as part of EL-823, so I thought no point changing that view here.

## Checklist


app/views/question_flow/applicant.html.slim
```diff
@@ -26,14 +26,10 @@
...
 
-    = form.govuk_radio_buttons_fieldset(:partner,
-            legend: { text: t(".partner.legend") },
-            hint: -> { content_for(:partner_hint) })
-      = form.govuk_radio_button :partner,
-              true,
-              label: { text: t("generic.yes_choice") },
-              link_errors: true
-      = form.govuk_radio_button :partner, false, label: { text: t("generic.no_choice") }
+    = form.govuk_collection_radio_buttons(:partner,
```
MazOneTwoOne, line 29: question - so are we using the `govuk_collection_radio_buttons` class/element, rather than the partial added to `govuk_radio_buttons_fieldset`, because these options are just a generic "Yes/No" (and we have nice helper methods)?? 

I'm guessing that if the radio button options are used more than once we'd want to stick them here: **app/helpers/forms_helper.rb** and use `govuk_collection_radio_buttons`. Otherwise, we'd use `govuk_radio_buttons_fieldset` with the `shared/radio_button_set` partial. I'm sure it depends on the scenario.

app/views/question_flow/forms/_employment_status.html.slim
```diff
@@ -1,17 +1,23 @@
 - partner ||= false
 = form_for(@form, url: request.path, method: :put) do |form|
-  = render "shared/heading",
-            header_text: t("question_flow.#{i18n_key}.heading"),
-            tag: (:partner if partner)
+  = render "shared/heading", tag: (:partner if partner)
```
MazOneTwoOne, line 3: ok - this is just us tidying up how we populate the headings (nothing to do with radio buttons).

app/views/question_flow/forms/_employment_status.html.slim
```diff
@@ -1,17 +1,23 @@
...
-  = form.govuk_radio_buttons_fieldset :employment_status, legend: nil
-    = form.govuk_radio_button :employment_status, EmploymentStatusForm::EMPLOYMENT_STATUSES[:in_work],
-            label: { text: t("question_flow.#{i18n_key}.in_work_including_self_employed") },
-            hint: { text: t("question_flow.#{i18n_key}.in_work_including_self_employed_description_#{@check.level_of_help}") }
-    = form.govuk_radio_button :employment_status, EmploymentStatusForm::EMPLOYMENT_STATUSES[:unemployed],
-            label: { text: t("question_flow.#{i18n_key}.unemployed") },
-            hint: { text: t("question_flow.#{i18n_key}.unemployed_description") }
+  = form.govuk_radio_buttons_fieldset :employment_status,
+                                      legend: { text: t("question_flow.#{i18n_key}.heading"), size: "xl", tag: "h1" }
+    ruby:
```
MazOneTwoOne, line 9: question - in the instances where we use ruby, it's either because:

- we want, the responses (i.e `@check.level_of_help`) known about by another page in the flow, not exclusively for the end result page?
- it's fiddly to get the labels from the translations/hashes of where they currently are?

MazOneTwoOne (approved):
Approved (with a bunch of questions just for my benefit).

app/views/shared/_radio_button_set.html.slim
```diff
@@ -0,0 +1,17 @@
+/ Use this whenever we need to add radio buttons to a govuk_radio_buttons_fieldset
+/ The reason is that this handles the easily-forgettable "link_errors" param.
+
+/ Required values:
+/ form - a form object yielded by a `form_for` or `form_with` call
+/ field_name - the name of the field shared by all radio buttons
+/ buttons - an array of objects. Each object is either a hash,  with `value` and `options` keys,
+/           or it's the symbol `:divider`
+
+/ Optionally a block can be provided, which will be the conditional reveal if the first item is selected
```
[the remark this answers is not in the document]
patrick-laa: Yes, although from EL-1171 I think we've learned that "block_given?" doesn't behave the way I think it behaves, so I can probably remove it from line 17 as I think it always returns `true`. I'll check and ditch it if so.

app/views/question_flow/forms/_employment_status.html.slim
patrick-laa: Yes, this is standardising - normally for 1-question pages we set the header as the legend of the form element, but we hadn't done that here.

app/views/question_flow/forms/_employment_status.html.slim
patrick-laa: So not exactly. In these slim files, you can write a line of ruby code either by starting the line with `-` (or `=`) or by indenting it under a `ruby:` statement. The linter suggests that if you have _multiple_ lines of ruby code you need to write, it's best to put them under `ruby:`, and that also makes it easy to write multi-line ruby statements, because slim sometimes gets confused about interpreting line breaks.

PR #960, opened 2023-09-15
Try to handle stuck deploys

patrick-laa:


willc-work (approved):

exonian (commented):
My only concern is that, while failing rollbacks are caught and aborted, it looks to me as though if rollbacks work but deploys still for some reason don't, then this goes infinite?

I don't know if that's at all likely, I don't know if it matters that much if it happens, so therefore I don't know if it's worth having a RETRIES var that limits this 🤔

patrick-laa:
@exonian it doesn't actually loop - it tries `deploy_branch`, if it fails it tries to rollback and then tries `deploy_branch` a second time, and then the script ends. I didn't think it was worth trying the rollback trick more than once - it'll either work or it won't.

exonian:
> @exonian it doesn't actually loop - it tries `deploy_branch`, if it fails it tries to rollback and then tries `deploy_branch` a second time, and then the script ends - if it failed the second time it'll end with a non-zero exit code and CI will abort. I didn't think it was worth trying the rollback trick more than once - it'll either work or it won't.

Of course you're right! Cos I was looking at just the diff on github it had scrunched away the middle of the file where there were no changes, and I'd failed to notice. So I thought that the retry was *within* `deploy_branch` (and hadn't noticed that `deploy_main` was referenced but apparently didn't exist...)

PR #962, opened 2023-09-15
EL-1129: dependant income content

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1129)

## What changed and why

Add content to make clear about dependant maximum allowable income.

## Checklist


app/views/shared/_heading.html.slim
```diff
@@ -33,7 +33,7 @@ ruby:
         ul style="list-style-type: none; margin: 0; padding: 0;"
           li = header_text
     - else
-      h1.govuk-heading-xl class=("govuk-!-margin-bottom-3" if post_header_texts.any?) = header_text
+      h1.govuk-heading-xl class=("govuk-!-margin-bottom-3" if post_header_texts.any? || end_block.present?) = header_text
```
MazOneTwoOne, line 36: question - I'm overthinking this, but why not add the spacing (`"govuk-!-margin-top-3"`) to `govuk_warning_text`? 

There might (this word is doing a lot of 'heavy lifting') be scenarios when we just want the element below the header to stand out more and not be part of the header??

MazOneTwoOne (approved):

app/views/shared/_heading.html.slim
patrick-laa: We could, but if we want an element below the header to stand out and not be part of the header, all we need to do is not provide that element as a block indented within the shared/heading block (e.g. in dependant_details.html.slim we could unindent the new line 31 by 2 new spaces). By doing it this way, we say:

- if you want the new thing to be tucked up tight under the header, provide it as a block to this partial
- if you want it not to, don't.

MazOneTwoOne: Oh yes! I knew I was overthinking it.

PR #975, opened 2023-09-19
EL-1134: Show negative numbers and replace blank fields with zeroes

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1134)

## What changed and why

* Show negative numbers if we have them
* If an entire _section_ is irrelevant we skip it, otherwise we fill in every field with a zero if we don't have an actual value for it

## Guidance to review

I'm taking a new approach after discussion with Will on #958.

## Checklist


app/lib/controlled_work_mappings/cw5_header.yml
```diff
@@ -21,325 +21,389 @@
...
 # Rent / Partner
 - name: FillText38
   type: text
+  section: partner_income
   source: partner_rent
 
 # Dependents' allowances / Partner
 - name: FillText39
   type: text
+  section: income
```
HettieS, line 316: should this be `partner_income`?

HettieS (commented):
This looks great, just one small error I think.

Have tried to follow the thread on this one as know there was a lot of discussion - just want to check that in `def skip?` where we use our `Steps::Helper` to know whether to skip a section, this is acceptable and we don't need to add zeros for any scenarios? I'm just trying to think if there could be any cases where we would want zeros instead of a blank field.

app/lib/controlled_work_mappings/cw5_header.yml
patrick-laa: @HettieS I don't think it should. The partner allowance is not in the partner section, it's in the client section. If the client doesn't have a partner I don't think it should be left blank, I think it should be zero (in the same way that if the client doesn't have child dependants the child dependant field is left zero).

Conceptually, it's because  if the client doesn't have a partner, "How much partner allowance does the client get?" is still a meaningful question, while, for example, "What are the partner's gross monthly earnings?" is not.

patrick-laa:
@HettieS my understanding is that where a section is explicitly skipped, we don't want to fill in _anything_ in that section. If it's not explicitly skipped, we want to fill in _everything_ in that section.

Although thinking this through did reveal a bug - we want to skip the income section if the client receives a passporting benefit. But we passporting benefit question itself needs to not be skipped. I had had it marked as inside the income section, but while it is visually, conceptually it's in its own section, one that is only skipped if the client receives asylum support.

app/lib/controlled_work_mappings/cw5_header.yml
HettieS: Ooooh I see. My mistake for not looking at the forms properly. I noticed for partner dependant allowance, in some forms you have `partner_income` and others you have `income` - so on cw5 it's `income` and in cw2 it is `partner_income` for example. I'm just looking at the forms trying to work it out....

patrick-laa: Ohhh, my bad, so yeah, I got it wrong on CW2. Fixing now.

HettieS (approved):

PR #976, opened 2023-09-19
EL-1171: Extra spacing on "Yes/No" radio buttons

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1171)

## What changed and why

Making sure no additional spacing is created, when selecting 'Yes' on the "Yes/No" radio buttons. 

## Guidance to review

I initially added in a bit of CSS to this file: **app/assets/stylesheets/add-another.scss**, but this removes the space from the conditional reveal input field, hence taking the approach with a new partial. Open to another approach. 

> ```
> .govuk-radios__conditional {
>   margin-bottom: 10px;
> }
> ```

Before:
Screenshot 2023-09-19 at 14 07 20

After:
Screenshot 2023-09-19 at 14 07 35

## Checklist


app/views/shared/add_another/_yes_no_radios_conditional_reveal.html.slim
```diff
@@ -0,0 +1,53 @@
...
+/ value
+/ counter
+
+/ Optional:
+/ hint
+/ Block with conditional reveal content for 'Yes' choice
+ruby:
+  radio_name_pattern = "#{model_name}[items][ID][#{field_name}]"
+  radio_name = radio_name_pattern.gsub("ID", counter.to_s)
+  conditional_reveal_pattern = "conditional-reveal-ID-#{field_name.dasherize}" if block_given?
```
patrick-laa, line 16: You can delete `if block_given?` because there will _always_ be a block given.

app/views/shared/add_another/_yes_no_radios_conditional_reveal.html.slim
```diff
@@ -0,0 +1,53 @@
...
+    .govuk-radios data-module="govuk-radios"
+      = render "shared/add_another/radio_item",
+                id_pattern: yes_id_pattern,
+                name_pattern: radio_name_pattern,
+                counter:,
+                value: "true",
+                checked: value,
+                label: t("generic.yes_choice"),
+                conditional_reveal_pattern:
+      - if block_given?
```
patrick-laa, line 42: And likewise you don't need to wrap this bit in an `- if block_given?` because there always will be a block given.

patrick-laa (approved):
This approach works and is entirely reasonable.

The "root cause" seems to be that when I wrote this code I assumed that if I didn't provide content for a conditional reveal when using the "yes_no_radios" partial, `block_given?` would be `false`. However it turns out it was always returning `true`, and so we were always rendering the `.govuk-radios__conditional` block with nothing inside it, which was causing layout issues.

An  alternative approach would be to find a better test than `block_given?`. One thing that works is if at the top of yes_no_radios you say `conditional_reveal_content = yield`. If you've actually provided content for a conditional reveal, it will be rendered to a string assigned to this variable. Otherwise `conditional_reveal_content` will be an empty string. Then you can replace `block_given?` with `conditional_reveal_content.present?` and `= yield` with `= conditional_reveal_content`.

BUT I don't think that's objectively better than just splitting out the conditional reveal one into a separate partial, so I'm very happy to go along with your chosen approach.

PR #978, opened 2023-09-20
EL-1141: Add error messaging when dependant income hits the limit

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1141)

## What changed and why

Dependants that earn too much should not be counted in the assessment. If a user tries to add a dependant income that hits this limit, then the income field will present an error and will not allow them to add the amount. It will also prompt them to go back and amend their dependant count. 

## Guidance to review

Nothing particular on this one.

## Checklist


app/models/dependant_income_model.rb
```diff
@@ -5,14 +5,31 @@ class DependantIncomeModel
...
+  validate :income_too_high
+
   def frequency_options
     IncomeModel.income_frequency_options(skip_year: true)
   end
+
+  def income_too_high
+    return unless amount
+
+    if DEPENDANT_INCOME_UPPER_LIMITS[frequency] <= amount
```
patrick-laa, line 31: This will break if the user has submitted an amount (to get past the guard clause on line 29) but not a frequency.

spec/forms/dependant_income_details_form_spec.rb
```diff
@@ -48,4 +48,78 @@
...
+
+  describe "dependant income limits" do
+    context "when choosing monthly frequency" do
+      let(:monthly_limit) { "338.90" }
+
+      it "shows an error message when on the limit" do
+        fill_in "1-amount", with: monthly_limit
+        choose "1-frequency-monthly"
+        click_on "Save and continue"
+        expect(page).to have_content "Dependant income cannot be more than £338.90 per month."
```
patrick-laa, line 60: This to me looks incorrect. If the income is 338.90, an error message saying it cannot be _more_ than 338.90 doesn't seem like it applies. I suspect that we should only show the message if dependant income is 338.91 or above?

HettieS: I was going off the original A/C which I can see has just been changed, so will update accordingly.

app/models/dependant_income_model.rb
HettieS: 💡 silly mistake, will fix!

config/locales/en.yml
```diff
@@ -64,6 +64,7 @@ en:
               inclusion: Select when the dependant normally gets this income
             amount:
               blank: Enter income the dependant gets
+              dependant_income_too_high: Dependant income cannot be more than £338.90 per month. If you have counted someone as a dependant and they have income over £338.90 per month, you must remove them from the dependants entered on the previous page to continue with this check.
```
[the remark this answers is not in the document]
HettieS: good question - as we state '£338.90 per month' I was thinking that would be enough for the user to understand why their amount for other frequencies isn't accepted. But I'll ask UCD folks!

app/models/dependant_income_model.rb
patrick-laa: Ooh, there's one other niche issue here. `amount` is of type `gbp`, which has a slightly unusual behaviour: If the user enters a valid number, it will be parsed to a decimal, because functionally it's a number. _But_ if the user enters something invalid, the original value entered will be preserved as a string, so that when the page is re-rendered the original value is pre-populated in the field.

So there is a possibility that `amount` will be a string. What we do in other situations to work around this (e.g. EmploymentForm#net_income_must_be_positive or HousingCostsForm#housing_benefit_does_not_exceed_costs) is a guard clause checking `blah.to_i.positive?` or `errors.none?`, which will ensure if there is some other validation error, this logic won't run. But there may be a neater way.

HettieS: Oh thank you for this spot! Will fix this up too

HettieS:
@patrick-laa would you mind taking another look over this one please? 🙏

patrick-laa (approved):

PR #980, opened 2023-09-21
EL-1139: Dynamic add another messaging

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1139)

## What changed and why

* All error messages are now annotated to tie them to a specific 'add another' item
* When an item is removed, also remove all error messages that relate to it
* If we have no error messages left, remove the error summary box
* When an item is removed, update the numberings of all error messages for items that come after it
* And also update links
* (Separately: ensure we don't attach multiple event listeners to 'remove' buttons as that leads to remove functionality running twice)

## Checklist


app/forms/concerns/add_anotherable.rb
```diff
@@ -68,7 +69,10 @@ def items_valid?
   def error_message(error, number_of_items, model_class, position)
     key = "activemodel.errors.models.#{model_class.to_s.underscore}.attributes.#{error.attribute}.#{error.type}"
     if number_of_items > 1
-      I18n.t("#{key}_when_many", position:)
+      position_tag = tag.span(position, data: { add_another_role: "errorPosition" })
+      content = I18n.t("#{key}_when_many",
+                       position: position_tag)
```
MazOneTwoOne, line 74: ok - so we're doing this to add `[data-add-another-role="errorPosition"]` & `[data-add-another-role="errorMessage"]`to an error if there is more than one of them, and it used by the **updateErrorMessages** function.

app/javascript/add-another.js
```diff
@@ -63,6 +63,10 @@ const setUpRemoveButton = (section) => {
   if (!removeButton) {
     return;
   }
+  if (removeButton.dataset.removeListenerSet) {
```
MazOneTwoOne, line 66: question - am I right in thinking we are adding in an attribute `removeListenerSet` (set to `true`) to each of the remove buttons? I'm not following, why don't we set it up as an attribute from the beginning instead of searching for all the remove buttons and adding it then?

app/javascript/add-another.js
```diff
@@ -105,6 +110,38 @@ const showItemRemovedFeedback = (section) => {
...
+    // Decrement the numbers of all error messages that are for items that come after the removed item.
+    } else if (reachedSectionToRemove) {
+      const newPosition = index;
+      document.querySelectorAll(`[data-add-another-role="errorMessage"][data-add-another-item-position="${currentSectionCurrentPosition}"]`).forEach((elementToUpdate) => {
+        elementToUpdate.dataset.addAnotherItemPosition = newPosition;
+        elementToUpdate.querySelectorAll('[data-add-another-role="errorPosition"]').forEach((positionText) => {
+          positionText.innerHTML = newPosition;
+        })
+
+        elementToUpdate.closest("a").href = elementToUpdate.closest("a").href.replace(`-${currentSectionCurrentPosition}-`, `-${newPosition}-`);
```
MazOneTwoOne, line 132: ok - so once we've worked out which are the error messages after the one we've just removed, this line, is the _thing_ that replaces the current postion (`currentSectionCurrentPosition`) with the new position (`newPosition`).

MazOneTwoOne (approved):
Approved with comments.

p.s. It did take me a long while for me to get my "bearings" with the `updateErrorMessages` function. There are some bits I'm think I'm reading correctly, but could be way off.  That was a "long way" of saying, please do get another opinion if needed.

app/javascript/add-another.js
patrick-laa: What we are doing is adding this attribute to mark that we've set up the "remove" listener, so that for a given button we don't accidentally set up the listener twice. Otherwise when we click the remove button we might trigger the logic to run twice.

The logic basically says "If I haven't already marked this task as done, do the task and then mark the task as done"

app/javascript/add-another.js
patrick-laa: Yes, line 129 sets the visible text, line 132 updates the location of the clickable link

PR #981, opened 2023-09-21
EL-1161: Certificated Flow - Domestic Abuse & Upper tribunal questions 

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1161)

## What changed and why

Simplify the content for the start of the certificated journey. Adding new page on domestic abuse & new page for immigration and asylum upper tribunal cases. 

## Guidance to review

- Was going to conditional reveal content on the "which-matter-type" screen, based on 'level of help',   but wouldn't be able to change url. 
- Opted for 2 news screens and to adjust the flow at the start of the certificated journey. 

## Checklist


app/lib/steps/logic.rb
```diff
@@ -12,10 +12,14 @@ def upper_tribunal?(session_data)
         if controlled?(session_data)
           session_data["immigration_or_asylum"]
         else
-          session_data["matter_type"].in?(%w[immigration asylum])
+          session_data["immigration_or_asylum_type_upper_tribunal"].in?(%w[immigration_upper asylum_upper]) && !domestic_abuse_applicant?(session_data)
         end
       end
 
+      def domestic_abuse_applicant?(session_data)
+        session_data["domestic_abuse_applicant"]
```
patrick-laa, line 20: If we're being extra thorough, this should really have a `!controlled?(session_data)` as well - just in case someone changes their answers after saying yes to this question

app/services/journey_logger_service.rb
```diff
@@ -29,13 +29,15 @@ def build_attributes(calculation_result, check)
...
         asylum_support: check.asylum_support || false,
-        matter_type: matter_type(check),
+        matter_type: proceeding_type(check),
       }
     end
 
-    def matter_type(check)
-      if !check.controlled?
-        check.matter_type
+    def proceeding_type(check)
```
patrick-laa, line 36: I don't think we should call this proceeding type. "Proceeding type" means something quite specific, and is normally represented in the code by something like "DA001" or "IM030" etc. In this case, it is the matter type we are talking about. (There are multiple proceeding types associated with every matter type, I believe).

app/services/journey_logger_service.rb
```diff
@@ -29,13 +29,15 @@ def build_attributes(calculation_result, check)
...
+      if !check.controlled? && !check.domestic_abuse_applicant
+        check.immigration_or_asylum_type_upper_tribunal
```
patrick-laa, line 38: Previously, in a certificated check if someone chose the immigration option for certificated work we stored a value of "immigration". This code will cause it to store "immigration_upper". Which will make it harder to analyse our data. It'd be good to preserve the existing behaviour.

app/services/journey_logger_service.rb
```diff
@@ -29,13 +29,15 @@ def build_attributes(calculation_result, check)
...
+      elsif check.domestic_abuse_applicant
+        "domestic abuse"
```
patrick-laa, line 40: Likewise this should really be "domestic_abuse" to be consistent.

app/views/question_flow/domestic_abuse_applicant.html.slim
```diff
@@ -0,0 +1,24 @@
...
+  = t(".question")
+- content_for :back do
+  = back_link(:domestic_abuse_applicant, @check, @back_buttons_invoke_browser_back_behaviour)
+
+.govuk-grid-column-two-thirds
+  = render "shared/heading"
+  = form_for(@form, url: request.path, method: :put) do |form|
+    = form.govuk_error_summary t("generic.error_summary_title")
+
+    h1.govuk-heading-xl = t(".question")
```
patrick-laa, line 11: To be consistent with other screens, it would be better to set this as the `legend` attribute in the govuk_collection_radio_buttons rather than have it as its own element. I believe this affects exactly how classes are applied when there's a validation error.

config/locales/en.yml
```diff
@@ -10,10 +10,6 @@ en:
           attributes:
             asylum_support:
               inclusion: Select if your client receives section 4 or section 95 Asylum Support
-        matter_type_form:
-          attributes:
-            matter_type:
-              blank: Select what type of matter this is
```
patrick-laa, line 16: You'll need to add appropriate error messages for the new forms you've added.

config/locales/en.yml
```diff
@@ -650,13 +646,11 @@ en:
...
+      paragraph_1: This must be for legal representation for a domestic abuse matter (injunction or other order for protection from harm, or committal for breach), FGM protection order, or forced marriage protection order.
+      paragraph_2: We need to know this for the check calculations because upper eligibility limits are waived if your client is an applicant in a domestic abuse matter.
     matter_type:
-      title: Which type of matter is this?
-      immigration: Immigration (Upper Tribunal)
-      asylum: Asylum (Upper Tribunal)
-      other: Another category of law
-      domestic_abuse: Domestic abuse
-      domestic_abuse_hint: Upper eligibility limits are waived if your client is an applicant in a domestic abuse matter
       domestic_abuse_guidance:
```
patrick-laa, line 654: I think this should be under `domestic_abuse_applicant`, not `matter_type`

spec/forms/matter_type_form_spec.rb
```diff
@@ -1,22 +0,0 @@
-require "rails_helper"
```
patrick-laa, line 1: You should add equivalent form specs for the new screens you've added.

spec/services/cfe/proceedings_payload_service_spec.rb
```diff
@@ -9,7 +9,8 @@
       let(:session_data) do
         {
           "level_of_help" => "certificated",
-          "matter_type" => "other",
+          "domestic_abuse_applicant" => "false",
```
patrick-laa, line 12: This should strictly be `"domestic_abuse_applicant" => false,` (no quotation marks around the value).

patrick-laa (commented):
The overall approach is solid, we just need a few little tidy-ups.

app/views/question_flow/domestic_abuse_applicant.html.slim
MazOneTwoOne: Hey @patrick-laa , I just remembered why I did this. I separated out the heading from radio buttons, in order to follow the design on the ticket. On both of the new designs, there is a gap between heading and the paragraph text.

I did do something like this initially, but it doesn't look like the screen: 

```
- content_for :hint
  p.govuk-boy = t(".paragraph_1")
  p.govuk-boy = t(".paragraph_2")

.govuk-grid-column-two-thirds
  = render "shared/heading"
  = form_for(@form, url: request.path, method: :put) do |form|

    = form.govuk_collection_radio_buttons :domestic_abuse_applicant,
            yes_no_options, :first, :last,
            legend: { text: t(".question"), size: "xl", tag: "h1" },
            hint: -> { content_for(:hint) }
```

Would I add some stylings to create a gap??

patrick-laa: Ah ok. So that gap is present on the level of help screen, so you could copy the approach there (i.e. keep legend as `nil`, but instead of having a manual `h1`, pass in the heading as a `header_text` variable to the `shared/heading` partial). But also, I think I'm going to ask about this in the UCD channel, because looking at it we're already wildly inconsistent with this stuff and I think the new designs add yet more inconsistencies. It might be that if I point that out, the designs will change to something more standard.

MazOneTwoOne: I've added in some spacing. 👍🏽

HettieS:
Just had a thought on this one which may not be relevant - have you considered any potential changes to the metrics services? There may not need to be any changes but I've seen there are some attribute mappings in `Metrics::ForUserJourneyDashboardService`

MazOneTwoOne:
> @MazOneTwoOne Just had a thought on this one which may not be relevant - have you considered any potential changes to the metrics services? There may not need to be any changes but I've seen there are some attribute mappings in `Metrics::ForUserJourneyDashboardService`

Hey @HettieS. Yes, have looked at the potential impact of these changes to our metrics. 

We decided to update the `matter_type` attribute in `app/services/journey_logger_service.rb`. This way we hopefully still get these values`%w[immigration asylum domestic_abuse other]` mapped accordingly, from a certificated journey and correctly plugged into the right column into our metrics database.

patrick-laa (approved):
Looks good to me

PR #989, opened 2023-09-27
EL-1192: Update results page print and save PDF functionality

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1192)

## What changed and why

We want to remove the 'print' CTA from the results page as we don't think we need both 'print' and 'save' CTAs, when they more or less do the same thing and having both options doesn't benefit users.

## Guidance to review

- Please check this won't affect metrics, I don't think it will but the :print_results screen will no longer be a thing. Do we need to update any Blazer queries?
- I was a bit unsure if any of my name changes will affect the Grover set-up by I don't think it does, the PDF exports nicely still
- I've removed some javascript which I think was just for the old 'print' button, which is now now longer present

## Checklist


app/helpers/results_helper.rb
```diff
@@ -74,35 +74,20 @@ def pdf_friendly_paragraphs(text, is_pdf)
...
-        tag.ul(style: "list-style-type: none; margin: 0; padding: 0;") do
-          concat tag.li(I18n.t("results.print.legal_aid"))
-          concat tag.li(I18n.t("results.print.agency"))
-        end
-      end
-    else
-      tag.span(class: "gem-c-organisation-logo__name") do
-        concat tag.span(I18n.t("results.print.legal_aid"))
-        concat tag.br(I18n.t("results.print.agency"))
+  def pdf_friendly_logo
```
HettieS, line 77: I've removed `is_pdf` here - it was causing the code coverage to drop as we don't actually use this method other than on the download view which always takes `is_pdf: true` - so the `else` block never gets rendered. I couldn't work out the best way to run through it and test it seeing as we don't use it in any other context, so seemed a bit contrived!

app/helpers/results_helper.rb
```diff
@@ -74,35 +74,20 @@ def pdf_friendly_paragraphs(text, is_pdf)
...
+    tag.span(class: "gem-c-organisation-logo__name") do
+      tag.ul(style: "list-style-type: none; margin: 0; padding: 0;") do
+        concat tag.li(I18n.t("results.download.legal_aid"))
+        concat tag.li(I18n.t("results.download.agency"))
       end
     end
   end
 
-  def pdf_friendly_date(is_pdf)
-    if is_pdf
```
HettieS, line 94: same as above method

spec/system/accessibility_spec.rb
```diff
@@ -111,36 +111,4 @@
       expect(page).to be_axe_clean.skipping("aria-allowed-attr")
     end
   end
-
-  describe "Print results page" do
```
HettieS, line 115: deleted this spec as the print page is no longer a thing. I did update this to run through the code in download.html.slim but in running through this check, it caused the PDF to actually download and get added as a new file, and couldn't work out how to stop it from doing that!

patrick-laa (approved):
One minor refactor suggestion and one thing that can be safely deleted now, but other than that all looks good to me.

PR #997, opened 2023-10-03
EL-479: User feedback widget

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-497)

## What changed and why

Add feedback widget throughout the service. There are two variants - 'freetext' and 'satisfaction'. We will save the information entered by users into two new database tables so they can be accessed from the admin CMS.

## Checklist


db/schema.rb
```diff
@@ -24,20 +24,6 @@
     t.index ["record_type", "record_id", "name"], name: "index_action_text_rich_texts_uniqueness", unique: true
   end
 
-  create_table "active_admin_comments", force: :cascade do |t|
```
HettieS, line 27: not sure where this has gone so will fix

HettieS: I'm not sure why this table seems to drop in the migration - I can't see that I've mis-run anything. I'm not sure what this table is and it doesn't seem to exist in our Blazer schema. 

Should I keep the schema file like this or should I unstage this change (the table getting removed)?

patrick-laa: Ok good spot. So as context: Our database schema is _defined_ by migrations and _described_ by the schema. In practice, that means that every time you run `rails db:migrate`, the migrations get applied to your local database, and then Rails looks at the state of your database afterwards and describes what it sees in the schema file.

If you look through the migrations, you'll see we don't have one that creates an `active_admin_comments` table. So according to the definition of our database schema, this table doesn't exist. Every time you run migrations locally, you end up with a local database that doesn't have this table, so the schema file is updated without that table. And the schema file as it stands, with this table removed, is "correct".

So, how did this get in in the first place? Well, ActiveAdmin is an alternative to RailsAdmin that I played around with when first building the admin CMS. My guess is that I set up ActiveAdmin locally and ran migrations to set up a table that it used, then when I realised it didn't work I deleted the ActiveAdmin migrations without reverting them, and didn't notice that I still had this table defined in my local database, meaning that every time I ran any migration, the table would appear in the schema file (because it's next to a bunch of other auto-generated tables like "active_storage_attachments" etc). My bad, but because the schema file is largely for informational purposes only, it didn't have any knock-on consequences, which is why it wasn't noticed until someone _else_ started making PRs involving schema changes.

app/views/layouts/_feedback.html.slim
```diff
@@ -0,0 +1,56 @@
...
+        h2.govuk-heading-s class="govuk-!-margin-bottom-0 govuk-!-margin-right-2"
+          = t("feedback.satisfaction.thank_you_message")
+        p.govuk-body-m class="govuk-!-margin-bottom-0 govuk-!-margin-right-6"
+          = t("feedback.satisfaction.feedback_link_html")
+- else
+  .gem-c-feedback__prompt data-module="freetext-feedback-area"
+    .govuk-width-container data-module="prompt-area"
+      .gem-c-feedback__prompt-content
+        .gem-c-feedback__prompt-questions--something-is-wrong
+          button.gem-c-feedback__prompt-link data-freetext-feedback="query"
```
patrick-laa, line 31: "query" isn't the clearest name to me. In technical terms, we have SQL queries and querystrings. In common parlance, a query is a question. But this button doesn't display a question, it displays a statement, and in particular I think it's a _prompt_ (it's in the prompt area after all).

config/locales/en.yml
```diff
@@ -2356,3 +2354,16 @@ en:
...
+     button: Give feedback on this page
+     heading: What is your feedback on this page?
+     heading_hint: Don't include personal information.
+     button_send: Send
+     button_cancel: Cancel
+     thank_you_message: Thank you for your feedback
+   satisfaction:
+     question: Were you satisfied with this service?
+     thank_you_message: Thank you for your feedback
+     feedback_link_html: <a href="https://eu.surveymonkey.com/r/ccqpublicbeta" target="_blank">Tell us more in our 2 minute survey (opens in new tab)</a>
```
patrick-laa, line 2368: We're trying to add `rel="noopener noreferrer"` to all external links. The linter picks up on this for ones in ruby code, but doesn't spot them automatically in en.yml.

spec/controllers/feedbacks_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
+require "rails_helper"
+
+RSpec.describe FeedbacksController, type: :controller do
+  describe "#create" do
+    before do
+      allow(controller).to receive(:level_of_help).and_return("controlled")
```
patrick-laa, line 6: Rather than fake this, I think it's worth creating a session and passing an assessment_code param in your tests.

spec/controllers/feedbacks_controller_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    it "saves freetext feedback correctly" do
+      post :create, params: { type: "freetext", text: "foo", page: "bar" }
+      expect(FreetextFeedback.find_by(text: "foo", page: "bar", level_of_help: "controlled")).not_to be_nil
+    end
+
+    it "saves satisfaction feedback correctly" do
+      allow(controller).to receive(:outcome).and_return("eligible")
```
patrick-laa, line 15: Likewise here, we're not really checking that our logic to retrieve data from the session is working correctly. Better to have some session data and an assessment code param.

spec/models/freetext_feedback_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe FreetextFeedback, type: :model do
```
patrick-laa, line 3: I'm not convinced these model specs add any long-term value. We're basically checking that validations validate.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+        )
+      else
+        FreetextFeedback.create!(
+          text: @form.text,
+          page: @form.page,
+          level_of_help:,
+        )
+      end
+    else
+      track_validation_error
```
patrick-laa, line 25: This is kinda an abuse of this method - we don't want to create a "validation_message" analytics event (which is what this method does), because if this line of code gets hit the user _doesn't_ see a validation message. By all means save an analytics event, but I think it should be a custom one.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
```
patrick-laa, line 27: If this succeeds, the server should return an appropriate status code. Likewise if it fails. (Practically speaking this means adding something like an explicit `head :created` or `head :unprocessable_entity`)

app/views/layouts/_feedback.html.slim
```diff
@@ -0,0 +1,56 @@
...
+        = t("feedback.freetext.heading")
+        p.govuk-body-m class="govuk-!-margin-bottom-4"
+          = t("feedback.freetext.heading_hint")
+          = form_with(url: feedbacks_path(assessment_code: params[:assessment_code],
+                                          type: "freetext",
+                                          page: @step),
+                                          local: false) do |form|
+            = form.govuk_text_area :text,
+                                    label: nil,
+                                    hint: nil,
```
patrick-laa, line 45: If we add `required: true` to this I think we can prevent the form from being submitted if no text is entered. What we need to do then is maybe _not_ hide this section if the field is blank (the browser should show a default indicator, which I think is fine).

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,50 @@
+class FeedbacksController < ApplicationController
+  FEEDBACK_TYPES = %w[freetext satisfaction].freeze
+
+  def create
+    if params[:type].nil? || !FEEDBACK_TYPES.include?(params[:type])
```
patrick-laa, line 5: Technically the first half of this if statement is redundant. If params[:type] is nil, then FEEDBACK_TYPES will not include it.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+      raise "Feedback type needs to be specified"
```
patrick-laa, line 6: I'm not sure about this behaviour. If something tries to submit an invalid payload to this controller, I don't think we should return a 500 status code and trigger a sentry notification etc, unless we're sure this is a sign that our code is broken somewhere. Otherwise, I think returning a 422 or 400 feels more appropriate

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    @form = params[:type].include?("satisfaction") ? satisfaction_model : freetext_model
```
patrick-laa, line 9: I'm not sure `include?` is the ideal check here - we're checking to see if params[:type] _equals_ satisfaction. It works out as practically the same thing, because any string that includes satisfaction but doesn't equal it would raise an error on the previous line, but for clarity, an equals might be better.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  def freetext_model
+    FreetextFeedback.new(params.permit(:text, :page))
+  end
+
+  def level_of_help
+    session_data["level_of_help"]
+  end
+
+  def outcome
+    session_data["api_response"].dig("result_summary", "overall_result")["result"]
```
patrick-laa, line 48: instead of `x[a].dig(b,c)[d]` you might consider `x.dig(a,b,c,d)` for clarity.

config/locales/en.yml
```diff
@@ -2356,3 +2354,16 @@ en:
           - "(if known) 'We estimate [X] checks were affected by this issue'"
         content: Resolution text
         resolve: Resolve problem
+  feedback:
+   freetext:
```
patrick-laa, line 2358: This indentation looks fractionally off - I think you need to indent this line and the stuff below it by one more space.

spec/system/feedback_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    end
+
+    context "when on the results page" do
+      it "I can successfully submit satisfaction feedback" do
+        start_assessment
+        fill_in_forms_until(:check_answers)
+        click_on "Submit"
+        expect(page).to have_content("Were you satisfied with this service?")
+        click_on "Yes"
+        expect(page).to have_content("Thank you for your feedback")
```
patrick-laa, line 25: It might be a good idea to verify that a record has been saved in our database here. Either by checking for SatisfactionFeedback objects or by in the test, logging in to the admin back end and verifying that they appear on screen.

app/javascript/feedback.js
```diff
@@ -0,0 +1,64 @@
+const initFeedback = () => {
+  const query = document.querySelector('[data-freetext-feedback="query"]');
+  const cancel = document.querySelector('[data-freetext-feedback="cancel"]');
+  const send  = document.querySelector('[data-freetext-feedback="send"]');
+  const yes_button = document.querySelector('#yesButton');
+  const no_button = document.querySelector('#noButton');
+
+  if (query) {
```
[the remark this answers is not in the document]
MazOneTwoOne: I like js [null safe operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) - that is handy.

app/javascript/feedback.js
```diff
@@ -0,0 +1,64 @@
...
+  document.querySelector('[data-module="freetext-area"]').hidden = false;
+  document.querySelector('[data-module="message-area"]').hidden = true;
+  document.querySelector('[data-module="freetext-feedback-area"]').style.backgroundColor = "white";
+}
+
+function freetextAreaHide() {
+  document.querySelector('[data-module="prompt-area"]').hidden = false;
+  document.querySelector('[data-module="freetext-area"]').hidden = true;
+  document.querySelector('[data-module="message-area"]').hidden = true;
+  document.querySelector('[data-module="freetext-feedback-area"]').style.backgroundColor = "#f3f2f1";
```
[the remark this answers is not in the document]
MazOneTwoOne: Yeah - that makes more sense. Was so focussed on getting CSS classes to work, that didn't see the simpler way of doing it.

app/javascript/feedback.js
```diff
@@ -0,0 +1,64 @@
...
+
+  if (no_button) {
+    no_button.addEventListener('click', () => {
+     linkAreaShow();
+    })
+  }
+};
+
+function freetextAreaShow() {
+  document.querySelector('[data-module="prompt-area"]').hidden = true;
```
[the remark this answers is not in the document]
MazOneTwoOne: This sent me for a loop :(

I kept getting _uncaught typeerror: cannot set property_ and realised that the query selector was trying to set something that could be could be `null` or `undefined` - I ended up with this

```
function showSection(sectionArea) {
  ['freetext', 'message', 'prompt', 'link', 'question'].forEach((section) => {
    const sectionElement = document.querySelector(`[data-feedback-section="${section}-area"]`);

     if (sectionElement) {
      sectionElement.hidden = (section !== sectionArea);
     }
  });
}

```

app/views/layouts/_feedback.html.slim
```diff
@@ -0,0 +1,56 @@
...
+        .gem-c-feedback__prompt-questions--something-is-wrong
+          button.gem-c-feedback__prompt-link data-freetext-feedback="query"
+            = t("feedback.freetext.button")
+
+    .govuk-width-container data-module="freetext-area" hidden=true
+      .gem-c-feedback__form-heading
+        = t("feedback.freetext.heading")
+        p.govuk-body-m class="govuk-!-margin-bottom-4"
+          = t("feedback.freetext.heading_hint")
+          = form_with(url: feedbacks_path(assessment_code: params[:assessment_code],
```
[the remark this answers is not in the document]
MazOneTwoOne: @HettieS can we work on this together?

app/views/layouts/_feedback.html.slim
```diff
@@ -0,0 +1,56 @@
...
+                                          type: "freetext",
+                                          page: @step),
```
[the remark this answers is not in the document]
MazOneTwoOne: @HettieS another one to work together on.

app/views/layouts/_feedback.html.slim
MazOneTwoOne: @HettieS another one to work on together, if possible.

app/views/layouts/application.html.slim
```diff
@@ -60,4 +60,7 @@ html dir="ltr" lang="en-GB" class="govuk-template"
 
           = yield
 
+    - unless params[:controller] == "start"
```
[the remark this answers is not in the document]
MazOneTwoOne: We've all assumed different things of when to show the 'freetext' widget, so have asked the question on the ticket to confirm. Will do your approach once confirmed 😊.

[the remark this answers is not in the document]
MazOneTwoOne: Updated this - not sure how this `@freeform_feedback_step = "check_answers"` part, fits in?

db/schema.rb
HettieS: to chat about when pairing

app/views/layouts/_feedback.html.slim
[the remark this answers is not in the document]
HettieS: to chat about when pairing

app/views/layouts/application.html.slim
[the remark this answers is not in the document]
HettieS: `@freeform_feedback_step` is now plumbed in for "check_answers" but lets chat about this too to make sure we haven't missed any pages that might work differently i.e. updates and help

spec/controllers/feedbacks_controller_spec.rb
HettieS: to chat about when pairing

spec/controllers/feedbacks_controller_spec.rb
HettieS: to chat about when pairing

app/controllers/feedbacks_controller.rb
HettieS: to chat about when pairing

app/views/layouts/_feedback.html.slim
HettieS: to chat about when pairing

app/controllers/feedbacks_controller.rb
HettieS: to chat about when pairing

spec/system/feedback_spec.rb
HettieS: to chat about when pairing

app/views/layouts/_feedback.html.slim
MazOneTwoOne: Updated the JS.

app/javascript/feedback.js
```diff
@@ -0,0 +1,47 @@
...
+  });
+
+  cancel?.addEventListener('click', () => {
+    showSection("prompt");
+  })
+
+  send?.addEventListener('click', (e) => {
+    const freetextField = document.getElementById("text-field")
+
+    if (freetextField.value === "") {
```
patrick-laa, line 20: Minor bug here: if the user enters ` ` (a space, or several spaces) this validation here will pass, but the back-end `presence` validation will fail, and we'll get a Sentry notification.

A fix would be to check if the string, with all whitespace removed, is empty here. I.e. `freeTextField.value.replace(/\s+/g, '') === ""`

app/views/layouts/_feedback.html.slim
```diff
@@ -0,0 +1,58 @@
...
+                                        local: false) do |form|
+          = form.govuk_submit "No", class: "gem-c-feedback__prompt-link-small-button", id: "data-satisfaction-feedback-no"
+
+    .govuk-width-container data-feedback-section="link-area" hidden=true
+      span.gem-c-feedback__prompt-question-answer
+        h2.govuk-heading-s class="govuk-!-margin-bottom-0 govuk-!-margin-right-2"
+          = t("feedback.satisfaction.thank_you_message")
+        p.govuk-body-m class="govuk-!-margin-bottom-0 govuk-!-margin-right-6"
+          = t("feedback.satisfaction.feedback_link_html")
+- if @feedback == :freetext
```
patrick-laa, line 26: Fot clarity, maybe this could be `elsif`.

app/views/layouts/application.html.slim
```diff
@@ -60,4 +60,7 @@ html dir="ltr" lang="en-GB" class="govuk-template"
 
           = yield
 
+    - if @feedback
```
patrick-laa, line 63: We can remove this check, because `@feedback` is now _always_ set.

patrick-laa (approved):
Some very small comments, otherwise this looks good to me.

app/javascript/feedback.js
MazOneTwoOne: of course.

spec/controllers/feedbacks_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
+require "rails_helper"
+
+RSpec.describe FeedbacksController, type: :controller do
+  describe "#create" do
+    let(:session) do
+      { "03ac674216f3e15c761ee1a5e255f067953623c8b388b4459e13f978d7c846f4" =>
```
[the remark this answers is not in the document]
patrick-laa: More or less. Assessment ID is made by combining assessment code and cookies['SessionData']. So what we're saying is "1234" + blank => "03ac674...". So ideally maybe phrase as `This long string is the corresponding assessment_id when the assessment_code is "1234" and cookies["SessionData"] is blank.`

PR #1005, opened 2023-10-12
EL-1226: Flag up client-side errors in our system tests

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1226)

## What changed and why

Flag up client-side errors in our system tests
Same for warnings.
Also fix a broken image reference (because it was flagged up by our tests)

## Checklist


HettieS (approved):
This looks good to me, I tested this by checking out the branch and starting the Rails server twice in application.js - it triggered the rails-ujs error that we saw last week. I couldn't work out how to trigger the 'warning' behaviour whilst testing.

As this new config will only run in our system specs, I'm wondering if we should add something to the README for our system specs, to call out that they are the place to test javascript? But might not be necessary.

PR #1006, opened 2023-10-12
EL-1228: Add client-side Sentry monitoring

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1228)

## What changed and why

Put environment-specific Sentry keys in config, and use them to init Sentry client-side error reporting, adjusting our CSP to allow it.

## Checklist

### DIFF
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -5,8 +5,15 @@ if (sentryDsn) {
   Sentry.init({
     dsn: sentryDsn,
     integrations: [new Sentry.BrowserTracing(), new Sentry.Replay()],
-    tracesSampleRate: 1.0,
+    // All errors are captured
+    sampleRate: 1,
+    // 10% of non-erroring sessions are captured for performance monitoring
+    tracesSampleRate: 0.1,
+
+    // 10% of sessions are replayable
     replaysSessionSampleRate: 0.1,
+
+    // All erroring sessions are replayable
     replaysOnErrorSampleRate: 1.0,
   });
 }
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -7,14 +7,13 @@ if (sentryDsn) {
     integrations: [new Sentry.BrowserTracing(), new Sentry.Replay()],
     // All errors are captured
     sampleRate: 1,
-    // 10% of non-erroring sessions are captured for performance monitoring
-    tracesSampleRate: 0.1,
-
-    // 10% of sessions are replayable
-    replaysSessionSampleRate: 0.1,
-
     // All erroring sessions are replayable
     replaysOnErrorSampleRate: 1.0,
+
+    // No non-erroring sessions are captured for performance monitoring
+    tracesSampleRate: 0,
+    // No non-erroring sessions are replayable
+    replaysSessionSampleRate: 0,
   });
 }
 
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -46,5 +46,3 @@ if (!window._rails_loaded) {
 initAll();
 initResults();
 initFeedback();
-
-
--- a/app/javascript/application.js
+++ b/app/javascript/application.js
@@ -1,6 +1,7 @@
 import { initAll } from "govuk-frontend";
 import initResults from "./results";
 import Rails from '@rails/ujs';
+import * as Sentry from "@sentry/browser";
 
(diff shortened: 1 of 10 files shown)


app/javascript/application.js
```diff
@@ -1,3 +1,16 @@
+import * as Sentry from "@sentry/browser";
+
+const sentryDsn = document.querySelector("body").dataset.sentryDsn;
+if (sentryDsn) {
+  Sentry.init({
+    dsn: sentryDsn,
+    integrations: [new Sentry.BrowserTracing(), new Sentry.Replay()],
+    tracesSampleRate: 1.0,
+    replaysSessionSampleRate: 0.1,
```
HettieS, line 9: Just wondering if we need to sample any sessions without errors?

app/views/layouts/application.html.slim
```diff
@@ -29,7 +29,7 @@ html dir="ltr" lang="en-GB" class="govuk-template"
 
     = stylesheet_link_tag "application", "data-turbo-track": "reload"
     = javascript_include_tag "application", "data-turbo-track": "reload", defer: true, nonce: true
-  body class="govuk-template__body"
+  body class="govuk-template__body" data-sentry-dsn=ENV["SENTRY_CLIENT_SIDE_DSN"]
```
HettieS, line 32: What's the reason that we are creating a new variable `SENTRY_CLIENT_SIDE_DSN` instead of using the existing `SENTRY_DSN`?

app/javascript/application.js
```diff
@@ -1,3 +1,16 @@
+import * as Sentry from "@sentry/browser";
+
+const sentryDsn = document.querySelector("body").dataset.sentryDsn;
+if (sentryDsn) {
```
HettieS, line 4: We may not need to but I wonder if we need to check for the Sentry feature flag - was just testing this locally, I had set the `SENTRY_CLIENT_SIDE_DSN` in my local dev settings, but hadn't enabled the feature flag. The alert went straight through to Sentry with the flag not enabled. Maybe that's fine?

app/views/layouts/application.html.slim
HettieS: I'm now wondering if this is something to do with my above comment re. feature flag!

HettieS (commented):
Quick question re feature flag but this does look good!

app/javascript/application.js
patrick-laa: So as I understand it the tracesSampleRate gives us that, which is mildly useful for performance tracking.

app/views/layouts/application.html.slim
patrick-laa: So these are 2 different Sentry keys, one for server-side and one for client-side. I guess the reason is one is a private secret and one is public (because for it to be available to the JS client-side it needs to be available in the source of the page).

app/javascript/application.js
HettieS: I'm sorry, my question was super unclear - I think what I meant to ask is, why do we need to sample sessions without errors? And it was looking at `replaysSessionSampleRate: 0.1` that made me wonder that. I was thinking that config is going to generate a replay for 10% of our checks, is that something we want?

patrick-laa: I have tweaked this and added comments

app/javascript/application.js
patrick-laa: Fixed.

app/javascript/application.js
patrick-laa: Ah, OK, sorry, basically errorless session monitoring is something that Sentry throws in "for free". On the back end we use it to monitor the speed of our servers and which endpoints get called the most. On the front end it adds a little more data that has the opportunity to show is if something starts going wrong. But it is maybe more than we actually need, and potentially pushes into some murky privacy water. So maybe best to remove everything except error cases.

app/views/layouts/application.html.slim
patrick-laa: As discussed on Slack, as you correctly pointed out, these are _not_ two separate keys - Sentry provides the same value for client-side and server-side DSNs. So we don't need to store it in 2 variables.

HettieS (approved):
Looks good - is there a follow up action to remove the sentry DSN from `kube-secrets` ?

patrick-laa:
@HettieS yes indeed - once this is merged and deployed we should do exactly that.

PR #1012, opened 2023-10-16
EL-1213: Add a migration script for moving redis data across namespaces

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1213)

## What changed and why

Add a script for migration of redis data from one namespace to another.

## Checklist


lib/scripts/migrate-redis.sh
```diff
@@ -0,0 +1,166 @@
...
+
+  return 0
+}
+
+# Run the RIOT tool that will push data from old to new in batches
+run_riot() {
+  echo "10/11 Running riot..."
+  riot -h localhost -p $OLD_REDIS_LOCAL_PORT --pass $OLD_REDIS_PASSWORD --tls --tls-verify NONE \
+      replicate \
+      -h localhost -p $NEW_REDIS_LOCAL_PORT --pass $NEW_REDIS_PASSWORD --tls --tls-verify NONE --batch 10000 \
```
HettieS, line 121: looks good, I'm not too familiar with Riot - just so I'm understanding, if the migration fails here midway through, the redis pods get deleted anyway. So we wouldn't end up in a situation with half of the redis data imported. Is that right?

patrick-laa: Ah, not exactly. What get deleted are the 'port forwarding' pods - the things that allow us to access redis from our local machines.

So in the worst case scenario, riot runs and gets half of the redis data migrated, and then fails out. Everything gets tidied up, the port forwarding pods get deleted, and when the dust clears the new namespace has half of the data from the old namespace. I can't see a riot option that does the whole transaction atomically, so if this happens, as it could, our recourse would have to be to use the `redis-cli` to do a FLUSHALL on the new namespace and try again or similar. Which isn't ideal, but I can't see a better way, and I think it's manageable if it comes to it.

PR #1014, opened 2023-10-17
EL-1217: Dual running new UAT namespace

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1217)

## What changed and why

This PR gets dual deployments running to old and new UAT namespaces. Alongside changes here, changes are made in Circle CI Config and `kube-secrets`. These changes are detailed on the ticket. The same approach should be able to be used for staging and prod. 

This PR also makes changes to _env.tpl and the deploy script which should set the groundwork for dual running for other environments.

We are temporarily referring the the new namespace UAT environment as `ccq-uat` so that we can deploy successfully to a different URL.

## Checklist

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -7,7 +7,7 @@ ENVIRONMENT=$1
 
 deploy_branch() {
 # Set the deployment host, this will add the prefix of the branch name e.g el-257-deploy-with-circleci or just main
-  RELEASE_HOST="$BRANCH_RELEASE_NAME-check-client-qualifies-legal-aid-$ENVIRONMENT.cloud-platform.service.justice.gov.uk"
+  RELEASE_HOST="$BRANCH_RELEASE_NAME-$BRANCH_HOST_SUFFIX.cloud-platform.service.justice.gov.uk"
 # Set the ingress name, needs release name, namespace and -green suffix
   IDENTIFIER="$BRANCH_RELEASE_NAME-laa-estimate-eligibility-$K8S_NAMESPACE-green"
   echo "Deploying CIRCLE_SHA1: $CIRCLE_SHA1 under release name: '$BRANCH_RELEASE_NAME'..."
--- a/bin/deploy
+++ b/bin/deploy
@@ -7,15 +7,15 @@ ENVIRONMENT=$1
 
 deploy_branch() {
 # Set the deployment host, this will add the prefix of the branch name e.g el-257-deploy-with-circleci or just main
-  RELEASE_HOST="$BRANCH_RELEASE_NAME-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk"
+  RELEASE_HOST="$BRANCH_RELEASE_NAME-check-client-qualifies-legal-aid-$ENVIRONMENT.cloud-platform.service.justice.gov.uk"
 # Set the ingress name, needs release name, namespace and -green suffix
-  IDENTIFIER="$BRANCH_RELEASE_NAME-laa-estimate-eligibility-laa-estimate-financial-eligibility-for-legal-aid-uat-green"
+  IDENTIFIER="$BRANCH_RELEASE_NAME-laa-estimate-eligibility-$K8S_NAMESPACE-green"
   echo "Deploying CIRCLE_SHA1: $CIRCLE_SHA1 under release name: '$BRANCH_RELEASE_NAME'..."
 
   helm upgrade $BRANCH_RELEASE_NAME ./helm_deploy/laa-estimate-eligibility/. \
                 --install --wait \
                 --namespace=${K8S_NAMESPACE} \
-                --values ./helm_deploy/laa-estimate-eligibility/values/uat.yaml \
+                --values ./helm_deploy/laa-estimate-eligibility/values/$ENVIRONMENT.yaml \
                 --set image.repository="$ECR_ENDPOINT" \
                 --set image.tag="branch-$CIRCLE_SHA1" \
                 --set ingress.annotations."external-dns\.alpha\.kubernetes\.io/set-identifier"="$IDENTIFIER" \
--- a/helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
+++ b/helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
@@ -1,4 +1,4 @@
-# Default values for laa-estimate-eligibility-uat.
+# Default values for laa-check-client-qualifies-uat.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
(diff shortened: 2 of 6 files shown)


HettieS:
I ran the redis script we wrote in https://github.com/ministryofjustice/laa-estimate-financial-eligibility-for-legal-aid/pull/1012 as a way of testing the import of data across UATs. 

There were some small issues importing old namespace UAT redis into new namespace UAT redis, I tried this three times and each time, the migration failed due to 1 or 2 'missing' keys. But we think this is because new keys were being generated as the migration was running.

The verify this, we amended the script slightly so that we could import new namespace UAT redis into new namespace UAT staging (we knew no one was using these at the time). The script imported successfully with no failures. This implies the first few fails were due to new keys being created at time of migration.

As we will put the site in maintenance mode when we run this in production, we think that should eliminate this type of failure, and it should run successfully.

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
```diff
@@ -0,0 +1,112 @@
...
+  type: ClusterIP
+  port: 80
+
+securityContext:
+  runAsUser: 1000
+
+ingress:
+  hosts:
+#    - host: main-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk
+    - host: main-ccq-uat.cloud-platform.service.justice.gov.uk
```
patrick-laa, line 44: My assumption is that we can't use the actual main UAT URL as the host name here because it's already in use in the old namespace and something will break if we declare the same host for both. But did we actually test that? If the weight is set to 0 here, would it actually _allow_ us to have both namespaces declare the same host?

I'm asking because we need to solve this problem at some point, and I've got solving this problem earmarked for EL-1214 which on JIRA is marked as blocking this ticket. If you've learned anything about this as part of this ticket, I'm eager to see if we can use it to make progress on EL-1214.

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
```diff
@@ -0,0 +1,112 @@
...
+  # Similarly this is available in the page source
+  dsn: https://04ec9fc34a2146138f8a2da2d243a718@o345774.ingest.sentry.io/6747538
+
+app:
+  primaryHost: ""
+  seedAdmins: patrick.gleeson@digital.justice.gov.uk,michael.blatherwick@digital.justice.gov.uk,william.clarke@digital.justice.gov.uk,masum.khan@digital.justice.gov.uk,hettie.street@digital.justice.gov.uk
+
+google:
+  oauthClientId: 276075615370-ncqvmdi84jruv1jggfg4bu8nlvk1drj0.apps.googleusercontent.com
+  oauthRedirectUri: https://main-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk/auth/subdomain_redirect
```
patrick-laa, line 109: This should probably be set to `main-ccq-uat.cloud-platform.service.justice.gov.uk`, since the other "main" host might go away with the other namespace. But if so, we need to whitelist `main-ccq-uat...` in the Google Developer Console as a valid redirect URI for OAuth.

bin/deploy
```diff
@@ -7,15 +7,15 @@ ENVIRONMENT=$1
 
 deploy_branch() {
 # Set the deployment host, this will add the prefix of the branch name e.g el-257-deploy-with-circleci or just main
-  RELEASE_HOST="$BRANCH_RELEASE_NAME-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk"
+  RELEASE_HOST="$BRANCH_RELEASE_NAME-check-client-qualifies-legal-aid-$ENVIRONMENT.cloud-platform.service.justice.gov.uk"
```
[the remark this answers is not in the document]
patrick-laa: Ooh, could we put a BRANCH_HOST_SUFFIX in the two context we have, and in the `laa-estimate-eligibility-uat` context it could be `check-client-qualifies-legal-aid-uat`, but in the `laa-check-client-qualifies-uat` one it could be `ccq-uat` maybe? THat way this line could be:

```bash
RELEASE_HOST="$BRANCH_RELEASE_NAME-$BRANCH_HOST_SUFFIX.cloud-platform.service.justice.gov.uk"
```

And we'd end up with more concise UAT URLs?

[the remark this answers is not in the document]
willc-work: I am not even sure this $ENVIRONMENT is required, this script always deploys to UAT so it should never have a production or staging.

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
```diff
@@ -0,0 +1,112 @@
...
+  runAsUser: 1000
+
+ingress:
+  hosts:
+#    - host: main-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk
+    - host: main-ccq-uat.cloud-platform.service.justice.gov.uk
+      secret: false
+  annotations:
+    external-dns.alpha.kubernetes.io/set-identifier: "check-client-qualifies-laa-estimate-eligibility-laa-check-client-qualifies-uat-green"
+    # TODO: the weight is set to 0 for now but will need to be reverted to 100 at some point
```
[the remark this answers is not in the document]
willc-work: just a fail safe we agreed on, when host is changed this will already be at zero.

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
willc-work: no we did not test it because testing it could only happen after merging the PR so we are playing it safe by changing it for now. Once we can dual deploy we can run a branch to test exactly how it works, as this will apply to UAT/STG and PROD envs

bin/deploy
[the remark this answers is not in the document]
HettieS: Yes that is a nice idea! We should try and do that. I can take a look!

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
willc-work: we were not 100% sure of the correct approach for this, will make those changes.

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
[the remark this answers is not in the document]
HettieS: This was all quite trial and error - @willc-work and I had a good chat with @exonian about ingress settings. I think because we had managed to get the deployment working in this way, we stuck with it. I think 0 in this particular context might have the same affect as setting to 100? But you're right, we have changed the identifier, ingress and host so perhaps we don't need this. I will defer to Will and Michael!

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
HettieS: Happy to make this change although I'm not sure how to do the whitelisting, if someone could help with that @patrick-laa @willc-work

patrick-laa: Sure @HettieS I think this link will get you to the right screen: https://console.cloud.google.com/apis/credentials/oauthclient/276075615370-ncqvmdi84jruv1jggfg4bu8nlvk1drj0.apps.googleusercontent.com?orgonly=true&project=ccq-admin&supportedpurview=organizationId - all you need to do is add a new entry to the list of 'Authorised redirect URIs' and save.

HettieS: thank you! this should now be done

willc-work (approved):
Approving PR changes

PR #1017, opened 2023-10-19
EL-1229: Amalgamate dependabot PRs

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1229)

## What changed and why

Adapt a public workflow to combine PRs
(What I changed was to add automatic scheduling of the workflow to run on a Monday morning. Because of that, I needed the script to use default values for variables that would be specified when the workflow was run manually, so I added a few `|| 'some-string'` to a few places.

## Checklist


.github/workflows/combine-prs.yml
```diff
@@ -0,0 +1,150 @@
...
+                mergeFailedPRs.push(prString);
+              }
+            }
+
+            console.log('Creating combined PR');
+            const combinedPRsString = combinedPRs.join('\n');
+            let body = '✅ This PR was created by the Combine PRs workflow by combining the following PRs:\n' + combinedPRsString;
+            if(mergeFailedPRs.length > 0) {
+              const mergeFailedPRsString = mergeFailedPRs.join('\n');
+              body += '\n\n⚠️ The following PRs were left out due to merge conflicts:\n' + mergeFailedPRsString
```
MazOneTwoOne, line 141: Is that an emoji/wingding?

.github/workflows/combine-prs.yml
```diff
@@ -0,0 +1,150 @@
...
+
+# Controls when the action will run - in this case triggered manually
+on:
+  workflow_dispatch:
+    inputs:
+      branchPrefix:
+        description: 'Branch prefix to find combinable PRs based on'
+        required: true
+        default: 'dependabot'
+      combineBranchName:
```
MazOneTwoOne, line 12: question - so we want to have a PR which has every dependabot update, regardless of whether is passing tests (I see we have omitted `mustBeGreen`)? I guess we will still have the existing dependabot PRs so we can see where to focus our search for the combined PR???

MazOneTwoOne (approved):
It's copied over as suggested in the other resource and can see where and why you've made changes.

.github/workflows/combine-prs.yml
patrick-laa: We don't have mustBeGreen as a variable, but what we've removed from the code is the path where mustBeGreen was set to false. Now it's hard-coded that the branch must be green to be included in the amalgamation, there's no way of bypassing it. The reason being that it was really hard to have a boolean variable that defaulted to true when the workflow was called by the scheduler (in which case this `inputs` section is ignored), and it seemed to me unlikely we'd ever want to combine failing PRs.

.github/workflows/combine-prs.yml
patrick-laa: It's an emoji that seems only to render as a colourful emoji in some contexts.

PR #1020, opened 2023-10-19
EL-1214 UAT Main shared deploy host

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1214)

using a shared host for main on UAT so we can watch what happens when we send traffic from one to the other.
We can set the weight using the kubectl apply command directly in the namespaces (alternative is to update both of the values files).

## Checklist


patrick-laa:
Could we test this before merge by changing the _branch_ deploy host to be identical in both namespaces?

PR #1052, opened 2023-11-01
EL-1236: pending answers in change answer loops

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1236)

## What changed and why

* Renamed CheckAnswersController to ChangeAnswersController (because "Check answers" is how we describe the check answers screen, and having it also mean where you go _from_ that screen was confusing me).
* In ChangeAnswersController, use temporary copy of the session data that we keep in a key called 'pending', and only override the main session data with the temporary copy when we complete a loop 

## Guidance to review

The session_data method in ChangeAnswersController overrides the default method in ApplicationController
## Checklist

### DIFF
--- a/app/controllers/change_answers_controller.rb
+++ b/app/controllers/change_answers_controller.rb
@@ -1,4 +1,4 @@
-class CheckAnswersController < QuestionFlowController
+class ChangeAnswersController < QuestionFlowController
   before_action :set_back_behaviour
 
   def update
@@ -31,13 +31,16 @@ private
   # stored in a section of the session called 'pending'.
   def session_data
     data = super
-    if data[:pending] && !params[:begin_editing]
-      data[:pending]
-    else
+
+    # Set up a fresh copy of the answers into the pending section if it's blank or we're
+    # starting a new loop
+    if !data[:pending] || params[:begin_editing]
       pending = data.dup
       pending[:pending] = nil
       session[assessment_id][:pending] = pending
       pending
+    else
+      data[:pending]
     end
   end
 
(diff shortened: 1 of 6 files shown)


app/controllers/change_answers_controller.rb
```diff
@@ -25,6 +27,23 @@ def update
...
+  # While we're in a 'change answers loop', we want to be working with a temporary copy of the answers
+  # stored in a section of the session called 'pending'.
+  def session_data
+    data = super
+
+    # Set up a fresh copy of the answers into the pending section if it's blank or we're
+    # starting a new loop
+    if !data[:pending] || params[:begin_editing]
+      pending = data.dup
+      pending[:pending] = nil
```
MazOneTwoOne, line 39: Question - So I'm guessing that CFE doesn't care that we are also sending it the extra key value pair `pending => nil`?

MazOneTwoOne (approved):
Had a look at this with Hettie, as it was confusing at first (in particular the object inheritance & guarding against creating a `pending` key recursively).

Just had the one question.

app/controllers/change_answers_controller.rb
patrick-laa: Ah, but we're not: we don't send the session data straight to CFE. Instead, the CfeParamBuilders classes pick specific parts of the session data out and transform them into the format CFE wants. So the 'pending' key never gets near CFE.

MazOneTwoOne: Oh yeah. Neat.

PR #1060, opened 2023-11-02
EL-1214 switch staging ingress hosts

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1214)

Use the staging domain names in the new namespace

## Checklist

### DIFF
--- a/helm_deploy/laa-estimate-eligibility/values/ccq-staging.yaml
+++ b/helm_deploy/laa-estimate-eligibility/values/ccq-staging.yaml
@@ -30,15 +30,13 @@ service:
 
 ingress:
   hosts:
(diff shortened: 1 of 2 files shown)


patrick-laa (approved):
Approving so you can test it. My guess is that if anything were to cause problems it would be that CircleCI deploys to both staging environments simultaneously, so each attempted release might say "You can't use that hostname, it's already in use". It might be we have to first remove host 1 from namespace A and deploy to A, then add host 1 to namespace B and remove host 2 from namespace B, then deploy to B, then add host 2 to namespace A and deploy to namespace A again, if you see what I mean. But let's see!

PR #1088, opened 2023-11-09
EL-1256: Results page - controlled eligible result

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1256)

## What changed and why

- Content changes & reordering of CTAs on the results page for controlled work checks returning an eligible result. 
- No change to the page, when a controlled eligible asylum result
- Content changes behind feature flag

## Guidance to review

- N/A

## Checklist

### DIFF
--- a/app/views/results/show.html.slim
+++ b/app/views/results/show.html.slim
@@ -19,6 +19,26 @@
       - %i[gross_income disposable_income capital].each
         = render("summary", section: _1, links: true) if @model.calculated?(_1)
 
+  - if FeatureFlags.enabled?(:end_of_journey, @check.session_data) && @model.level_of_help == "controlled" && @model.decision == "eligible"
+    h2.govuk-heading-m = t(".save_results")
+    p.govuk-body = t(".save_results_intro")
+    ul.govuk-list.govuk-list--bullet
+      - t(".save_results_bullets").each
+        li = _1
+
+    .govuk-button-group
+      = link_to t(".save_as_pdf"),
+            download_result_path(params[:assessment_code]),
+            target: "_blank", rel: "noopener", class: "govuk-button govuk-button--secondary"
+
+    h2.govuk-heading-m = t(".complete_a_cw_form")
+    p.govuk-body = t(".complete_a_cw_form_paragraph")
+    h2.govuk-heading-s = t(".download_a_cw_form_header")
+
+    = link_to t("results.show.continue_to_cw"),
+              controlled_work_document_selection_path(assessment_code: params[:assessment_code]),
+              class: "govuk-button", data: { module: "govuk-button" }
+
   - if @model.level_of_help == "certificated"
     h2.govuk-heading-m = t(".what_happens_next")
     - if @check.immigration_or_asylum?
@@ -60,6 +80,7 @@
         - accordion.with_section(heading_text: t(".capital_calculation"))
           = render "capital_table", caption_size: "m"
 
+- unless FeatureFlags.enabled?(:end_of_journey, @check.session_data)
   .govuk-button-group
     / For eligible controlled checks there is no primary CTA in this group, as it appears below.
     / For other checks, 'Save as PDF' is the primary CTA.
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -47,6 +47,6 @@ private
   end
 
   def specify_feedback_widget
-    @feedback = :satisfaction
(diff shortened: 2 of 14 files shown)


patrick-laa:
@MazOneTwoOne because this branches off EL-1257 but is pointing at `main`, it's currently showing all changes from both tickets. If you `Edit` the PR (top right) to change the `base` to the EL-1257 branch, hopefully we'll be able to see just your changes without mine here.

spec/system/feedback_spec.rb
```diff
@@ -26,6 +26,22 @@
...
+        fill_in_forms_until(:check_answers)
+        click_on "Submit"
+        expect(page).to have_content("Give feedback on this page")
+        click_on "Give feedback on this page"
+        fill_in "freetext-input-field", with: "some feedback!"
+        click_on "Send"
+        expect(page).to have_content("Thank you for your feedback")
+
+        # Could not get this line of the test to work
+        # expect(FreetextFeedback.find_by(text: "some feedback!", page: "results_show", level_of_help: "certificated")).not_to be_nil
```
patrick-laa, line 41: I suspect it's because the `page` is `show_results` not `results_show` - the default format is action name first then controller name after.

app/views/results/show.html.slim
```diff
@@ -1,3 +1,4 @@
+- end_of_journey = FeatureFlags.enabled?(:end_of_journey, @check.session_data)
```
patrick-laa, line 1: The logic on this page gets quite complicated, so to keep it clear we want to keep our variables clearly named. An `end_of_journey` boolean suggests it represents whether this is the end of the journey. But this variable actually represents whether the flag is enabled. Can we rename to `end_of_journey_flag` or similar?

app/views/results/show.html.slim
```diff
@@ -19,6 +20,26 @@
       - %i[gross_income disposable_income capital].each
         = render("summary", section: _1, links: true) if @model.calculated?(_1)
 
+  - if end_of_journey && @model.level_of_help == "controlled" && @model.decision == "eligible" && !@check.immigration_or_asylum?
```
patrick-laa, line 23: I think `!@check.immigration_or_asylum?` is wrong here. The discussion on Jira is about the special case where the applicant receives asylum support, but the code is just looking at whether this is an immigration or asylum case.

app/views/results/show.html.slim
```diff
@@ -60,20 +81,22 @@
...
-    / For eligible controlled checks there is no primary CTA in this group, as it appears below.
-    / For other checks, 'Save as PDF' is the primary CTA.
-    - if @model.level_of_help == "controlled" && @model.decision == "eligible"
-      = link_to t(".save_as_pdf"),
-                download_result_path(params[:assessment_code]),
-                target: "_blank", rel: "noopener", class: "govuk-button govuk-button--secondary"
-    - else
-      = link_to t(".save_as_pdf"), download_result_path(params[:assessment_code]), target: "_blank", rel: "noopener", class: "govuk-button"
-      = link_to t(".start_another_check"), new_check_path, class: "govuk-button govuk-button--secondary", data: { module: "govuk-button" }
+  - unless end_of_journey && !@check.immigration_or_asylum?
```
patrick-laa, line 84: A few issues with this. First, the previous (possibly before your time) agreement in the team has been that `unless` is harder to get your head around than `if` so we tend to only use it for simple clauses. To keep consistent, can we change `unless A && !B` to `if !A || B`.

Second, as above, I don't think `immigration_or_asylum?` is the test here, I think it's asylum support.

Third, as it stands this means if the flag is enabled, non-immigration _certificated_ checks don't get _any_ buttons on this screen. Or ineligible non-immigration controlled checks. You need the logic on line 84 to be the inverse of the logic on line 23. To make it clearer, you could maybe at the top of the page have a line saying `show_buttons_at_top = end_of_journey_flag && @check.controlled? && !@model.ineligible && !@check.asylum_support` and then on line 23 you say `if show_buttons_at_top` and on line 84 you say `if !show_buttons_at_top`

patrick-laa (commented):
I think a little bit of logic wrangling needed on the results page, but we're not far off.

app/views/results/show.html.slim
MazOneTwoOne: You summary about the 'third issue' was really helpful. Thank you.

patrick-laa (approved):

PR #1089, opened 2023-11-10
EL-1261: Spike early eligibility

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1261)

## What changed and why

This spike contains commits for a couple different implementations of this feature. Firstly we tried only sending gross income and disposable income early checks at the end of the sections, with a interrupting form/screen presented to the user.

We decided we wanted to try continuous checks, so this PR now implements that. We use a widget style warning component to let the user know once they have been pushed into ineligibility. They can then choose to proceed to check answers or continue with their check anyway.

For gross income, we send a request to CFE after every gross income screen and render widget accordingly. This is the case for client and partner gross income inputs.

For disposable income, we send the request at the end of the whole section the ineligibility works in such a way that we really need all the information before we can accurately assess. This is also a widget and works for client and partner.

We were going to look at creating continuous checks for capital but I was getting confused with property stuff, and seeing as it's right at the end of the service, and we think users are more likely to benefit from this ineligibility check happening at early check points, I have only implemented this early check to happen after the client asset screen and the partner asset screen. 

Change links are definitely flakey as it was hard to understand what the implementation should be. I have added some basic first step to get change links working but I think this needs a bit more thinking maybe. So change links are kind of broken-ish. 

## Guidance to review

See above!

## Checklist


app/controllers/forms_controller.rb
```diff
@@ -16,4 +17,24 @@ def update
...
+protected
+
+  def check_early_eligibility
+    if gross_income_details_valid?
+      session_data["api_result"] = CfeService.call(session_data, early_eligibility: true)
+      @early_eligibility_result = EarlyEligibility.new(session_data)
+    end
+  end
+
+  def gross_income_details_valid?
```
patrick-laa, line 30: The difficulty with this approach to seeing if we're ready for the early eligibility check (seeing if the relevant forms are valid) is that you need to build in a bunch of knowledge here about which are the relevant forms. It's not just income and other income, it's benefits, and it's potentially all these details for a partner too. You don't want to be re-implementing the logic from the Steps module.

I had thought you could do it by adding an attribute to some Steps::BlahSection classes that say 'when the last step in this section is completed (whatever that is for this particular check - you can work it out dynamically) do/don't do an early eligibility check'. That way in this controller you could simply call 'Steps::Helper.do_the_thinking_for_me(step, session_data)`. But the problem is that Steps::BlahSections don't actually map very well to gross income, disposable income and capital, and _can't_ because the gross income steps don't all follow one another sequentially in the flow, and nor do the disposable income steps.

I wonder if you could update Flow::Handler::STEPS to tag each step with things like `gross_income`, `disposable_income` etc. That way you could see for a given check "have I just completed the last gross income step?" by comparing which steps apply for the current check and what each step is tagged with.

app/controllers/forms_controller.rb
```diff
@@ -16,4 +17,24 @@ def update
       render "question_flow/#{step}"
     end
   end
+
+protected
+
+  def check_early_eligibility
+    if gross_income_details_valid?
+      session_data["api_result"] = CfeService.call(session_data, early_eligibility: true)
+      @early_eligibility_result = EarlyEligibility.new(session_data)
```
patrick-laa, line 26: This assignment won't do much, as it'll get wiped when we redirect the user to a new page. What you'll need is for the _next_ page to be rendered to know that it needs to show something about early eligibility. My thinking (based on the untested assumption that this early eligibility indicator will make more sense as a standalone screen than a widget above a form, as it contains such important information), was that you could add a new screen into the flow just like you'd add any other screen, called `post_gross_income_early_eligibility`. You'd add a new Steps::PostGrossIncomeEarlyEligibilitySection containing the logic about whether to skip that screen, and that logic would be "skip it if the early eligibility CFE result is 'eligible'".

The helpful thing about _that_ is that that new screen can then easily contain a form, like all the other screens. And that form can say "Would you like to skip to the end or not?", and we can store the user's answer in the session, just like for every other form, and then because that answer is stored in the session, we can use it to determine, in the other Steps::BlahSection classes, via Steps::Logic, whether to display or skip all the _subsequent_ screens in the flow. i.e. we can re-use our existing navigation system to handle the navigation options we give to the user after an early eligibility result.

app/services/cfe/regular_transactions_payload_service.rb
```diff
@@ -1,6 +1,7 @@
 module Cfe
   class RegularTransactionsPayloadService < BaseService
     def call
+      return if early_eligibility?
```
patrick-laa, line 4: I'm guessing you've added in this guard clause in because this method blows up otherwise? If so, I think this highlights an important problem: Cfe methods are designed to blow up if presented with an incomplete or invalid session, and the early eligibility result depends on calling Cfe with an invalid session. This is quite hard to unpick, e.g. you _can't_ just exit early out of this method because it's in this method that we add a bunch of gross income  data. This method packages up a bunch of income and outgoings data into the same place.

I think the general approach (pass in an 'early eligibility' flag, and when it's set to true, get Cfe to work with what it's got instead of blowing up), is probably the right one, but sadly I think it's going to be a case of doing more than guard clauses. We're going to have to go all the way into, e.g. CfeParamBuilders::RegularTransactions so that if needed it can add data from the income_form but not the outgoings_form.

app/controllers/forms_controller.rb
HettieS: Thanks for this! I had a feeling just using the forms in this way wouldn't be quite right - I will look at the Flow::Handler and go from there with your suggestion

app/services/cfe/regular_transactions_payload_service.rb
HettieS: Thanks for this! I added this is as a first step to stop things blowing up as you've said, to sort of get on the right track - will think on this one based on your comment

app/lib/steps/helper.rb
```diff
@@ -49,6 +49,7 @@ def all_sections
          ApplicantDetailsSection,
          IncomeSection,
          PartnerSection,
+         PostGrossIncomeEarlyEligibilitySection,
```
MazOneTwoOne, line 52: I'm guessing we'll also have a `PostCapitalEarlyEligibilitySection` section/screen (potential not `PostDisposableIncomeEarlyEligibilitySection` as we need to know it for contributions)?

patrick-laa: I think it's kinda the other way round. There's no point doing a `PostCapitalEarlyEligibilitySection` because once you've completed the capital questions you're at the end of the flow anyway. 

With disposable income, the point is that we shouldn't try to obtain a result part-way through the outgoings questions, but once we've reached the _end_ of the outgoins questions, a `PostDisposableIncomeEarlyEligibilitySection` would allow us to potentially skip the capital questions entirely if we know a client is ineligible.

(The complication is that _at the moment_ the last disposable income screen is almost at the end of the flow, so there's very little time saving here either, but that might change if we think a reshuffle could save some people some time)

MazOneTwoOne: I see, that makes more sense.

PR #1101, opened 2023-11-15
EL-1263: Satisfaction comments

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1263)

## What changed and why

Allow commenting on satisfaction feedback
Add basic protection to prevent commenting on other people's feedback
Refactor JS now that the 2 feedback flows are so similar.

## Checklist


app/controllers/feedbacks_controller.rb
```diff
@@ -7,23 +7,39 @@ def create
     end
 
     if params[:type] == "satisfaction"
-      SatisfactionFeedback.create!(
+      model = SatisfactionFeedback.create!(
         satisfied: params[:satisfied],
         level_of_help:,
         outcome:,
       )
+      user_satisfaction_feedback_ids << model.id
```
MazOneTwoOne, line 15: question - so we are storing the ID of any new entry of the Satisfaction Feedback table to a new table `user_satisfaction_feedback_ids`, just so we can check if it exists or not?

HettieS (approved):

app/controllers/feedbacks_controller.rb
HettieS: We are collecting the satisfaction feedback ids that the user has created, so that when we update with a comment, we can make sure the user can only update the feedback that they entered, so they can't edit a satisfaction feedback that they didn't create. But I wanted to say good spot @patrick-laa! In practice how would the user have changed the feedback ID in the params?

patrick-laa: This is to make sure that users can only add comments to SatisfactionFeedbacks that they have created. Otherwise a random user could hack the HTML of our web pages to submit a thousand comments to the URL /feedbacks/X (where X is 1 to 1,000) with the text "LAA stinks", and any SatisfactionFeedback with an ID of 1 to 1,000 would have its 'comment' attribute overwritten. This way, we check any incoming comments against the IDs of feedbacks that a specific user has created, so that if the feedback doesn't belong to them, they can't update it.

patrick-laa: @HettieS it wouldn't be too hard - you could just 'Inspect Element' in chrome and manually change the form `action` from `/feedbacks/blah` to `/feedbacks/some-other-blah` right before you hit "Send". There'd be no way of pulling information out of the system, but if you just wanted to "vandalise" by destroying information, that would do it.

PR #1137, opened 2023-11-27
EL-1277: Under 18s - CLR - Select CW form - 2nd Attempt

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1277)

## What changed and why

- showing only a CW2 or CW1&2 controlled work form, if a client is under 18 and the level of help is (CLR) controlled legal representation.

## Guidance to review

N/A

## Checklist


app/models/controlled_work_document_selection.rb
```diff
@@ -16,7 +16,16 @@ class ControlledWorkDocumentSelection
...
+      OPTIONS.select { %i[cw2 cw1_and_2].include?(_1) }
+            .map { { value: _1, options: { label: { text: I18n.t("controlled_work_document_selections.new.option.#{_1}") } } } }
+    else
+      OPTIONS.select { !check.asylum_support || %i[cw1 cw2].include?(_1) }
+            .map { { value: _1, options: { label: { text: I18n.t("controlled_work_document_selections.new.option.#{_1}") } } } }
+    end
+  end
+
+  def client_under_18_feature_flag_and_clr
+    FeatureFlags.enabled?(:under_eighteen, @check.session_data) && @check.session_data["controlled_legal_representation"]
```
patrick-laa, line 29: You want to be careful about this. If they've said the client is under 18 and the case is CLR, then they go back and say the client is _not_ under-18, this check will still return `true`. Because the _feature flag_ is still enabled, and their original CLR answer will still be stored in the session data, even though it's no longer relevant.

Better would be to just check `@check.controlled_legal_representation`. If the feature flag is not enabled, or the client is not under 18, then this method will always return nil (which is falsey). Otherwise it will return the boolean value stored in the session under `controlled_legal_representation`.

(Basically `check` has some clever logic so that when you call a method on it that matches an attribute on a form object, it checks to see whether the step associated with that form is relevant for the current check, and if so, it retrieves that attribute from the session. )

app/models/controlled_work_document_selection.rb
```diff
@@ -16,7 +16,16 @@ class ControlledWorkDocumentSelection
                        if: -> { FeatureFlags.enabled?(:welsh_cw, without_session_data: true) }
 
   def options
-    OPTIONS.select { !check.asylum_support || %i[cw1 cw2].include?(_1) }
-           .map { { value: _1, options: { label: { text: I18n.t("controlled_work_document_selections.new.option.#{_1}") } } } }
+    if client_under_18_feature_flag_and_clr
+      OPTIONS.select { %i[cw2 cw1_and_2].include?(_1) }
```
patrick-laa, line 20: There's a lot of repetition in the two halves of this if/else. Ideally we'd do something like:

```ruby
relevant_forms = if check.controlled_legal_representation
                   %i[cw2 cw1_and_2]
                 elsif check.asylum_support
                   %[cw1 cw2]
                 else
                   OPTIONS
                 end

relevant_forms.map { { value:...
```

That way all the `map`ping and `I18n.t`ing only happens once.

app/models/controlled_work_document_selection.rb
MazOneTwoOne: Thanks, this comment helped, especially the bit about `check`

patrick-laa (approved):

PR #1139, opened 2023-11-28
EL-1285 Update method to account for new CFE result type

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1285)

## What changed and why
 
updated method to determine if a result has been calculated, instead of checking for pending we check if any of the completed results are included.  This ensures that future CFE changes to incomplete result options do not affect us until we have alternative outcomes for these options.

## Checklist

### DIFF
--- a/app/models/calculation_result.rb
+++ b/app/models/calculation_result.rb
@@ -30,7 +30,7 @@ class CalculationResult
   end
 
   def calculated?(section)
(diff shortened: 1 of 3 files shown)


patrick-laa (approved):
Arguably you could make it clearer by changing:

```ruby
....none? { ....exclude?(...) }
```

to:

```ruby
....all? { ....include?(...) }
```

But I don't think it makes much difference either way

PR #1142, opened 2023-11-29
EL-1278: Update to view and added in some check methods.

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1278)

## What changed and why

- Removing references to evidence gathering and 'evidence you might need' on ‘You’ve reached the end of this service' screen in the controlled work flow when a client is under 18 and doesn’t need a means test. 

## Guidance to review

N/A

## Checklist

### DIFF
--- a/app/views/checks/end_of_journey.html.slim
+++ b/app/views/checks/end_of_journey.html.slim
@@ -1,4 +1,3 @@
-- under_18_and_does_not_need_means_test = @check.under_eighteen_controlled_clr? || @check.under_eighteen_no_means_test_required?
 - content_for :page_title
   = t(".title")
 - content_for :back do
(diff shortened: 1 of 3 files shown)


app/views/checks/end_of_journey.html.slim
```diff
@@ -1,3 +1,4 @@
+- under_18_and_does_not_need_means_test = @check.under_eighteen_controlled_clr? || @check.under_eighteen_no_means_test_required?
```
patrick-laa, line 1: This is a bit redundant - ` Steps::Logic.under_eighteen_no_means_test_required?(session_data)` already returns true if ` Steps::Logic.controlled_clr?(session_data)` returns true (see logic.rb line 34), so you don't need to re-check that scenario here.

MazOneTwoOne: Yeah, that makes sense. I was focussed on writing it, so I didn't miss both scenarios that I missed that. Will change now.

patrick-laa (approved):

PR #1153, opened 2023-12-01
EL-1289: Additional property mortgage validation

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1289)

## What changed and why

Fix bug in mortgage validation

## Checklist

### DIFF
--- a/app/models/additional_property_model.rb
+++ b/app/models/additional_property_model.rb
@@ -16,7 +16,8 @@ class AdditionalPropertyModel
   validates :mortgage,
             numericality: { greater_than: 0, allow_nil: true },
             is_a_number: true,
-            presence: { if: :owned_with_mortgage? }
+            presence: true,
+            if: :owned_with_mortgage?
 
   attribute :percentage_owned, :fully_validatable_integer
   validates :percentage_owned,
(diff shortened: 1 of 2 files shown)


willc-work (approved):

app/models/additional_property_model.rb
```diff
@@ -16,7 +16,8 @@ class AdditionalPropertyModel
   validates :mortgage,
             numericality: { greater_than: 0, allow_nil: true },
             is_a_number: true,
-            presence: { if: :owned_with_mortgage? }
+            presence: true,
+            if: :owned_with_mortgage?
```
MazOneTwoOne, line 20: question - why do we have  `if: :owned_with_mortgage?` on a new line instead of the same line as line 19? i.e.
`presence: true, if: :owned_with_mortgage?`

patrick-laa: Because the `if` bit doesn't just refer to the `presence` bit any more. It is equally relevant to `is_a_number` and `numericality` now, so having it line up with `presence` but not the other two would be kinda misleading.

That's effectively the bug: the "is_a_number" validation was firing when the property was _not_ owned with a mortgage, because the "if" clause used to be exclusively associated with the "presence" validation

MazOneTwoOne: Oh...! Thank you.

PR #1155, opened 2023-12-01
EL-1290 speed up testing

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1290)

## What changed and why

added a `:slow` tag to 12 tests, all of which were taking longer than 3 seconds each.
removed one journey from the AXE tests so now we only check `certificated`. The `certificated` AXE test takes longer but covers more of the journey, however if we feel that as we reuse components, nothing would be missed if we used `controlled` I am happy to switch to that journey. 

to run the tests locally without the slow tests: 
`bundle exec rspec -tag=~slow`

## Checklist

### DIFF
--- a/spec/system/accessibility_spec.rb
+++ b/spec/system/accessibility_spec.rb
@@ -78,19 +78,17 @@ RSpec.describe "Accessibility" do
       allow(CfeConnection).to receive(:state_benefit_types).and_return([])
     end
 
-    %w[controlled certificated].each do |level_of_help|
-      it "has no AXE-detectable accessibility issues on any page" do
-        start_assessment
-        fill_in_level_of_help_screen choice: level_of_help
-        assessment_code = current_path.split("/").reverse.second
+    it "has no AXE-detectable accessibility issues on any page", :slow do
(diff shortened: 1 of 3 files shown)


spec/system/accessibility_spec.rb
```diff
@@ -105,7 +103,7 @@
       click_on "Submit"
     end
 
-    it "has no AXE-detectable accessibility issues" do
+    it "has no AXE-detectable accessibility issues", :slow do
```
MazOneTwoOne, line 106: question - how did you find out which tests took the longest? I can't see where I would see that information when running the test's locally or viewing on CircleCI.

MazOneTwoOne (approved):

spec/system/accessibility_spec.rb
willc-work: `bundle exec rspec -p` the `-p` is shorthand for `--profile`
and it gives you the 10 slowest running tests

I split the tests out into 2 separate tests and ran them with the profile flag.

PR #1160, opened 2023-12-06
EL-1294 add metabase 

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1294)

## What changed and why
add metabase deployment and service yaml files
update ingress to point metabase at new hostname

When metabase is enabled in UAT this deployment fails because it cannot read the secret 'metabase'. This secret only exists on staging at present. 

I have confirmed that it creates a metabase pod and it is during this process that the metabase secret is required.

Once this PR is reviewed and it deploys to staging there will to be further work to get metabase working and to hide it behind authentication.

## Checklist

### DIFF
--- a/helm_deploy/laa-estimate-eligibility/templates/ingress.yaml
+++ b/helm_deploy/laa-estimate-eligibility/templates/ingress.yaml
@@ -22,6 +22,10 @@ spec:
       secretName: '{{- .secret }}'
     {{- end }}
   {{- end }}
+  {{- if .Values.metabase.enabled }}
+    - hosts:
+      - "ccq-dashboard-{{ .Values.cfe.environment_name }}.cloud-platform.service.justice.gov.uk"
+  {{- end }}
   rules:
   {{- range .Values.ingress.hosts }}
     - host: '{{- .host }}'
@@ -35,3 +39,16 @@ spec:
                 port:
                   name: http
   {{- end }}
+    {{- if .Values.metabase.enabled }}
+    - host: "ccq-dashboard-{{ .Values.cfe.environment_name }}.cloud-platform.service.justice.gov.uk"
+      http:
+        paths:
+          - path: "/"
+            pathType: ImplementationSpecific
+            backend:
+              service:
+                name: {{ $fullName }}-metabase-service
+                port:
+                  number: 80
+    {{- end }}
+
--- a/helm_deploy/laa-estimate-eligibility/templates/metabase/metabase-deployment.yaml
+++ b/helm_deploy/laa-estimate-eligibility/templates/metabase/metabase-deployment.yaml
@@ -1,0 +1,47 @@
+{{- if .Values.metabase.enabled }}
+{{- $fullName := include "app.fullname" . -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ $fullName }}-metabase
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ $fullName }}-metabase
+  template:
+    metadata:
+      labels:
+        app: {{ $fullName }}-metabase
+    spec:
+    {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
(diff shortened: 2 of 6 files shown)


helm_deploy/laa-estimate-eligibility/templates/_envs.tpl
```diff
@@ -130,5 +130,12 @@ env:
         key: GOOGLE_OAUTH_CLIENT_SECRET
   - name: SEED_ADMINS
     value: {{ .Values.app.seedAdmins }}
+  {{ if .Values.metabase.enabled }}
+  - name: MB_DB_CONNECTION_URI
```
patrick-laa, line 134: As I understand it, putting this here will make the MD_DB_CONNECTION_URI available as an env var in the Rails application pods. Do we need it there? You've got it set up as an env var in metabase-deployment.yaml. Isn't that the only place it needs to be?

helm_deploy/laa-estimate-eligibility/templates/ingress.yaml
```diff
@@ -35,3 +39,16 @@ spec:
                 port:
                   name: http
   {{- end }}
+    {{- if .Values.metabase.enabled }}
+    - host: "ccq-dashboard-staging.cloud-platform.service.justice.gov.uk"
```
patrick-laa, line 43: As this file gets used for UAT, staging and production environments, this will try to register the same host in every environment. Which becomes a problem when metabase is enabled in more than one. Can we interpolate e.g. .Values.cfe.environment_name in place of "staging" here to make it different for each environment? Or define the metabase host name in the values files?

helm_deploy/laa-estimate-eligibility/templates/metabase/metabase-deployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+        securityContext:
+          allowPrivilegeEscalation: false
+          runAsUser: 1000
+          runAsGroup: 1000
+        env:
+          - name: MB_DB_CONNECTION_URI
+            valueFrom:
+              secretKeyRef:
+                name: metabase
+                key: jdbc_url
```
patrick-laa, line 46: So are we creating a new RDS database for Metabase? Or is this URL somehow connected to our existing (staging) database? I'd have thought we'd only want Metabase to talk to our existing database, in which case this MB_DB_CONNECTION_URI maybe needs to be constructed from the `rds-postgresql-instance-output` secrets?

helm_deploy/laa-estimate-eligibility/values/ccq-uat.yaml
```diff
@@ -42,6 +42,8 @@ ingress:
   hosts:
     - host: main-check-client-qualifies-legal-aid-uat.cloud-platform.service.justice.gov.uk
       secret: false
+    - host: ccq-dashboard-staging.cloud-platform.service.justice.gov.uk
```
patrick-laa: As you're now setting up the metabase hosts directly in ingress.yaml, can you delete it from the lists of hosts here and in ccq-staging.yaml?

helm_deploy/laa-estimate-eligibility/templates/metabase/metabase-deployment.yaml
willc-work: you are correct, we should be linking it to our existing RDS, fixed.

helm_deploy/laa-estimate-eligibility/templates/metabase/metabase-deployment.yaml
```diff
@@ -42,6 +42,6 @@ spec:
           - name: MB_DB_CONNECTION_URI
             valueFrom:
               secretKeyRef:
-                name: metabase
+                name: kubernetes_secret
```
patrick-laa, line 45: I believe this secret's name is `rds-postgresql-instance-output`

patrick-laa (approved):
One config thing, but I'm excited to see this working!

PR #1171, opened 2023-12-11
Attempt to update to govuk-frontend 5

patrick-laa:
As dependabot splits into npm and bundler, but this is a change affecting both, here is a PR trying to upgrade all affected packages and get things up and running with the new breaking changes.


HettieS (approved):
I'm happy to approve this. Everything passing and I had a quick run through a flow on the app and didn't spot any issues. The only thing was the tag changing which we are already aware of.

PR #1172, opened 2023-12-12
EL-1288: Means aggregation

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1288)

## What changed and why

Add a new screen with a blank form object so we can show the appropriate guidance text.

## Guidance to review

The `--headless=new` thing is because on my machine, headless chrome is no longer providing browser logs without it (h/t https://github.com/SeleniumHQ/selenium/issues/13112 for solving this)

## Checklist

### DIFF
--- a/app/views/question_flow/how_to_aggregate.html.slim
+++ b/app/views/question_flow/how_to_aggregate.html.slim
@@ -17,7 +17,7 @@
       li = _1
   p.govuk-body
     = t(".paragraph_2_html",
-        link: link_to(t(".link_text"), ExternalLinkService.call(document: :legislation_CLAR_2013_individual_resources)))
+        link: link_to(t(".link_text"), document_link(:legislation_CLAR_2013_individual_resources)))
   ul.govuk-list.govuk-list--bullet
     - t(".bullets_2").each
       li = _1
--- a/app/views/question_flow/how_to_aggregate.html.slim
+++ b/app/views/question_flow/how_to_aggregate.html.slim
@@ -1,0 +1,32 @@
+- content_for :page_title
+  = t(".title")
+- content_for :back
+  = back_link(:how_to_aggregate, @check, @back_buttons_invoke_browser_back_behaviour)
+
+.govuk-grid-column-full
+  = render "shared/heading"
+  .govuk-panel.panel-blue.govuk-panel--confirmation class="govuk-!-text-align-left govuk-!-margin-bottom-9"
+    .govuk-panel__body
+      h1.govuk-heading-l = t(".title")
+      p.govuk-body = t(".panel_text")
+  h2.govuk-heading-s = t(".what_happens_next")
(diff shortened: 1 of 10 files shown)


app/views/question_flow/how_to_aggregate.html.slim
```diff
@@ -0,0 +1,31 @@
...
+      p.govuk-body = t(".panel_text")
+  h2.govuk-heading-s = t(".what_happens_next")
+  - t(".paragraphs_1").each
+    p.govuk-body = _1
+  ul.govuk-list.govuk-list--bullet
+    - t(".bullets_1").each
+      li = _1
+  p.govuk-body
+    = t(".paragraph_2_html",
+        link: link_to(t(".link_text"), ExternalLinkService.call(document: :legislation_CLAR_2013_individual_resources)))
```
HettieS, line 20: just thinking on this, because we are calling the service directly rather than going via the documents controller, we don't track an analytics event here. Maybe that's fine, just a small point.

HettieS (approved):
Approved with one very small comment

app/views/question_flow/how_to_aggregate.html.slim
patrick-laa: You're absolutely right, I forgot how the link system worked! Will fix.

config/locales/en.yml
```diff
@@ -715,6 +715,27 @@ en:
...
+      title: You are going to aggregate your client's means with another person's means
+      panel_text: If your client's means are aggregated with the means of another person, such as a maintaining adult, they must have a full means test.
+      what_happens_next: What happens next
+      paragraphs_1:
+        - "For all questions that refer to 'your client' or 'the client', such as if your client has dependants or owns the home they live in, answer them with reference to both your client and the maintaining adult."
+        - "For questions that ask for financial information:"
+      bullets_1:
+        - add any relevant income or spending amounts for your client to those of the maintaining adult
+        - enter these combined amounts in the relevant fields that refer to 'your client'
+      paragraph_2_html: "If the maintaining adult has a partner, only include the partner if their finances can be included in the means assessment, as per regulation 16 of %{link}, and:"
```
MazOneTwoOne, line 728: @patrick-laa unless I'm mistaken, but I don't think this link open's up into a new tab?

I think we have to change it to:

>     paragraph_2_html: "If the maintaining adult has a partner, only include the partner if their finances can be included in the means assessment, as per regulation 16 of <a href='%{link}' target='_blank'>The Civil Legal Aid (Financial Resources and Payment for Services) Regulations 2013 (opens in new tab)</a>, and:"

Then we can, get rid of` link_text:` and on `app/views/question_flow/how_to_aggregate.html.slim` update it to:

>   p.govuk-body
>     = t(".paragraph_2_html", link: document_link(:legislation_CLAR_2013_individual_resources))

PR #1176, opened 2023-12-14
EL-1305: Conditional reveals on outgoings pages

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1304)

## What changed and why

Add conditional reveal pattern to outgoings pages

## Guidance to review

To make some CFE logic reusable, I changed the other_income boolean names to end with "_relevant" rather than "_received", because for outgoings the money is _paid_ rather than received, so I thought it would make more sense for the attributes in both case to indicate whether the value field is relevant to the check.

Also I found a couple of old housing costs partials that are orphaned, because there is no dedicated housing costs page any more, so I deleted them.

## Checklist


app/services/cfe_param_builders/regular_transactions.rb
```diff
@@ -48,8 +48,8 @@ def self.build_payments(cfe_translations, form, operation)
     end
 
     def self.value(form, local_name)
-      if FeatureFlags.enabled?(:conditional_reveals, form.check.session_data) && form.is_a?(OtherIncomeForm)
-        form.send("#{local_name}_conditional_value") if form.send("#{local_name}_received")
+      if FeatureFlags.enabled?(:conditional_reveals, form.check.session_data)
```
MazOneTwoOne, line 51: question - Should we use a ternary operator to send `0`, as previously user's had to enter `0` for all the values? I guess it's not breaking at the moment when sending to CFE.

app/views/question_flow/forms/_outgoings.html.slim
```diff
@@ -1,28 +1,51 @@
...
-            field: :"#{type}_value",
-            width: 10,
-            label_text: t("question_flow.#{i18n_key}.enter_amount"),
-            set_aria: true
-    = form.govuk_collection_radio_buttons :"#{type}_frequency",
-                                          @form.frequencies,
-                                          :first,
-                                          :last,
-                                          legend: -> { tag.p(t("question_flow.#{i18n_key}.enter_frequency")) }
+    - if FeatureFlags.enabled?(:conditional_reveals, @check.session_data)
```
MazOneTwoOne, line 12: question - why not use the `conditionals` variable you created?

app/views/question_flow/forms/_outgoings.html.slim
```diff
@@ -1,28 +1,51 @@
 - partner ||= false
+- conditionals = FeatureFlags.enabled?(:conditional_reveals, @check.session_data)
```
MazOneTwoOne, line 2: comment - maybe re-name to `conditionals_reveal_flag`

MazOneTwoOne (approved):
Just some small things and bits I was confused with.

config/locales/en.yml
```diff
@@ -370,8 +384,6 @@ en:
               blank: Enter payments towards legal aid in a criminal case, if this does not apply enter 0
               not_a_number: Payments towards legal aid in a criminal case must be a number, if this does not apply enter 0
               greater_than_or_equal_to: Payments towards legal aid in a criminal case  must be 0 or more
-            housing_payments_frequency:
-              blank: Select frequency of housing payments
             childcare_payments_frequency:
               blank: Select frequency of childcare payments
```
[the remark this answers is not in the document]
patrick-laa: Yeah so I didn't really want to touch the frequency field. The only way of having different error messages depending on whether the feature flag is turned on or off (without rewriting our own system for mapping error messages to validations) is to have the field named differently if the flag is on, but it's a right pain to have to accommodate 2 different field names - I did it for "value" and "conditional_value" but I wasn't keen to do it again. I was hoping I could get away, for the purposes of the accessibility audit, of leaving the error messages for frequency untouched, but I'll ask Kate to confirm whether that's OK.

app/services/cfe_param_builders/regular_transactions.rb
patrick-laa: This class is already filtering out zeroes (see above `.select { |_cfe_name, local_name| value(form, local_name).to_i.positive? }`, so it makes no ends if we return nil or 0 here - it'll be filtered out regardless.

app/views/question_flow/forms/_outgoings.html.slim
patrick-laa: Good idea

app/views/question_flow/forms/_outgoings.html.slim
patrick-laa: Because I'm a muppet. Will fix.

app/services/cfe_param_builders/regular_transactions.rb
MazOneTwoOne: Oh yeah, that makes sense. I didn't scroll up.

PR #1177, opened 2023-12-15
BUG: When forms are reordered, property isn't skipped when means test should be skipped

patrick-laa:
## What changed and why

The outgoings section was always asking the property question even in situations where capital questions should be skipped.

## Checklist


exonian (approved):
This looks to me like it works. I'm not familiar with the code, nor the conventions, but the order feels a _little_ unnatural to me. A bit inside out, maybe? I wonder if a flow of handling `skip_capital_questions?` before checking the feature flag would make things any easier to read, or just be complexity for its own sake?

Absolutely happy for this suggestion to be taken under advisement and you to merge this as is, just thought I'd throw an observation in.

PR #1178, opened 2023-12-15
EL-1311: Skip irrelevant CW fields when u18

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1311)

## What changed and why

Some CW fields were being filled in even when irrelevant when an u18 client was exempt from a means test.

## Checklist

### DIFF
--- a/spec/services/controlled_work_document_value_mapping_service_spec.rb
+++ b/spec/services/controlled_work_document_value_mapping_service_spec.rb
@@ -470,6 +470,27 @@ RSpec.describe ControlledWorkDocumentValueMappingService do
     end
   end
 
+  context "when the client is under 18 and needs no means test", :under_eighteen_flag do
+    let(:session_data) do
+      FactoryBot.build(
+        :minimal_complete_session,
+        client_age: "under_18",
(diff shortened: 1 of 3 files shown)


spec/services/controlled_work_document_value_mapping_service_spec.rb
```diff
@@ -470,6 +470,27 @@
     end
   end
 
+  context "when the client is under 18 and needs no means test", :under_eighteen_flag do
```
MazOneTwoOne, line 473: question - do you think we should also have a mapping for the Welsh forms or is that excessive?

MazOneTwoOne (approved):

spec/services/controlled_work_document_value_mapping_service_spec.rb
patrick-laa: I think it's excessive - what we're testing is not a particular form but a behaviour common to all forms - so we just need to pick one, and it could be either an English or a Welsh one, but I don't think we get more reassurance from doing both.

PR #1182, opened 2023-12-20
EL-1312: Remove U18 feature flag

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1312)

## What changed and why

- Removed the 'under 18' feature flag, updated applicant_form (by removing `:over_60` option) and updated relevant tests

## Guidance to review

N/A

## Checklist


app/models/check.rb
```diff
@@ -98,16 +98,8 @@ def eligible_for_childcare_costs?
     ChildcareEligibilityService.call(self)
   end
 
-  def show_over_60_question?
-    !FeatureFlags.enabled?(:under_eighteen, session_data)
-  end
-
   def consolidated_client_age
```
willc-work, line 101: is this method still required? it appears to be only used once in journey_logger_service.rb can we just use client_age?

willc-work (approved):
one query/comment but not a blocker

app/models/check.rb
MazOneTwoOne: @willc-work have updated and removed this method

PR #1184, opened 2023-12-21
EL-1292: In CircleCi, run tests in parallel - WIP

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1292)

## What changed and why

add parallelism to circleci for running specs
remove 100% coverage requirement on circle
store the test results
add new job to collate results
add new rake task to collate results and ensure 100% coverage

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -136,6 +136,7 @@ jobs:
           template: basic_fail_1
           branch_pattern: main
   run_specs:
+    parallelism: 5
     executor: test-executor
     steps:
       - browser-tools/install-browser-tools:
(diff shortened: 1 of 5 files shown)


Rakefile
MazOneTwoOne, line 1: question - we now have an empty Rakefile, what does it do?

.circleci/config.yml
```diff
@@ -278,7 +313,6 @@ workflows:
           environment: ccq-uat
           context: laa-check-client-qualifies-uat
           requires:
-            - linters
```
MazOneTwoOne, line 281: comment - hopefully you agree we should have 2 `linters` in this requires section.

.circleci/config.yml
```diff
@@ -136,7 +136,9 @@ jobs:
           template: basic_fail_1
           branch_pattern: main
   run_specs:
+    parallelism: 4
     executor: test-executor
+    resource_class: small
```
MazOneTwoOne, line 141: comment - thinking about it, we possible don't event need to put `resource_class: small` here as it's already defined in `executor: test-executor`

MazOneTwoOne (commented):
Just some comments. I would approve, but I am the author of the PR.

Rakefile
willc-work: its not an empty rakefile, the changes to the file were modifying the permissions

.circleci/config.yml
willc-work: yes looks like duplication so good to remove it

Rakefile
willc-work: but those modifications turned out to be unnecessary so I have reverted them

exonian (approved):
Looks good. Speeds things up well, and we can tune the number of nodes if the docker layer caching PR automatically speeds up the setup. If not, we can make optimisation its own ticket.

PR #1187, opened 2023-12-28
EL-1293: Speed up deployment in CircleCi through 'build_and_push' job

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1293)

## What changed and why

- Updated orb version
- added `docker_layer_caching: true` to the `aws-ecr: circleci/aws-ecr@9.0.1` orb

## Guidance to review

My resources included:
- [website on custom executors](https://dev.to/otomato_io/how-to-configure-circleci-what-you-might-not-know-27e)
- [closed query about aws-ecr-orb](https://github.com/CircleCI-Public/aws-ecr-orb/issues/205) 
- ['money to pirsoners' repo](https://github.com/ministryofjustice/money-to-prisoners-api/blob/c282731ddb9d15659ee6452983be5743d96ea2d7/.circleci/config.yml#L56)

## Checklist


Dockerfile
```diff
@@ -15,22 +15,21 @@ RUN addgroup -g 1000 -S appgroup \
   && adduser -u 1000 -S appuser -G appgroup
 
 # Add the timezone (builder image) as it's not configured by default in Alpine
-RUN apk add --update --no-cache tzdata && \
```
exonian, line 18: Do these changes speed things up?
My understanding is that this `--no-cache` is an argument to `apk add`, rather than Docker.

Perhaps this does slow things down, or maybe theoretically it speeds things up slightly (if it doesn't have to store things it won't need, I dunno), but I think it wouldn't cause the docker layer to have to be rebuilt, which as we saw the other week is the real time sink.

I'm interested why the change was made, if you folks experimented with it both ways after getting the docker cache working, if it came from some other example, or if it kinda got caught up with the other stuff.

Basically if we haven't tested it both ways since enabling docker layer caching, I would like us to.

.circleci/config.yml
```diff
@@ -216,8 +217,9 @@ jobs:
             [[ "$CIRCLE_BRANCH" == "main" ]] && ECR_TAG="main-$CIRCLE_SHA1" || ECR_TAG="branch-$CIRCLE_SHA1"
             echo "export ECR_TAG=$ECR_TAG" >> "$BASH_ENV"
             source "$BASH_ENV"
-      - aws-ecr/build-image:
-          push-image: true
+      - aws-ecr/build_image:
+          push_image: true
+          account_id: $AWS_ECR_REGISTRY_ID
```
exonian, line 222: Is this related to the version change of the `aws-ecr` orb?

exonian (commented):
Generally looks great, just a couple of queries, mostly around the caching *within* the docker layers (where we're now caching the whole layer)

.circleci/config.yml
willc-work: Yes this field is now required for build_image
https://circleci.com/developer/orbs/orb/circleci/aws-ecr#commands-build_image

willc-work: and the env_var was already in use in circleci

Dockerfile
```diff
@@ -30,7 +30,6 @@ COPY .ruby-version Gemfile Gemfile.lock ./
 
 # Install gems and remove gem cache
 RUN bundler -v && \
-    bundle config set no-cache 'true' && \
```
willc-work, line 33: should this also be reverted, this is a bundler command.

Dockerfile
MazOneTwoOne: @exonian - Yes, I had a false positive with the many variations I experimented with to the `config.yml` and `dockerfile` files. Now reverting `dockerfile` changes back.


If you're curious, can follow along with my commit history and [CircleCi pipeline](https://app.circleci.com/pipelines/github/ministryofjustice/laa-check-client-qualifies?branch=EL-1293-build-and-push-job);

Variation 1 (false postive)
- From [this commit](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1187/commits/1974f3c4db18426525ca2ad0c3842625ceae2e1a) (1974f3c) - I set-up `config.yml` to have docker layer caching (dlc), with no changes to Dockerfile.
- Caching was not happening in the subsequent [commit](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1187/commits/93d4aa3bcc656dbf21d9f1c7b1ced8f5418ffcee) (93d4aa3)
- I then updated Dockerfile in the next [commit](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1187/commits/e07473b77ee1290e807e9f9f6a82a1d3e116f801) (e07473b). My naive rationale for making changes to the `dockerfile` was, if we want to docker to cache things then why were we using [`--no-cache`](https://stackoverflow.com/questions/35594987/how-to-force-docker-for-a-clean-build-of-an-image) in the `dockerfile`. 😅
- Then the subsequent [commit](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1187/commits/43e36772b39a324ce69ac2fea935e114075cb4fe) (43e3677), dlc was taking place.

Variation 2 (confirming that removing `--no-cache ' an argument doesn't make a difference)
- In [this commit](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1187/commits/3eeaffe5423235a7cdeefae8189c55ade8dabdca) (3eeaffe) dlc set-up in `config.yml`, and added `--no-cache` back into `dockerfile`.
- Subsequent [commit](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1187/commits/5ce8fcc234eba75b6409d40a3cc96e8669789ed9) (5ce8fcc) dlc set-up and` build_an push` step still sped up.

Dockerfile
MazOneTwoOne: yes, now reverted.

Dockerfile
exonian: What a fantastic summary, thank you! I was gonna look commit-wise but it was 45 commits so decided against 😅 So this is really helpful

.circleci/config.yml
exonian: Thanks!

exonian (approved):

PR #1193, opened 2024-01-02
EL-1318 in circle only install required browsers

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1318)

The circleci orb browser-tools installs a few browsers by default. we currently only use chrome. the firefox install was taking 22 seconds and we dont use it so this PR stops these unused browsers being installed.

This PR also stops specifying the version of chrome and will use an existing chrome if it is available. The issue that required us to do this appears to have been fixed in an update to the orb

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,6 +141,7 @@ jobs:
       - browser-tools/install-browser-tools:
           install-firefox: false
           install-geckodriver: false
+          install-chromedriver: false
       - checkout
       - attach_workspace:
           at: ./
@@ -314,7 +315,7 @@ workflows:
           environment: ccq-uat
           context: laa-check-client-qualifies-uat
           requires:
-            - run_specs
+            - coverage
             - end2end_tests
             - linters
             - build_and_push
@@ -330,7 +331,7 @@ workflows:
           environment: ccq-staging
           context: laa-check-client-qualifies-staging
           requires:
-            - run_specs
+            - coverage
             - end2end_tests
             - linters
             - build_and_push
@@ -350,7 +351,7 @@ workflows:
           environment: ccq-production
           context: laa-check-client-qualifies-production
           requires:
-            - run_specs
+            - coverage
             - end2end_tests
             - linters
             - build_and_push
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -140,8 +140,10 @@ jobs:
     executor: test-executor
     steps:
       - browser-tools/install-browser-tools:
-          chrome-version: 116.0.5845.96 # TODO: remove when chromedriver downloads are fixed
```
MazOneTwoOne, line 143: question - just checking that this [Issue resolution for the orb](https://github.com/CircleCI-Public/browser-tools-orb/issues/95#issuecomment-1718306019), confirms that ChromeDriver's can now be installed? 

Just asking as there is a recent [open issue](https://github.com/CircleCI-Public/browser-tools-orb/issues/99), where chrome failed to start. Not sure if we'll encounter the same thing?

.circleci/config.yml
```diff
@@ -140,8 +140,10 @@ jobs:
...
-          replace-existing-chrome: true
+          replace-existing-chrome: false
```
MazOneTwoOne, line 143: question - do we even need to have `replace-existing-chrome` if it defaults to [false](https://circleci.com/developer/orbs/orb/circleci/browser-tools)?

MazOneTwoOne (approved):
Just a few questions, but happy to approve.

.circleci/config.yml
willc-work: No  we do not, will remove

.circleci/config.yml
willc-work: the original issue and the linked one both refer to issues with chromedriver being installed, in this PR I have removed the installation of chromedriver and the tests are all working as expected.

.circleci/config.yml
```diff
@@ -314,7 +316,7 @@ workflows:
           environment: ccq-uat
           context: laa-check-client-qualifies-uat
           requires:
-            - run_specs
+            - coverage
```
MazOneTwoOne, line 319: comment - I think it make sense that `coverage` is now the dependancy that needs to happen before we deploy.

PR #1197, opened 2024-01-04
EL-1315: Fix lowercase tag text

patrick-laa:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1315)

## What changed and why

Use localised string for issue tags to fix capitalisation

## Checklist


app/views/updates/_issue.html.slim
```diff
@@ -4,7 +4,8 @@
   hr class="govuk-!-margin-bottom-4"
   h3.govuk-heading-s class="govuk-!-margin-bottom-0"
     = issue.title
-    .govuk-tag class="govuk-!-margin-left-3 govuk-tag--#{issue.status == Issue.statuses[:active] ? "red" : "grey"}" = issue.status
+    .govuk-tag class="govuk-!-margin-left-3 govuk-tag--#{issue.status == Issue.statuses[:active] ? "red" : "grey"}"
+      = t("updates.index.tags.#{issue.status}")
```
MazOneTwoOne, line 8: comment - so instead of using the value of `issue.status` to be displayed in the tag element, we are now mapping to a translation file?

MazOneTwoOne (approved):

app/views/updates/_issue.html.slim
patrick-laa: First point - yes. We were relying on the fact that the value of the status attribute happened to be roughly the same as the text we wanted displayed. This was a dodgy assumption, as demonstrated by the capitalisation change. So introducing a layer of abstraction, by now making the value of the status attribute the I18n key that holds the text we want displayed, protects us from further issues and gives us flexibility if we want to change the wording of the tags in future without having to rewrite any other logic.

Second point - yes, we will need to set up some active and resolved issues on the UAT branch in order to make these tags visible.

PR #1201, opened 2024-01-08
EL-1320: Part 2 - Speed up run_specs

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1320)

## What changed and why

Our `run_specs` testing Job now runs in parallelised mode, meaning some number of containers (currently 3) share out the tests and run them simultaneously. This allows us to the run tests much more quickly, but not as quickly as (total time if run in one container)/(number of containers), due to setup time.

- Updated image used for `test-executor`, using a GitHub workflow/action based on a new dockerfile. 
- New `run_specs` steps are 
  - download image with chrome and puppeteer installed (as part of the `test-executor`)
  - checkout code
  - [existing steps remain the same]
- Added new CircleCi & GitHub `env_vars` / `secrets`
  - DOCKERHUB_USER_CCQ
  - DOCKERHUB_PAT_CCQ
 - Created a shared DockerHub account

## Checklist


.circleci/config.yml
```diff
@@ -8,15 +8,18 @@ executors:
   test-executor:
     resource_class: small
     docker:
-      - image: cimg/ruby:3.2.2-browsers
+      - image: checkclientqualifiesdocker/welcome-to-docker:latest
```
willc-work, line 11: can we change this `welcome-to-docker` bit and call it something like `circleci_image`?

MazOneTwoOne: @willc-work now updated 👍🏽

Dockerfile_browser_tools.dockerfile
```diff
@@ -0,0 +1,18 @@
+# Build browser tools for test-executor
+# Use the cimg/ruby:3.2.2-browsers image as the base image to extend out
+FROM cimg/ruby:3.2.2-browsers
+
+WORKDIR /app
+
+# Install Chrome
+RUN npx puppeteer browsers install chrome
```
psweeting1, line 8: You can combine line 8 and line 11 together like so:

```
RUN npx puppeteer browsers install chrome \
    && sudo yarn add puppeteer@21.5.0
```

Dockerfile_browser_tools.dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# Install Puppeteer with Chromium
+RUN sudo yarn add puppeteer@21.5.0
+
+# Install PDFTK
+RUN sudo add-apt-repository --yes ppa:malteworld/ppa || true
+RUN sudo apt update --allow-unauthenticated || true
+RUN sudo apt install pdftk --allow-unauthenticated || true
```
psweeting1, line 16: Like wise here we can put them into the same RUN command like so: 

```
RUN sudo add-apt-repository --yes ppa:malteworld/ppa \
    && sudo apt-get update --allow-unauthenticated \
    && sudo apt-get install -y pdftk --allow-unauthenticated
```

Just a question though, don't we want things to fail so we know there is an issue with any of the commands if we use `|| true`?

psweeting1 (commented):
Great work! I just added some tidy comments and a question around a shell command

Dockerfile_browser_tools.dockerfile
MazOneTwoOne: @willc-work - Also agree with getting rid of the `|| true`, I should have removed this now that we have trialed and tested this.

PR #1204, opened 2024-01-10
EL-1308: Early eligibility, client gross income flow - WIP

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1308)

## What changed and why

Opening a WIP so we can all have a look and work on it 

## Checklist


app/controllers/forms_controller.rb
```diff
@@ -16,4 +23,8 @@ def update
       render "question_flow/#{step}"
     end
   end
+
+  def show_early_result_screen?
```
starswan, line 27: This feels like we are calling the feature flag too often. The flow model revolves around 'completing' screens, which I think we need to keep in step with. I think it might be worth adding an extra 'Form' for the interruption screen, that 'completes' when visited (if that is the required behaviour) and is shown when ineligible_gross_income is true. This form could then be inserted into the flow only when the feature flag is true. WDYT?

app/services/cfe/base_service.rb
```diff
@@ -1,12 +1,13 @@
 module Cfe
   class BaseService
-    def self.call(session_data, payload)
-      new(session_data, payload).call
+    def self.call(session_data, payload, early_eligibility = nil)
```
starswan, line 3: Think this can only be nil if the feature flag is false - think it might be better being explicitly passed in if so, otherwise it might get left in the codebase when the feature flag is removed.

app/controllers/early_results_controller.rb
```diff
@@ -0,0 +1,29 @@
+class EarlyResultsController < ApplicationController
```
starswan, line 1: This feels like it is outside the normal flow, which is really hard to reason about. I think we have a decent data model in here - maybe it should be made more explicit in a refactoring prior to this work? If the data model were more obvious in the code, maybe this work could be either incorporated into it or use an 'adapter' pattern to extend it if that it what we need to do? [In CFE civil, we moved the calculation into one place, and then wrote an adapter to take the 'wrong' result and correct it based on the (possible) incompleteness of the data)

app/controllers/forms_controller.rb
HettieS: I would be interested to talk this through and see what it looks like, I think adding a form could be a good idea. Agree we are calling the flag a lot and if we can find a way to do that less/more cleanly, that would be good.

app/services/cfe/base_service.rb
HettieS: This can be nil when the CFE service is called at the end of the journey -  I added an optional argument for when we call the CFE service early, when we have less information to send over but can still provide an ineligible outcome. This is because currently I am using the presence of `early_eligibility` as a guard clause in the CFE service, so we know we can ignore sending certain bits of info to CFE. But this was a starting point so maybe there is a better refactor!

app/controllers/early_results_controller.rb
HettieS: Originally when we were speccing this out with design, we were thinking of implementing a screen that could pop up at any point in the flow and signal an early ineligible result. Because of that, and because of the way we currently generate the different sections of steps, we went with an idea of leaving the flow, and hooking back into it again if the user decided to resume their check and add more information. For a V1 implementation we went with showing the interruption screen at the end of certain sections (with the understanding that we'd like to throw this interruption screen into the flow at potentially any point in the future) but maybe leaving the flow makes less sense in this current V1 scope as we know where the screen will be. I'm not sure about the adapter pattern but would be interested to hear more :)

app/controllers/forms_controller.rb
starswan: I'm slightly concerned that adding the interruption screen might be a bit 'too simple' - I need to think some more

app/services/cfe/base_service.rb
starswan: My suggestion was just to move the default up - so making the parameter nil explicitly in the caller

app/controllers/early_results_controller.rb
starswan: Putting it in at any point does mess up the existing (software) design somewhat. Wonder if there would be any mileage in having 3 screens with the same content to simplify things? Could the step be dependant on the CFE result somehow? [If its only at the end of the sections, then think a simple version of adding a screen works]

HettieS: We did actually have a go at spiking something like that out in one of our earlier attempts! We added the relevant screen section and the idea was to have a different form/view for each screen type - something like this commit: https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1089/commits/a10db289d654c6e6cad1764b3ebf52047bffda2f

We moved away from that idea when it seemed we might be showing the screen at any point in the journey, and also weren't sure if we should duplicate the screen logic for the different screen types (the continuous check, where the screen could pop up at any point is still the aim after the V1 implementation as far as I'm aware, but that might change). I'm not against adding in a screen that follows the normal pattern though if we think it is the better way to go. Just might be harder to refactor later on I guess.

app/forms/ineligible_gross_income_form.rb
```diff
@@ -0,0 +1,9 @@
+class IneligibleGrossIncomeForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include SessionPersistable
+
+  ATTRIBUTES = [:check_answers].freeze
+  attribute :check_answers
+  validates :check_answers, inclusion: { in: %w[gross return], allow_nil: true }
```
HettieS, line 8: change this attribute name - early_eligibility_selection

app/services/cfe_service.rb
```diff
@@ -1,6 +1,16 @@
...
+      invoke(session_data, relevant_steps)
+    end
+
+    def result(session_data, completed_steps)
+      CfeResult.new invoke(session_data, completed_steps)
+    end
+
+  private
+
+    def invoke(session_data, relevant_steps)
```
HettieS, line 13: we may not need invoke anymore - split this out for testing

app/services/cfe/regular_transactions_payload_service.rb
```diff
@@ -7,7 +7,7 @@ def call
         relevant_form?(:housing_costs) ||
         relevant_form?(:benefit_details)
 
-      outgoings_form = instantiate_form(OutgoingsForm)
+      outgoings_form = instantiate_form(OutgoingsForm) if relevant_form?(:outgoings)
       income_form = instantiate_form(OtherIncomeForm)
```
HettieS, line 11: do we need to check the form is relevant?

app/controllers/change_answers_controller.rb
```diff
@@ -55,4 +74,15 @@ def anchor
   def page_name
     "check_#{step}"
   end
+
+  def is_last_step_in_applicant_case_details?(the_step)
```
HettieS, line 78: last_non_finance_step

app/controllers/change_answers_controller.rb
```diff
@@ -55,4 +74,15 @@ def anchor
...
+    non_finance_steps = Steps::Helper.steps_for_section(session_data, Steps::CaseDetailsSection) +
+      Steps::Helper.steps_for_section(session_data, Steps::ApplicantDetailsSection)
+    non_finance_steps.last == the_step || the_step == :other_income
+  end
+
+  def step_in_applicant_or_case_details?(the_step)
```
HettieS, line 84: non_finance_step?

app/lib/steps/logic.rb
```diff
@@ -137,6 +137,18 @@ def dependants_get_income?(session_data)
 
         session_data["dependants_get_income"]
       end
+
+      def user_chose_to_continue_check?(session_data)
+        session_data["check_answers"] == "return"
```
HettieS, line 142: change the attribute and value where necessary

HettieS: form constants for the value

app/controllers/change_answers_controller.rb
```diff
@@ -9,6 +9,25 @@ def update
...
+              next_step = nil
+            elsif next_step.present?
+              flash[:notice] = I18n.t("service.change_eligibility")
+            end
+          else
+            cfe_result = CfeService.result(session_data, completed_steps)
+            if Steps::Logic.data_stops_before_outgoings?(session_data) && cfe_result.ineligible_gross_income?
+              next_step = nil
+            end
+            if Steps::Logic.data_stops_before_outgoings?(session_data) && !cfe_result.ineligible_gross_income?
```
HettieS, line 26: and next_step.present?

app/controllers/checks_controller.rb
```diff
@@ -31,4 +36,8 @@ def assessment_code
   def specify_feedback_widget
     @feedback = action_name == "end_of_journey" ? :satisfaction : :freetext
   end
+
+  def show_eligibility_banner
```
HettieS, line 40: remove

app/views/shared/_heading.html.slim
```diff
@@ -20,6 +20,8 @@ ruby:
       p.govuk-body-m class="govuk-!-margin-top-5" = t("service.beta_message_pilot_body_1")
       p.govuk-body-m = t("service.beta_message_pilot_body_2_html", public_beta_link: ENV["PRIMARY_HOST"])
 
+= render "layouts/flash_banner" if @show_eligibility_banner
```
HettieS, line 23: remove the method

spec/features/downloads_spec.rb
```diff
@@ -6,7 +6,7 @@
   it "gives me a download option for an eligible certificated check" do
     allow(CfeConnection).to receive(:state_benefit_types).and_return([])
 
-    allow(CfeService).to receive(:call).and_return(api_response)
+    allow(CfeService).to receive(:invoke).and_return(api_response)
```
HettieS, line 9: add stubbing flag and remove invoke stuff

spec/flows/early_result_change_answers_flow_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+      confirm_screen("check_answers")
+    end
+
+    it "does not display the flash when they are still ineligible" do
+      within "#table-income" do
+        click_on "Change"
+      end
+      fill_in_income_screen(gross: "2700", frequency: "Every month")
+      confirm_screen("check_answers")
+      expect(page).not_to have_content("Based on the answers you changed, your client is now within the limit for legal aid")
```
HettieS, line 114: test for content and selector for banner

HettieS: make content a let

spec/flows/early_result_controlled_non_passported_spec.rb
```diff
@@ -0,0 +1,58 @@
+require "rails_helper"
+
+RSpec.describe "Early Result controlled, non-passported flow", :stub_cfe_gross_ineligible, :early_eligibility_flag, type: :feature do
```
HettieS, line 3: delete this spec file and rename to early eligibility journey

spec/forms/other_income_form_spec.rb
```diff
@@ -1,11 +1,13 @@
 require "rails_helper"
 
-RSpec.describe "other_income", type: :feature do
+RSpec.describe "other_income", :stub_cfe_calls, type: :feature do
   let(:assessment_code) { :assessment_code }
   let(:feature_flags) { {} }
 
   before do
-    set_session(assessment_code, "level_of_help" => "controlled", "feature_flags" => feature_flags)
+    set_session(assessment_code, "level_of_help" => "controlled", "feature_flags" => feature_flags,
```
HettieS, line 8: remove this and make it into a feature spec that the runs through the flow until

spec/services/cfe/regular_transactions_payload_service_spec.rb
```diff
@@ -166,6 +166,8 @@
       end
       let(:relevant_steps) { %i[outgoings partner_outgoings housing_costs] }
 
+      let(:completed_steps) { %i[outgoings partner_outgoings housing_costs] }
```
HettieS, line 169: delete

spec/views/check_answers_page/client_content/assets_spec.rb
```diff
@@ -139,7 +139,7 @@
         context "when no other assets" do
           let(:session_data) do
             build(:minimal_complete_session,
-                  bank_accounts: [{ "amount" => 0 }],
+                  bank_accounts: [{ "amount" => 0, "account_in_dispute" => false }],
```
HettieS, line 142: test if this can go

spec/views/check_answers_page/client_content/dependants_spec.rb
```diff
@@ -37,8 +37,8 @@
       context "when no dependants" do
         let(:session_data) do
           build(:minimal_complete_session,
-                child_dependants: nil,
-                adult_dependants: nil)
+                child_dependants: false,
```
HettieS, line 40: double check this

spec/views/check_answers_page/client_content/housing_benefit_spec.rb
```diff
@@ -14,15 +14,15 @@
   describe "client sections" do
     context "when housing benefit" do
       context "when client has housing benefit" do
-        let(:session_data) { build(:minimal_complete_session, housing_benefit_value: 400, housing_benefit_frequency: "every_week") }
+        let(:session_data) { build(:minimal_complete_session, housing_payments: 1000, housing_payments_frequency: "every_week", housing_benefit_value: 400, housing_benefit_frequency: "every_week") }
```
HettieS, line 17: check if required but fine to leave

spec/views/check_answers_page/client_content/special_client_types/passported_client_spec.rb
```diff
@@ -33,7 +33,13 @@
     end
 
     context "with a partner" do
-      let(:session_data) { build(:minimal_complete_session, partner: true, passporting: true) }
+      let(:session_data) do
```
HettieS, line 36: double check

spec/views/check_answers_page/partner_content/other_benefits_spec.rb
```diff
@@ -61,7 +61,7 @@
       end
 
       context "when there are no other partner benefits" do
-        let(:session_data) { build(:minimal_complete_session, :with_partner, benefits: [], partner_benefits: []) }
+        let(:session_data) { build(:minimal_complete_session, :with_partner, partner_receives_benefits: false, benefits: [], partner_benefits: []) }
```
HettieS, line 64: double check

spec/views/check_answers_page/partner_content/outgoings_spec.rb
```diff
@@ -35,6 +35,8 @@
         let(:session_data) do
           build(:minimal_complete_session,
                 :with_partner,
+                partner_maintenance_payments_value: 0,
```
HettieS, line 38: double check

app/services/cfe/regular_transactions_payload_service.rb
HettieS: Stephen has implemented this change here: https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1261

so won't do on this PR

spec/forms/other_income_form_spec.rb
HettieS: we actually have so many of these form specs so instead of getting into refactoring this one now, maybe we want to take a look at the together as a bigger test refactor. I have removed the changes to the session and the tests still pass with flag on and off so would like to leave them as is for now.

willc-work (approved):
lgtm

willc-work:
as the commits are not atomic we might as well squash this into a single commit before it is merged

app/forms/ineligible_gross_income_form.rb
```diff
@@ -0,0 +1,13 @@
+class IneligibleGrossIncomeForm
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+  include SessionPersistable
+
+  SELECTION = :early_eligibility_selection.to_s
+
+  ATTRIBUTES = [:early_eligibility_selection].freeze
+  VALID_OPTIONS = { gross: "gross", continue: "continue_check" }.freeze
```
willc-work, line 9: can we change gross an income to something more descriptive at first glance:
early_gross_exit and early_eligibility_continue seem like good options, that allow us to extend for disposable an cpaital

willc-work (approved):
one comment on names

willc-work (approved):
Just one small suggestion over naming things that isnt a blocker

PR #1213, opened 2024-01-24
EL-1273 update maintenance page

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1273)

Update content for maintenance period

Turned on maintenance mode in UAT to check content in UAT, this will be reverted before merge

## Checklist

### DIFF
--- a/public/maintenance.html
+++ b/public/maintenance.html
@@ -49,8 +49,16 @@
         <div class="govuk-grid-row">
           <div class="govuk-grid-column-two-thirds">
             <h1 class="govuk-heading-l">Sorry, the service is unavailable</h1>
-            <p class="govuk-body">You will be able to use the service from 8:30pm.</p>
-            <p class="govuk-body">If you are part way through a check, we saved your answers. They will be available for 14 days from when you started the check.</p>
+            <p class="govuk-body">You will be able to use the service from 8pm on Wednesday 24 January 2024.</p>
+            <p class="govuk-body">Use our
+              <a href="https://www.gov.uk/guidance/funding-and-costs-assessment-for-civil-and-crime-matters">means assessment resources</a> for guidance on checking a client's eligibility.
+            </p>
+            <p class="govuk-body">Contact our customer services team for support with legal aid administration or training:</p>
+            <ul class="govuk-list govuk-list--bullet">
+              <li>telephone: 0300200 2020</li>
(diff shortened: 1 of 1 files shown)


public/maintenance.html
```diff
@@ -48,9 +48,8 @@
       <main class="govuk-main-wrapper govuk-main-wrapper--l" id="main-content" role="main">
         <div class="govuk-grid-row">
           <div class="govuk-grid-column-two-thirds">
-            <h1 class="govuk-heading-l">Sorry, the service is unavailable</h1>
-            <p class="govuk-body">You will be able to use the service from 8:30pm.</p>
-            <p class="govuk-body">If you are part way through a check, we saved your answers. They will be available for 14 days from when you started the check.</p>
+            <h1 class="govuk-heading-l">Sorry, the service is unavailable because of planned maintenance</h1>
+            <p class="govuk-body">It will not be available on Wednesday 24 January 2024 from 6pm for up to 2 hours.</p>
```
MazOneTwoOne, line 52: Hey @willc-work, should we call this part 1 of EL-1273? 

I guess part 2 would be to put maintenance mode to `NOT_ENABLED` and add in the generic text as [per the ticket](https://dsdmoj.atlassian.net/browse/EL-1273)?

MazOneTwoOne (approved):

public/maintenance.html
willc-work: this only turns maintenance mode on in UAT, once it has been UAT'd that will be reverted before it is merged

willc-work: I will clarify with Laurence re: the generic content

public/maintenance.html
```diff
@@ -48,9 +48,17 @@
...
+            <p class="govuk-body">Use our
+              <a href="https://www.gov.uk/guidance/funding-and-costs-assessment-for-civil-and-crime-matters">means assessment resources</a> for guidance on checking a client's eligibility.
```
MazOneTwoOne, line 54: @willc-work Should we add `target="_blank"` within the html anchor tag so the links open up in a new tab?

willc-work: I didnt because if we wanted that to be the beahviour we would include that notice in the link and the docs dont show that,but I will confirm with Laurence

MazOneTwoOne: Yeah that makes sense.

PR #1218, opened 2024-01-24
Revert "EL-1320: Part 2 - Speed up run_specs"

MazOneTwoOne:
Reverts ministryofjustice/laa-check-client-qualifies#1201


willc-work (approved):
Reverting the changes that broke the pipeline

PR #1224, opened 2024-01-25
Add arm64-darwin-23 to platforms in Gemfile.lock

starswan:
No ticket

## What changed and why

ARM M1 with Sonoma OS is a new platform for the project, and so needs to be added to Gemfile.lock

## Checklist

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -561,6 +561,7 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
+  arm64-darwin-23
   x86_64-darwin-19
   x86_64-darwin-20
   x86_64-darwin-21


Gemfile.lock
```diff
@@ -561,6 +561,7 @@ GEM
 PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
+  arm64-darwin-23
```
MazOneTwoOne, line 564: Silly question - I think I'm also on macOS Sonoma. How come mine doesn't show up/change the list of platforms? 

Or does it go by the first time I used the service and log my machine that way??

MazOneTwoOne (approved):

Gemfile.lock
starswan: no idea - it's very persistent on my machine after I do almost anything

PR #1242, opened 2024-02-08
EL-1308 spike to split out interpreting Cfe data from display

starswan:
https://dsdmoj.atlassian.net/browse/EL-1308

## What changed and why

Split out the 'parse and interpret Cfe result' from 'display result on results page' logic so that the CfeResult can be used (and extended) to allow early eligibility to check things like 'amount over the threshold' and 'gross_income_ineligible?' from the CfeResult without the formatting code currently in CalculationResult.

CalculationResult is currently in the models directory, looks (certainly now) more like a 'presenter' object that extracts common code to do with displaying (e.g consistent formatting for monetary values) 

## Checklist

### DIFF
--- a/app/models/calculation_result.rb
+++ b/app/models/calculation_result.rb
@@ -6,9 +6,9 @@ class CalculationResult
 
   attr_reader :level_of_help
 
-  delegate :decision, :calculated?, :has_partner?, :ineligible?, :capital_items,
+  delegate :decision, :calculated?, :has_partner?, :ineligible?,
            :raw_capital_contribution, :raw_income_contribution,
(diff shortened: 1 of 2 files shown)


willc-work (approved):
👍 looks good

app/models/calculation_result.rb
```diff
@@ -1,46 +1,27 @@
 class CalculationResult
```
MazOneTwoOne, line 1: query - should we change the name of this class to be `CalculateResultForView` or is that a bit much?

HettieS: Good spot, the name is not ideal especially now we have extracted out the logic that pulls out the data. But we're not sure what the name should be so we will revisit that later on.

PR #1243, opened 2024-02-09
EL-1308 Extract code for relevant step to send to CFE

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1308)

## What changed and why

Extracting some code to change the way we determine what is valid to send to CFE. This is to set groundwork for the upcoming early eligibility stuff.

## Guidance to review

Previously we have been using the same method (def valid_step?) to return steps both for navigation, and to determine what we send to CFE (in def relevant_form?). This has worked until now as we have built around the assumption that we will have a whole set of data to send to CFE at the end of the flow. With the development of early eligibility this is now changing and it is possible to send a lot less data to CFE, and to do this mid service rather than at the end of all the possible steps. 

We want to move away from using def valid_step? as a means of deciding what data should be sent, and look more at the actual data itself to see what is valid (what def relevant_step is doing). 

There is one place in def method_missing in the Check model where I haven't yet replaced valid_step? as I was unsure of the side effects at this point. 

## Checklist


spec/flows/steps/helper_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "rails_helper"
+
+RSpec.describe Steps::Helper do
+  describe "#relevant_step?" do
+    context "when there is asylum support data plus income data" do
```
MazOneTwoOne, line 5: question - do we need to specify this is a certificated check in the spec and doe we want a separate one for controlled?

HettieS: hmm possibly - I might add a few more specs to beef this file up 👍

starswan (approved):

spec/flows/steps/helper_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+              "income_frequency" => "monthly",
+              "gross_income" => 100,
+              "income_tax" => 20,
+              "national_insurance" => 3,
+            },
+          ],
+        }
+      end
+
+      it "returns the correct relevant step" do
```
willc-work, line 23: I assume this is required for coverage? hence the corresponding addition in en.yml.

config/locales/en.yml
```diff
@@ -477,6 +477,8 @@ en:
               not_a_number_when_many: Total money in bank account %{position} must be a number. Enter 0 if this does not apply.
               greater_than_or_equal_to: Total money in bank account must be 0 or more. Enter 0 if this does not apply.
               greater_than_or_equal_to_when_many: Total money in bank account %{position} must be 0 or more. Enter 0 if this does not apply.
+            account_in_dispute:
```
willc-work, line 480: please add a TODO to the related spec

spec/flows/steps/helper_spec.rb
HettieS: I didn't add these specs for the coverage so much, but just to check the new method was returning directly what I expected it to when the data was looking incohesive (thinking about changing answers). I wanted to test the method directly rather than with all the moving parts of a flow test

config/locales/en.yml
HettieS: I will double check but I think the translation error shows up in several specs with the new `def relevant_steps` method as we are now checking validity a lot more often to know which steps are relevant. And the custom method for bank account errors when it can't find inclusion for account in dispute (never a user facing error hence why there was no translation). If the spec can find the error then it happily passes over the form and returns it as invalid. I can double check which specs are affected and try and mark them? But it could be a few of them

willc-work: I did run it and this is the only test that breaks. it might be worth adding a todo so we review it as it sounds like it might be something that can be resolved in the test setup 🤷

spec/flows/steps/helper_spec.rb
willc-work: @HettieS I found the issue, essentially in ClientAssetsForm which calls BankAccountModel :smod_applicable is being returned as true. The simple fix is to add 
        immigration_or_asylum_type_upper_tribunal: "immigration_upper",
          asylum_support: false,
          
       to the test setup, just before you set empoloyment status. In the test setup you had assets was a valid step and smod was applicable. we just need smod to not be applicable because really we dont care at the point that EE is called.

PR #1249, opened 2024-02-15
EL-1341: Benefits suggestions list

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1341)

## What changed and why

- As per the relevant findings [for 2.1](https://uv3015-moj-ccq-legal-aid.uservisionaccessibility.co.uk/dropdown-menu-not-accessible-with-swipe-gestures-on-mobile-high.html)
  - have replaced `aria-owns` with `aria-controls`
  - removed `role="combobox"` from outer `div`, as attribute of `role="combobox"` is set in js for `input`
  - have added `.govuk_text_field` `div` to the input field so iPhone users can swipe onto the item
  - have added `aria-description` as per 'accessible autocomplete' [component](https://github.com/alphagov/accessible-autocomplete/blob/main/src/autocomplete.js) 
- As per the relevant findings [for 2.2](https://uv3015-moj-ccq-legal-aid.uservisionaccessibility.co.uk/dropdown-items-announced-as-not-selected-on-receiving-focus-medium.html)
  - have removed `.aria-selected="false"` attribute so "not selected" does not get read out by screen readers
 

## Checklist


app/views/question_flow/_benefit_section.html.slim
```diff
@@ -21,17 +20,19 @@
...
-      data-suggestions="suggestions"
-      data-module="govuk-input"
-      type="text"
-      name="benefit_model[items][#{counter}][benefit_type]"
-      value=model.benefit_type
-      id=id_pattern.gsub("ID", counter.to_s)
-      data-add-another-dynamic-elements="id:#{id_pattern},name:benefit_model[items][ID][benefit_type]"
-      class=("govuk-input--error" if model.errors.messages[:benefit_type].any?)
-    ]
+    .govuk_text_field
```
starswan, line 23: Think we need to add an aria-owns=1-type-suggestions here. JS sets an aria-owns on the input box parent which is redundant (apparently aria-owns is to communicate ownership when the relationship isn't specified in the DOM)

app/javascript/suggestions.js
```diff
@@ -102,7 +102,6 @@ Input.prototype.updateSuggestionsWithOptions = function (options) {
     li.textContent = option.textContent
     li.setAttribute('role', 'option')
     li.setAttribute('tabindex', '-1')
-    li.setAttribute('aria-selected', option.value === this.$module.value)
```
starswan, line 105: I wonder if we should comment this out instead with a comment. We now have 3 versions of this code, and it might be good to communicate which changes we have made to the original

app/views/question_flow/_benefit_section.html.slim
MazOneTwoOne: Have added `aria-owns="suggestions"`, which I think does the job and also removed adding `aria-owns` from the JS which was adding it to the parent node.

willc-work (approved):

PR #1258, opened 2024-02-20
EL-1375: Removing changes made to JS by previous implementation

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1375)

## What changed and why

- In this change we have re-moved the addition made to the JavaScript from the Crime Apply implementation, which we copied across into CCQ. 
  - It was trying to create a caption on a list element if this was true `(option.dataset.caption)`, and subsequently when the list was shown & list item selected, an additional condition was added to select the closet `li`, as it could have been focused on the `caption`. 
  - Neither of these conditions are relevant for CCQ and some people on the Edge browser were able to trigger the conditions and set an attribute to `null`

## Guidance to review

- CCQs JavaScript for the `input with suggestions` component is based on the [Crime Apply implementation](https://github.com/ministryofjustice/laa-apply-for-criminal-legal-aid/blob/main/app/javascript/local/suggestions.js)
- Crime Apply based their implementation on a GDS component which was [not merged into govuk-frontend](https://github.com/alphagov/govuk-frontend/pull/2453)

## Checklist


app/javascript/suggestions.js
```diff
@@ -37,7 +34,6 @@ Input.prototype.init = function () {
     this.$formGroup.appendChild(this.$suggestionsHeader)
     this.$formGroup.appendChild(this.$ul)
 
-    this.$module.removeAttribute('list') // [change: deactivate browser-native suggestions]
```
starswan, line 40: must admit hadn't spotted that one - this change is harmless as we don't support them, but should we? Presumably there are preventDefault() calls that would make this safe if we enabled the browser version of this?

MazOneTwoOne: I'm second guessing my rationale on this: 

- Based on my very basic understanding of how the [datalist](https://www.marioaraque.com/native-html5-autocomplete) element works, you would add the `list` attribute to the `input` field to get the browser-native suggestions.
- Because we don't add the `list` attribute to the `input` field, it didn't make sense to have `removeAttribute('list')`

I hadn't delved into the if we need to support them, in the first place. Please do correct me if I've mis-understood.

starswan: I think it might be one for Kristina

MazOneTwoOne: If that's the case, I start a slack conversation, to explore the the possibility of using the browser version's of the autocomplete/suggestion list? 

For the moment I'll add this bit back in so we can get the other bits merged.

starswan (approved):

PR #1261, opened 2024-02-21
EL-1377 dont try to send income or outgoings forms if they are invalid

starswan:
https://dsdmoj.atlassian.net/browse/EL-1377

Don't send income or outgoings forms to CFE if their contents is invalid

## Guidance to review

We had a crash on production due to trying to send an invalid outgoings form - https://ministryofjustice.sentry.io/issues/4999162546/?project=4504177640275968&query=&referrer=project-issue-stream.

It would seem sensible to add guards to prevent this scenario for occurring, hence this change - although its not exactly clear what the user did to cause this error 

## Checklist

### DIFF
--- a/app/services/cfe/regular_transactions_payload_service.rb
+++ b/app/services/cfe/regular_transactions_payload_service.rb
@@ -7,8 +7,8 @@ module Cfe
         relevant_form?(:housing_costs) ||
         relevant_form?(:benefit_details)
 
-      outgoings_form = instantiate_form(OutgoingsForm)
-      income_form = instantiate_form(OtherIncomeForm)
+      outgoings_form = instantiate_form(OutgoingsForm) if relevant_form? :outgoings
+      income_form = instantiate_form(OtherIncomeForm) if relevant_form? :other_income
       benefit_details_form = instantiate_form(BenefitDetailsForm) if relevant_form?(:benefit_details)
       housing_form = if relevant_form?(:mortgage_or_loan_payment)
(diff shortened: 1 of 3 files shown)


app/services/cfe/regular_transactions_payload_service.rb
```diff
@@ -7,8 +7,8 @@ def call
         relevant_form?(:housing_costs) ||
         relevant_form?(:benefit_details)
 
-      outgoings_form = instantiate_form(OutgoingsForm)
-      income_form = instantiate_form(OtherIncomeForm)
+      outgoings_form = instantiate_form(OutgoingsForm) if relevant_form? :outgoings
+      income_form = instantiate_form(OtherIncomeForm) if relevant_form? :other_income
```
MazOneTwoOne, line 11: Question: So if this is saying create an `outgoings_form` object(?), when calling CFE, if it exits - why do we need to worry about this now? Have we merged some early eligibility tickets laying the groundwork for the main early eligibility ticket to be mergd?

MazOneTwoOne (approved):
Approved with a question (not a blocker).

app/services/cfe/regular_transactions_payload_service.rb
starswan: Taking this code isolation, it requires only one of the forms to be valid to reach this point. Therefore not checking each of these here must be some sort of defect - which was shown out by the production crash. Are you suggesting that this isn't quite the right fix? [I think it becomes the right fix once early eligibility has been coded anyway]

PR #1268, opened 2024-02-26
EL-1352: Harmonise the versions of puppeteer that we use in our docker files

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1352)

## What changed and why

- using the latest Puppeteer version across dockerfiles 
- removing the need to download a specific version of chrome as [Puppeteer automatically downloads a recent version of Chrome for Testing](https://pptr.dev/#:~:text=Installation%E2%80%8B&text=When%20you%20install%20Puppeteer%2C%20it,guaranteed%20to%20work%20with%20Puppeteer.)

## Checklist

### DIFF
--- a/Dockerfile_browser_tools.dockerfile
+++ b/Dockerfile_browser_tools.dockerfile
@@ -10,7 +10,7 @@ RUN sudo add-apt-repository --yes ppa:malteworld/ppa
 RUN sudo apt install pdftk --allow-unauthenticated
 
 # Install Puppeteer via Yarn
-RUN yarn add puppeteer@22.3.0
+RUN yarn add puppeteer@22.1.0
 
 # Tell Puppeteer to skip installing Chromium. We'll be using the installed package.
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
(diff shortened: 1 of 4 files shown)


Dockerfile_browser_tools.dockerfile
```diff
@@ -10,10 +10,7 @@ RUN sudo add-apt-repository --yes ppa:malteworld/ppa
...
 
 # Install Puppeteer via Yarn
-RUN yarn add puppeteer@21.9.0
-
-# Install Chrome using Puppeteer command
-RUN npx puppeteer browsers install chrome@121
+RUN yarn add puppeteer@22.1.0
 
 # Tell Puppeteer to skip installing Chromium. We'll be using the installed package.
 ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
```
willc-work, line 16: I wonder if this needs to be above line 13 to have any effect.

willc-work: also it may not be  needed anymore if we are relying on puppetteer installing chromium

MazOneTwoOne: Thanks, have now removed. Will test if this works after sorting out the other Docker file.

PR #1278, opened 2024-03-01
El 1339 parallel specs

starswan:
https://dsdmoj.atlassian.net/browse/EL-1339

## What changed and why

Spike for parallel tests gem

## Checklist


.rspec_parallel
```diff
@@ -0,0 +1,6 @@
+--require spec_helper
```
MazOneTwoOne, line 1: L&D question - I don't quite understand how this get's invoked or why it's needed. It is mainly for formatting, helping with logs and tagging? Can't make sense of [the documentation](https://github.com/grosser/parallel_tests?tab=readme-ov-file#loggers).  

My only other thought is that this set-up help with the coverage report?

spec/system/feedback_spec.rb
```diff
@@ -23,12 +23,18 @@
         click_on "Continue to download and finish"
         expect(page).to have_content("Were you satisfied with this service?")
         click_on "Yes"
+        # TODO:  temp fix for parallel running
```
MazOneTwoOne, line 26: Question - is this a temp fix for when we get to [EL-1374](https://dsdmoj.atlassian.net/browse/EL-1374)?

MazOneTwoOne (approved):
Looks good. I'll update Jira ticket with more feedback as this is a spike. 

Also had one L&D question.

.rspec_parallel
starswan: This isn't actually required as all the tests require 'rails_helper' which includes this. I've taken it out. Its default for rspec, which is not rails-specific

spec/system/feedback_spec.rb
starswan: Yes, this is currently a problem anyway, but when running lots of processes (10 x browsers and 10 rspecs on my M1 machine by default) it happens so often that a clean build is almost impossible to create

PR #1287, opened 2024-03-06
EL-1385 EE results page

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1385)

show "not assessed" panels when the user has arrived at the results page from an early eligibility result.
ensure that the assessed panels show if they continue the journey and submit a full set of answers

## Checklist

### DIFF
--- a/app/controllers/results_controller.rb
+++ b/app/controllers/results_controller.rb
@@ -7,6 +7,7 @@ class ResultsController < ApplicationController
   end
 
   def show
+    @early_eligibility_selection = session_data.fetch("early_eligibility_selection", nil)
     @model = CalculationResult.new(session_data)
     track_completed_journey(@model)
     track_page_view(page: :view_results)
@@ -15,6 +16,7 @@ class ResultsController < ApplicationController
   end
 
   def download
+    @early_eligibility_selection = session_data.fetch("early_eligibility_selection", nil)
     track_page_view(page: :download_results)
     @model = CalculationResult.new(session_data)
     @sections = CheckAnswers::SectionListerService.call(session_data)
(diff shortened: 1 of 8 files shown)


MazOneTwoOne (changes requested):
Hey @willc-work I think the ticket also asked for this text "This was not assessed because it would not change the overall result" to be placed in the `app/views/results/_outgoings_table.html.slim` & `app/views/results/_capital_table.html.slim` when `early_result` is true

app/controllers/results_controller.rb
```diff
@@ -7,6 +7,7 @@ def create
   end
 
   def show
+    @early_eligibility_selection = session_data.fetch("early_eligibility_selection", nil)
     @model = CalculationResult.new(session_data)
     track_completed_journey(@model)
     track_page_view(page: :view_results)
```
MazOneTwoOne, line 13: I think we also need `@early_eligibility_selection = session_data.fetch("early_eligibility_selection", nil)` in the `download` method (line 20) so that the PDFs, can display the text.

app/views/results/_outgoings_table.html.slim
```diff
@@ -1,47 +1,48 @@
-= pdf_friendly_h2(t("results.show.client_outgoings"), "m", @is_pdf)
```
MazOneTwoOne, line 1: I think you've accidentally deleted `= pdf_friendly_h2(t("results.show.client_outgoings"), "m", @is_pdf)` - this should be put back in above line 4.

MazOneTwoOne (approved):
Looks good. 

Just a noted one change needed to help with PDF downloads. Noted one other thing which got accidently deleted in refactor.

app/views/results/_outgoings_table.html.slim
willc-work: excellent spot @MazOneTwoOne , no idea how I lost that line and the way that github was displaying the changes made it difficult to see that had happened ( for me anyway)

app/controllers/results_controller.rb
willc-work: fixed, thanks

PR #1301, opened 2024-03-14
EL-1344: Feedback, screen reader notification 

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1344)

## What changed and why

**Satisfaction feedback**
- ‘Thank you for your feedback’ is now announced by screen readers

**Freetext feedback**

Text shown when the following happens and announced by screen readers. 

- a user cancels and doesn’t send their feedback
  - Feedback not sent
- a user sends feedback, but the input box is empty
  - Feedback not sent, because you did not enter any feedback
- a user sends feedback
  - Thank you for your feedback

## Guidance to review

- The accessibility report and detail on this finding [can be viewed here.](https://uv3015-moj-ccq-legal-aid.uservisionaccessibility.co.uk/screen-reader-users-not-informed-of-feedback-update-low.html)

## Checklist


spec/system/feedback_spec.rb
```diff
@@ -23,10 +23,11 @@
         click_on "Continue to download and finish"
         expect(page).to have_content("Were you satisfied with this service?")
         click_on "Yes"
-        expect(page).to have_content("Your response has been sent, please tell us more")
+        assert page.has_content?("Your response has been sent, please tell us more")
+        assert page.has_content?("Don't include personal information")
```
starswan, line 27: This seems a curious change in style? expect(x).to have_content() should work fine

MazOneTwoOne: Thanks @starswan for the spot. This is me trying out some things, to stop the race between browser and server on this spec. I forgot to take them out when the didn't have much of an effect. I think I was following [this blog](https://nts.strzibny.name/avoid-sleep-rails-system-tests/). 

As an aside, there is this thing called [rspec-wait](https://github.com/laserlemon/rspec-wait), which we could use?

starswan: It's idiomatic to call find(<css-selector>) that waits a default amount of time before failing

starswan (approved):

PR #1308, opened 2024-03-18
Puppeteer 2250

starswan:
No Ticket

## What changed and why

Puppeteer still needs to be tweaked manually rather than upgraded with dependabot

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ executors:
   test-executor:
     resource_class: small
     docker:
-      - image: checkclientqualifiesdocker/circleci-image:main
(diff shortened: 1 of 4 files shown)


MazOneTwoOne (approved):
Question (not a blocker) - Do we need to add this to the Readme.md, as we might have to do this a couple of times a month? I can make a ticket, if required.

starswan:
> Question (not a blocker) - Do we need to add this to the Readme.md, as we might have to do this a couple of times a month? I can make a ticket, if required.

Hopefully this will all go away when we use the image to test and deploy from circle

PR #1323, opened 2024-03-27
EL-1439: Secondary conditional reveals announced by screenreaders

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1439)

## What changed and why

- Added in, visually hidden descriptions to radio buttons
- 'Add another' pattern, needed extra html layout and `aria-describedby` added in.

## Checklist


app/views/question_flow/forms/_outgoings.html.slim
```diff
@@ -1,4 +1,8 @@
 - partner ||= false
+
+- content_for :amount_and_frequency_hidden_description
+  = govuk_visually_hidden(t("question_flow.outgoings.hidden_description_amount_and_frequency"))
```
starswan, line 4: could this not be '.hidden_description_amount_and_frequency' rather than the full path?

starswan (approved):

app/views/question_flow/forms/_outgoings.html.slim
MazOneTwoOne: Hey @starswan, I've tried both ways and it looks like the full path is needed in this instance. 

I've changed it to this as the `i18n_key` is passed to this partial. 

> - content_for :amount_and_frequency_hidden_description
>   = govuk_visually_hidden(t("question_flow.#{i18n_key}.hidden_description_amount_and_frequency"))

MazOneTwoOne: Actually, it's much simpler to just have the full path. More than one view is using this partial. It's just better for it to be pointed to one place, rather than the same thins in different places.

PR #1349, opened 2024-04-17
EL-1417: Use new typography scale from design system

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1417)

## What changed and why

- Use new typography scale from design system

## Checklist


psweeting1 (approved):
LGTM

MazOneTwoOne:
> This might not be relevant - do we need to consider our maintenance page and error pages? @MazOneTwoOne

I think it is relevant. We should add this comment to the JIRA ticket so we can make another ticket out of it. My understanding is that this current ticket is just to turn on the feature flag and see what comes out of it.

PR #1402, opened 2024-05-31
EL-1540 use default values for feature flags

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1540)

When a check is created without a particular session feature flag we now use the default for that flag.
Moved some methods out of the enabled? method as the session_flags method was also using `enabled?` but didnt need all of the functionality. So 'overrideable_flag' and 'env_var_flag' were extracted to be used by both `session_flags` and `enabled?`
This allows us to revert to using the ENV_VAR to set the sessions flags at the start of the check, and also in the tests where we turn the flags on and off around specific tests. Elsewhere it will use the default for the flag.

## Checklist

### DIFF
--- a/app/lib/feature_flags.rb
+++ b/app/lib/feature_flags.rb
@@ -1,19 +1,24 @@
 class FeatureFlags
+  # public: false is a marker that this is a test flag
+  # all flags intended to go into production should have public: set to true
   ENABLED_AFTER_DATE = {
     example_2125_flag: { from: "2125-01-01", public: false },
-    mtr_accelerated: { from: "2024-06-24", public: false },
+    mtr_accelerated: { from: "2024-06-24", public: true },
   }.freeze
 
   # the values of some feature flags will come from the session and not the env variables.
   # "global" - feature flag value should be derived from the env variable
   # "session" - feature flag value should be derived from the session_data of the check
   STATIC_FLAGS = {
+    # example and example2 are just here for tests and should not be used in code.
     example: { type: "session", default: false },
+    example2: { type: "session", default: true },
     sentry: { type: "global", default: false },
     index_production: { type: "global", default: false },
     maintenance_mode: { type: "global", default: false },
     basic_authentication: { type: "global", default: false },
     early_eligibility: { type: "session", default: false },
+    legacy_assets_no_reveal: { type: "session", default: false },
   }.freeze
 
   class << self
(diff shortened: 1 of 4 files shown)


HettieS (approved):
I have approved with a comment to rename a test. I found the new complexity quite hard to follow - can we please add some information to the README on the purpose of the new default setting? I would quite like that to include a hierarchy of where the flag value will come from (or a few written use cases in the README to display it clearly)

spec/models/feature_flags_spec.rb
```diff
@@ -69,6 +69,34 @@
...
+        ENV["EXAMPLE_FEATURE_FLAG"] = "enabled"
+        each.run
+        ENV["EXAMPLE_FEATURE_FLAG"] = "disabled"
+      end
+
+      it "uses the ENV_VAR to set the feature flag" do
+        expect(described_class.session_flags.fetch("example")).to eq true
+      end
+
+      context "with ff" do
```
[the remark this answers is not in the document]
willc-work: oh sorry about this, it was a placeholder until we worked out what the test should be doing. Will fix.

app/lib/feature_flags.rb
```diff
@@ -22,19 +27,24 @@ def enabled?(flag, session_data = nil, without_session_data: false)
...
 
       if STATIC_FLAGS.key?(flag)
-        ENV["#{flag.to_s.upcase}_FEATURE_FLAG"]&.casecmp("enabled")&.zero? || STATIC_FLAGS.fetch(flag).fetch(:default)
+        # if the flag is a valid flag and of type `session` but does not
+        # exist in the session_data we return the default value for that flag
+        if STATIC_FLAGS[flag][:type] == "session"
+          STATIC_FLAGS[flag][:default]
+        else
+          # if it is a global flag we check the env_var
+          # if there is no env_var specified we use the default value for that flag
```
HettieS, line 45: I'm not sure if this is a bit scary or not. Why would we use the default value if there is no env_var? Could we risk accidentally turning something on?

HettieS: In fact maybe I am misunderstanding here. Is this setting purely for testing? Where we might want to have a default but there is no env var? I guess my fear is accidentally leaving a `default: true` and forgetting to set an env_var in prod. Maybe I am overthinking it. Maybe a comment would help here.

app/lib/feature_flags.rb
```diff
@@ -22,19 +27,24 @@ def enabled?(flag, session_data = nil, without_session_data: false)
...
-      end
+      possible_override = overrideable_flag(flag)
+      return possible_override unless possible_override.nil?
 
       if STATIC_FLAGS.key?(flag)
-        ENV["#{flag.to_s.upcase}_FEATURE_FLAG"]&.casecmp("enabled")&.zero? || STATIC_FLAGS.fetch(flag).fetch(:default)
+        # if the flag is a valid flag and of type `session` but does not
+        # exist in the session_data we return the default value for that flag
+        if STATIC_FLAGS[flag][:type] == "session"
+          STATIC_FLAGS[flag][:default]
```
HettieS, line 42: so do we never look at the env_var anymore for session flags (when it is not in session_data/there is no session_data)? We only ever use the default?

HettieS: How would we then manage session flags being switched on or off across multiple envs? Perhaps I am misunderstanding.

app/lib/feature_flags.rb
willc-work: Yes this is only ever expected to fire during testing when we provide an env_var in the before block and no session_data.
The defaults are also for tests only, they are added so that when a flag doesn't have a value we give it a value that will not break our tests.
I will add more to the comment to explain the env_var in tests.

app/lib/feature_flags.rb
willc-work: >never look at the env_var anymore for session flags
correct. If a check starts without a new feature and we enable the env_var to release the feature, we would be passing 'enabled' to the check, which would then break.
so the only time we should check the env_var (for session flags) is when the check starts using the method `session_flags`

willc-work: Setting env_vars across multiple envs, or even mutiple UATs will work because the env_var is used to set the initial state of the check being undertaken. These flags are then set on a per check basis and not a per environment depending on the configuration of the deployment at the time the check started, so we can change env_vars between environments or even between checks on the same environment, and the check will use the inital configuration as its flags.

HettieS: Yeah I think I'm overthinking it. I was thinking, is there a situation where a session flag might not have session data passed in...(where we set `without_session_data: true`) but tbh I don't _think_ those should/would be session flags. 

I'm thinking for places like the start page where we don't have a session. But I wonder if we might have grouped a bunch changes together under a session flag, because they're all part of the same feature, but not pass the session in, in places where we don't have it. Probably overthinking it but we will just need to be a bit careful with our flag setup...

HettieS: It feels like a very edgy case nasty bug lol

app/lib/feature_flags.rb
willc-work: i added a description as to what the default does rather than rename the attribute

PR #1448, opened 2024-07-10
EL-1572 Tidy ups

starswan:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1572)

## What changed and why

EL-1572's spike has uncovered a handful of simplifications to the mainline codebase. These are contained here to keep them apart from the spike itself

## Checklist

### DIFF
--- a/app/views/checks/_boolean_money_with_frequency.slim
+++ b/app/views/checks/_boolean_money_with_frequency.slim
@@ -3,11 +3,10 @@
     = label_text
 
   dd.govuk-summary-list__value
-    - if conditional
-      div class="govuk-!-margin-bottom-1"
-        = yes_no_boolean(boolean)
+    div class="govuk-!-margin-bottom-1"
+      = yes_no_boolean(boolean)
 
-    - if !conditional || boolean
+    - if boolean
       - if value.nil?
         = t("generic.not_applicable")
       - else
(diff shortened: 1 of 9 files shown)


app/views/checks/_boolean_money_with_frequency.slim
```diff
@@ -1,14 +1,12 @@
...
 
   dd.govuk-summary-list__value
-    - if conditional
-      div class="govuk-!-margin-bottom-1"
-        = yes_no_boolean(boolean)
+    div class="govuk-!-margin-bottom-1"
+      = yes_no_boolean(boolean)
 
-    - if !conditional || boolean
+    - if boolean
```
HettieS, line 9: Just looking at this - wondering if we should rename 'boolean' as my understanding here is it is more about the relevancy of the answer i.e. if they say 'yes', then the value is relevant. This did confuse me a bit. I might be misunderstanding the flow. Maybe this will all be refactored to be clearer once we have removed all the conditional behaviour.

HettieS (approved):
Checking this PR was a little confusing because I think some of the code might still be written in a way to work with the old conditional reveals feature flagged stuff! Even though I think we've removed those flags now. Perhaps there is a bit more tidy up to do. Even just taking out the word 'conditional' from some of the attributes e.g. 'friends_and_family_conditional_value' which I think is a hangover from that.

app/views/checks/_boolean_money_with_frequency.slim
starswan: Have implemented something based on your suggestion - the name 'boolean_money_with_frequency' didn't mean much to me either

PR #1450, opened 2024-07-11
EL-1614: Portal Metrics

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1614)

## What changed and why

- add attribute to `CompletedUserJourney`
- add a way to build attribute
- call JourneyLoggerService from ResultsController, check the signed in at this stage
- update `Providers` module to save data correctly
- call `JourneyLoggerService` with the parameters in the correct order
- run a migration so that we can add new column to `completed_user_journeys` & `providers` table
- update specs

## Checklist


spec/features/provider_login_spec.rb
```diff
@@ -9,7 +9,7 @@
     let(:email_address) { Faker::Internet.email }
 
     before do
-      sign_in create(:provider, email: email_address)
+      sign_in create(:provider, email: email_address, first_office_code: "LAA")
```
willc-work, line 12: This is very minor but can we change the 'LAA' code to something like 1Q630KL

willc-work (approved):
minor change requested but not a blocker

spec/features/provider_login_spec.rb
MazOneTwoOne: Agreed, have updated.

PR #1470, opened 2024-07-26
BUG: Patch rails admin data types to fix admin panel

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1659)

## What changed and why

Bug in the Rails Admin panel - it doesn't accept field type timestamptz...which seems to be the form it is now receiving the timestamp data in. We haven't changed anything much so not too sure why. I wonder if it is to do with RDS upgrade but not too sure.

I'm hoping this work around fixes the issue as documented here https://stackoverflow.com/questions/75858914/rails-admin-3-unsupported-field-datatype-timestamptz

I'm not sure if this will work or if will still need tinkering but need to merge it to see.

## Checklist


lib/rails_admin/config/fields/types/timestamptz.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# :nocov:
```
starswan, line 3: I think the :nocov: labels say that this doesn't quite work. I think it needs explicitly requiring somehow.

HettieS: How do the :nocov: labels mean that? 

I did wonder about the explicit requiring. There is another file I can add which does require the individual field types in Rails Admin - one person who had implemented this fix said it wasn't needed to work on their end....but perhaps I'll add it

starswan (approved):

PR #1491, opened 2024-08-09
EL-1685 Remove Early eligibility flag

starswan:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1685)

## What changed and why

Remove early eligibility flag

## Checklist


spec/features/invalid_session_spec.rb
```diff
@@ -2,16 +2,6 @@
 
 RSpec.describe "Invalid session", :throws_cfe_error, type: :feature do
   it "tells me if CfeService thinks my answers are invalid" do
-    allow(CfeService).to receive(:call).and_raise(Cfe::InvalidSessionError.new(DomesticAbuseApplicantForm.new))
```
psweeting1, line 5: Just curious, why was this test removed. Doesn't appear to be related to the feature flag stuff.

spec/rails_helper.rb
```diff
@@ -100,6 +100,10 @@
     )
   end
 
+  config.before(:each, :might_call_cfe) do
```
psweeting1, line 103: How come you can't use the `stub_cfe_calls`. Why create a `might_call_cfe` stub? 

Also `might_call_cfe` isn't a clear and very confident name, something that describes it what it does would be better.

If they are both necessary then can we make it clear how they are different from each other and used.

starswan: Agreed I couldn't think of a good name. I wanted to convey that this is to do with early eligibility - thought of that over the weekend. Will find a better name

psweeting1:
Recording Slack thread into github

**Patrick:** 
```
Hey Stephen,

Thanks for looking at my comment and making the changes. I just wanted to double check something.
Just because I didn't see a comment in the thread in regards to the `stub_cfe_calls` reference I made. Did you spot the `stub_cfe_calls` test I flagged? Because it just seemed weird having two when they both produce the same result in regards to the stubbed `calls_cfe_early_returns_not_ineligible` Unless I understand it wrong.
```

**Stephen:** 
```They don't produce quite the same result. stub_cfe_calls does look a bit weird, I will take another look.```

starswan:
> Recording Slack thread into github
> 
> **Patrick:**
> 
> ```
> Hey Stephen,
> 
> Thanks for looking at my comment and making the changes. I just wanted to double check something.
> Just because I didn't see a comment in the thread in regards to the `stub_cfe_calls` reference I made. Did you spot the `stub_cfe_calls` test I flagged? Because it just seemed weird having two when they both produce the same result in regards to the stubbed `calls_cfe_early_returns_not_ineligible` Unless I understand it wrong.
> ```
> 
> **Stephen:** `They don't produce quite the same result. stub_cfe_calls does look a bit weird, I will take another look.`

I have now renamed stub_cfe_calls to stub_cfe_calls_with_webmock - some tests need to be webmock focussed, whereas the new ones are less complex and they just want to ignore the CFE call for EE

psweeting1 (approved):
LGTM

PR #1497, opened 2024-08-15
EL-1669 CFE Payloads

starswan:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1669)

## Checklist

### DIFF
--- a/app/models/check.rb
+++ b/app/models/check.rb
@@ -29,7 +29,7 @@ class Check
                   else
                     attribute
                   end
-    form_class.from_session(session_data).send(method_name)
+    form_class.model_from_session(session_data).send(method_name)
   end
 
   def respond_to_missing?(attribute, include_private = false)
--- a/app/models/check.rb
+++ b/app/models/check.rb
@@ -156,10 +156,6 @@ class Check
     Steps::Logic.owns_property_with_mortgage_or_loan?(session_data)
   end
 
-  # def owns_property_outright?
-  #   Steps::Logic.owns_property_outright?(session_data)
-  # end
-
   def under_18_controlled_clr?
     Steps::Logic.controlled_clr?(session_data)
   end
@@ -227,9 +223,4 @@ class Check
   def partner_owns_additional_property?
     Steps::Logic.partner_owns_additional_property?(session_data)
   end
-
-  # def has_partner_assets?
-  #   Steps::Logic.partner?(session_data) &&
-  #     session_data.key?("partner_investments_relevant")
-  # end
 end
--- a/app/models/check.rb
+++ b/app/models/check.rb
@@ -147,4 +147,89 @@ class Check
       Steps::Helper.consistent?(session_data)
     end
   end
+
+  def owns_vehicle?
+    Steps::Logic.owns_vehicle?(session_data)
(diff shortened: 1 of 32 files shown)


app/models/check.rb
```diff
@@ -147,4 +147,72 @@ def consistent?
...
+  def has_outgoings?
+    session_data.key?("childcare_payments_conditional_value")
+  end
+
+  def any_benefits?
+    (benefits || []).any?
+  end
+
+  def has_other_income?
+    session_data.key?("friends_or_family_conditional_value")
```
willc-work, line 208: Just for my own understanding. This works because we check that the key `friends_or_family_conditional_value` exists in the session_data, and it will only exist if the page has been visited, regardless of whether they have entered yes or no to the friends_or_family question. So the check could be for `maintenance_in_conditional_value` and achieve the same result.

app/models/check.rb
```diff
@@ -147,4 +147,89 @@ def consistent?
...
+  end
+
+  def owns_property_with_mortgage_or_loan?
+    Steps::Logic.owns_property_with_mortgage_or_loan?(session_data)
+  end
+
+  # def owns_property_outright?
+  #   Steps::Logic.owns_property_outright?(session_data)
+  # end
+
```
willc-work, line 162: still needs to be removed

app/models/check.rb
```diff
@@ -147,4 +147,89 @@ def consistent?
...
+  end
+
+  def partner_owns_additional_property?
+    Steps::Logic.partner_owns_additional_property?(session_data)
+  end
+
+  # def has_partner_assets?
+  #   Steps::Logic.partner?(session_data) &&
+  #     session_data.key?("partner_investments_relevant")
+  # end
```
willc-work, line 234: lets also remove this commented out method

app/services/cfe/partner_payload_service.rb
```diff
@@ -26,37 +26,40 @@ def partner
...
+        benefits_form = instantiate_form(PartnerBenefitDetailsForm) if check.has_partner_benefits?
+        CfeParamBuilders::RegularTransactions.call(income_form, outgoings_form, benefits_form)
+      else
+        []
+      end
     end
 
     def additional_properties
-      return [] unless relevant_form?(:partner_additional_property_details)
+      # return [] unless completed_form?(:partner_additional_property_details)
```
willc-work, line 61: we can delete this line instead of commenting it out

spec/services/cfe/vehicle_payload_service_spec.rb
```diff
@@ -56,10 +55,10 @@
           "asylum_support" => true,
         }
       end
-      let(:relevant_steps) { [:asylum_support] }
+      let(:completed_steps) { [:asylum_support] }
```
willc-work, line 58: I dont think this renamed let is used by any tests so we can delete it

willc-work (approved):
Some deletions required for commented out code and one for an unused let in a spec file. nothing that would block though.

willc-work:
Decided not to merge this refactor becuause it was only a partial solution to removing the reliance on completed_steps

PR #1498, opened 2024-08-15
EL-1669 data access tidy up

starswan:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1669)

## What changed and why

Use model_from_session rather than from_session where a model is required (from_session adds an 'extra' model for form display). 

Renamed relevant_form? to completed_form? as it better reflects the desired result (i.e that the data is actually present from form x)

## Checklist


willc-work (approved):
One question but its not a blocker, happy to discuss

PR #1537, opened 2024-09-11
EL-1730 Add Feature Flag for CW Form Updates

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1730)

## What changed and why

Added a global feature flag for the updated CW forms. The ticket does not specify session or global, and I think when we release this we want everyone to use the new form regardless of when the session was started. This will not break anything as the data will exist in the session for the new inputs on the forms and if the data does not exist then we handle that in the forms themselves anyway.

## Checklist


HettieS (approved):
approved but did you mean to set UAT to NOT_ENABLED?

PR #1540, opened 2024-09-12
EL-1719 New English CW1 form (other income fields)

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1719)

## What changed and why

Added new PDF
Added new mapping file with new methods

## Guidance to review

added a nocov on the method used to determine the filename suffix, this will be amended for each new CW file and then removed when the feature is released

## Checklist


app/lib/controlled_work_mappings/cw1_new.yml
```diff
@@ -0,0 +1,441 @@
...
+    - name: otherincome_client
+      type: text
+      source: client_other_income
+
+    # Benefits & Tax Credits / Client
+    - name: benefits_client
+      type: text
+      source: client_benefits
+
+    # Other income / Client
```
MazOneTwoOne, line 303: I think `undefined_50` no longer exists

app/lib/controlled_work_mappings/cw1_new.yml
```diff
@@ -0,0 +1,441 @@
...
+    - name: undefined_50
```
MazOneTwoOne, line 304: I think `undefined_50` no longer exists

app/lib/controlled_work_mappings/cw1_new.yml
```diff
@@ -0,0 +1,441 @@
...
+      type: text
+      source: client_property_lodger_income
+
+    # Other Income / Client
+    - name: otherincome_client
+      type: text
+      source: client_other_income
+
+    # Benefits & Tax Credits / Client
+    - name: benefits_client
```
MazOneTwoOne, line 299: Could we move `benefits_client` to be above `maintenance_client` so it follows the layout of the PDF - just makes it easier to read?

app/lib/controlled_work_mappings/cw1_new.yml
```diff
@@ -0,0 +1,441 @@
...
+
+    # "The client is a child under the age of 18:"
+    - name: Child under 18
+      type: boolean_radio
+      source: under_eighteen?
+      yes_value: "Yes_2"
+      no_value: "No"
+
+    # "The child’s means are to be aggregated with a maintaining adult."
+    - name: Means to be aggregated with maintainng adult
```
MazOneTwoOne, line 20: 'maintaining' instead of 'maintainng'

app/models/concerns/income_summarisable.rb
```diff
@@ -39,6 +39,34 @@ def combined_disposable_income
     from_cfe_payload("result_summary.disposable_income.combined_total_disposable_income")
   end
 
+  def client_benefits
+    from_cfe_payload("assessment.gross_income.state_benefits.monthly_equivalents.all_sources.value")
```
MazOneTwoOne, line 43: Question - are we checking the `assessment` instead of the `result_summary`, because the `result_summary` amalgamates the specific values we need?

app/models/concerns/income_summarisable.rb
```diff
@@ -39,6 +39,34 @@ def combined_disposable_income
...
+  end
+
+  def client_maintenance_in
```
MazOneTwoOne, line 46: Question - what does the `_in` mean?

app/lib/controlled_work_mappings/cw1_new.yml
```diff
@@ -0,0 +1,441 @@
...
+
+    # "The client is directly or indirectly in receipt of S.4 or S.95 Asylum Support (Immigration and Asylum category work only):"
+    - name: Client in receipt of asylum support
+      type: boolean_radio
+      source: asylum_support?
+      yes_value: "Yes_2"
+      no_value: "No"
+
+    # "The client has a partner whose means are to be aggregated:"
+    - name: Has partner whose means are to be agrgregated
```
MazOneTwoOne, line 48: 'aggregated' instead of `agrgregated`

app/lib/controlled_work_mappings/cw1_new.yml
```diff
@@ -0,0 +1,441 @@
...
+- section: partner_income
+  fields:
+
+    # Gross monthly earnings / Partner
+    - name: undefined_49
+      type: text
+      source: partner_gross_income
+
+    # Other income / Partner
+    - name: undefined_51
```
MazOneTwoOne, line 394: I think we've missed the partner income mappings.

`undefined_51` no longer exists, we instead need to add in:
- benefits_partner
- maintenance_partner
- pensions_partner
- studentfinance_partner
- friendsandfamily_partner
- property_partner
- otherincome_partner

app/services/controlled_work_document_population_service.rb
```diff
@@ -20,6 +23,7 @@ def call(session_data, model)
 
     TEMPLATES = {
       "cw1" => "lib/cw1-form.pdf",
+      "cw1_new" => "lib/cw1-form-new.pdf",
```
MazOneTwoOne, line 26: Should we call this just `cw1` and the old one `cw1_legacy`?

spec/services/controlled_work_document_value_mapping_service_spec.rb
```diff
@@ -54,6 +54,31 @@
       expect(result).to include(representative_sample)
     end
 
+    it "can successfully populate the new fields on the CW1 form (other income section)" do
```
MazOneTwoOne, line 57: Will need to update tests with partner income questions.

spec/services/controlled_work_document_value_mapping_service_spec.rb
```diff
@@ -54,6 +54,31 @@
...
+        "Means test required" => "Yes_2", # This is always checked as CCQ is only relevant to means tested cases
+        "Passported" => "No", # Not passporting
+        "Client in receipt of asylum support" => "No", # Asylum supported not given
+        "Please complete Part A Capital Subject matter of dispute" => "No_4", # No SMOD
+        "Has partner whose means are to be agrgregated" => "Yes_3", # Has a partner
+        "undefined_26" => "250,000", # Property worth £250,000
+        "undefined_42" => "555", # Valuables
+        "undefined_40" => "222", # Investments
+        "undefined_38" => "111", # Savings
+        "undefined_30" => "25", # Percentage owned
```
MazOneTwoOne, line 71: you've also left in "# Percentage owned"

app/models/concerns/income_summarisable.rb
```diff
@@ -39,6 +39,34 @@ def combined_disposable_income
...
+    from_cfe_payload("assessment.gross_income.other_income.monthly_equivalents.all_sources.maintenance_in.value")
+  end
+
+  def client_pensions
```
MazOneTwoOne, line 50: When I run this locally, I get `Float does not have #dig method` error. 

So am assuming that this `dig` method can't be used on a float??


>   def from_cfe_payload(path)
>     session_data.dig("api_response", *path.split("."))
>   end
>

app/models/concerns/income_summarisable.rb
willc-work: exactly. the assessment section returns the monthly equivalents which are what we need to diaplay, the result section only shows aggregated income figures

spec/services/controlled_work_document_value_mapping_service_spec.rb
willc-work: removed the comments.

app/models/concerns/income_summarisable.rb
willc-work: fixed this, there was an issue with calling 'value'

app/models/concerns/income_summarisable.rb
MazOneTwoOne: Thanks for explaining - I thought as much, but was struggling to figure out how I would confim it.

app/lib/controlled_work_mappings/cw1_new.yml
```diff
@@ -0,0 +1,472 @@
...
+
+- section: partner_income
+  fields:
+
+    # Gross monthly earnings / Partner
+    - name: undefined_49
+      type: text
+      source: partner_gross_income
+
+#    # Other income / Partner
```
MazOneTwoOne, line 423: Should we get rid of this instead of commenting out?

app/lib/controlled_work_mappings/cw1_new.yml
MazOneTwoOne: still showing wrong spelling?

app/lib/controlled_work_mappings/cw1_new.yml
MazOneTwoOne: still showing wrong spelling?

app/lib/controlled_work_mappings/cw1_new.yml
willc-work: this is a spelling error in the PDF and as it is the name of the field we have to reproduce it exactly the same way here.

PR #1541, opened 2024-09-12
EL-1699: Early eligibility banner design

HettieS:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1699)

## What changed and why

Adding an early eligibility banner to replace our current page design

## Guidance to review

- back button from Results and metrics tracking to be done on separate tickets
- back button hopefully set up nicely with access to @previous_step in results controller
- have noted a couple points in comments where the metrics need to change/need potentially adding

## Checklist


app/controllers/change_answers_controller.rb
```diff
@@ -9,34 +10,21 @@ def update
...
-              redirect_to_next_question
-            end
-          else
-            save_and_redirect_to_check_answers
-          end
+      if FeatureFlags.enabled?(:ee_banner, session_data)
+        next_step = step_with_inconsistent_data
+        if next_step
+          redirect_to helpers.check_step_path_from_step(next_step, assessment_code)
+        elsif Steps::Helper.last_step_in_group?(session_data, step)
```
HettieS, line 17: might not need this branch anymore

willc-work: thus removes so much complexity 👏

app/models/cfe_result.rb
```diff
@@ -6,7 +6,11 @@ def initialize(api_response)
   end
 
   def ineligible_gross_income?
-    api_response.dig(:result_summary, :gross_income, :proceeding_types).first[:result] == "ineligible"
+    early_gross_income_result == "ineligible"
+  end
+
+  def early_gross_income_result
+    api_response.dig(:result_summary, :gross_income, :proceeding_types, 0, :result)
```
willc-work, line 13: how are we assuring ourselves that this is indeed an early result? This dig would return a value at any time it is called.

app/controllers/forms_controller.rb
```diff
@@ -31,4 +36,15 @@ def calculate_next_step(session_data, step)
       Steps::Helper.next_step_for(session_data, step)
     end
   end
+
+  def calculate_early_result
+    # for other groupings, we'll be able to implement like below (disposable, partner gross)
+    # we could also consider a check at each page in the gross income section
+    if last_tag_in_group?(:gross_income)
+      cfe_result = CfeService.result(session_data, Steps::Helper.completed_steps_for(session_data, step))
+      session_data["early_result"] = { "result" => cfe_result.early_gross_income_result,
```
willc-work, line 45: we are currently storing in the session:
`early_eligibility_selection` which has a value of `nil`, `gross` or `continued_check`
do you think we need to store this value with this new banner method?

app/views/results/_early_summary.html.slim
```diff
@@ -0,0 +1,45 @@
+- data = @model.summary_data(section)
```
HettieS, line 1: i added a new early summary section for now as looking at the other was making my head spin...but this way of doing it isn't very sustainable for a check at disposable too - thinking we can make a ticket to refactor our summary section when we start doing disposable early checks too maybe?

app/models/cfe_result.rb
HettieS: yeah you are right, it is badly named because it is actually just a gross income result!

app/controllers/results_controller.rb
```diff
@@ -6,9 +6,18 @@ def create
     redirect_to result_path(assessment_code:)
   end
 
+  def early_result
+    # hard coding this at the moment - we could look for last step with valid data but that might cause issues if
+    # banner displayed on multiple pages OR we can send the step down in the param?
+    session_data["api_response"] = CfeService.call(session_data, Steps::Helper.completed_steps_for(session_data, :other_income))
+    redirect_to result_path(assessment_code:, early_assessment: :gross_income)
```
HettieS, line 13: I was thinking I would rather pass something down the params but could also add to the session - it just means we need to deal with the clean up if a user goes back.

HettieS: changed functionality

app/controllers/change_answers_controller.rb
HettieS: confirmed that this branch is still needed for now

app/models/cfe_result.rb
HettieS: renamed this method

app/controllers/forms_controller.rb
HettieS: So i have actually added a new variable with a slightly different name, that does the same thing - but we don't grab it until later in the flow, in the results controller. I wasn't sure about this naming now we have got rid of the form. I've introduced a new variable called `@early_assessment`.

app/views/results/_early_summary.html.slim
willc-work: sounds like a good midway point. we dont know when disposable work will be done, better to get this feature delivered than trying to future proof it too early

app/controllers/results_controller.rb
```diff
@@ -6,16 +6,26 @@ def create
...
 
+  def early_result
+    @previous_step = params[:step].to_sym
+    session_data["api_response"] = CfeService.call(session_data, Steps::Helper.completed_steps_for(session_data, @previous_step))
+    set_early_result_type
+    redirect_to result_path(assessment_code:)
+  end
+
   def show
+    @early_assessment = session_data.dig("early_result", "type")
```
willc-work, line 17: can we rename this to early_result_type given that it is returning `gross`, and i assume in the future it will also be bale to return `disposable`.

I just think using early_assessment and early_result is confusing

app/controllers/results_controller.rb
```diff
@@ -6,16 +6,26 @@ def create
     redirect_to result_path(assessment_code:)
   end
 
+  def early_result
```
willc-work, line 9: perhaps this should be display_early_result or show_early_result

app/controllers/checks_controller.rb
```diff
@@ -32,6 +33,10 @@ def assessment_code
     params[:assessment_code]
   end
 
+  def clear_early_result
+    session_data["early_result"]&.clear
```
HettieS, line 37: this should always exist so going to remove the nil guarding

app/views/results/_capital_table.html.slim
```diff
@@ -1,4 +1,4 @@
-- if @early_eligibility_selection == "gross"
+- if @early_eligibility_selection == "gross" || @early_assessment == "gross_income"
```
willc-work, line 1: might be worth adding a comment here as when we take out the EE banner Feature Flag this may be missed. It wont be a huge issue but this:
`@early_eligibility_selection == "gross"`
could get missed.

app/views/results/_outgoings_table.html.slim
```diff
@@ -1,4 +1,4 @@
-- if @early_eligibility_selection == "gross"
+- if @early_eligibility_selection == "gross" || @early_assessment == "gross_income"
```
willc-work, line 1: same as above, a comment to remind us to remove it, we could add the flag as a comment so when searching this will be returned

willc-work (commented):
A few naming changes requested and some comments for future reference
Have not approved it as I would like a 2nd dev to review, but once the requested changes are made i am happy for it to be approved

app/controllers/results_controller.rb
HettieS: Because this method is a redirect on to the `def show`, I'm not sure about those names. This is kind of replicating what the full result method does above, where it intercepts to send off the CFE call and alter the session_data, then sends on to `def show`. So `display` and `show` might not be quite right - what about something like `def early_result_redirect`?

willc-work: yes that makes more sense

app/controllers/results_controller.rb
HettieS: yes that name is much clearer!

app/controllers/checks_controller.rb
HettieS: of course this actually doesn't always exist for checks that don't pass through income, doh!

app/controllers/results_controller.rb
HettieS: now changed :)

app/views/results/_capital_table.html.slim
HettieS: I've added in comments to the places that reference this variable so hopefully the clean up should be easier!

HettieS:
> A few naming changes requested and some comments for future reference Have not approved it as I would like a 2nd dev to review, but once the requested changes are made i am happy for it to be approved

Thanks @willc-work, the changes you requested should all be made now, let me know if you spot anything else. I've mentioned to the devs that we'd like a second review on this too

willc-work:
I am happy to approve this PR however as discussed previously, as it is a large and significant piece of work I would like a 2nd (and/oe 3rd opinion) before we merge. Although I think in the meantime we can put it into UAT to start testing.

MazOneTwoOne (commented):
Just leaving some comments, but this looks good.

app/controllers/change_answers_controller.rb
```diff
@@ -9,37 +9,21 @@ def update
...
-              session_data.delete IneligibleGrossIncomeForm::SELECTION
-              flash[:notice] = I18n.t("service.change_eligibility")
-              redirect_to_next_question
-            end
-          else
-            save_and_redirect_to_check_answers
-          end
+      if FeatureFlags.enabled?(:ee_banner, session_data)
+        next_step = step_with_inconsistent_data
+        if next_step && step != :aggregated_means
```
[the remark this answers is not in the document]
HettieS: Good spot! Yes I'll add that comment in!

app/controllers/checks_controller.rb
```diff
@@ -32,6 +33,11 @@ def assessment_code
     params[:assessment_code]
   end
 
+  def clear_early_result
+    # not all checks have an early result e.g. passported clients, U18 etc
+    session_data["early_result"]&.clear
+  end
```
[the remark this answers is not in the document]
HettieS: So the idea here was, if a user gets to the Check Answers page, even if they have had an early (ineligible) result previously, we just want to forget about it. This is because they've given us the information for a full check, and we want to make sure we treat their check, even if they go back and forth and do weird stuff, as a full check. This is to try and avoid any complexity associated with a previous partial check.

In relation to this bullet, `clear_early_result will trigger when early_result key exists and is nil/null` - def clear_early_result will clear the early result information in the session data, if it exists. If it doesn't exist, it'll just quietly fail as we know there are some situations where it won't exist.

Currently none of this new code triggers metrics tracking - my initial thoughts on this were, kick off an analytics event when we get the early ineligible result (probably inside def calculate_early_result or somewhere like that) - this event will save the current session data in the JSONB field. 

If they do a full check, I think it's okay to remove the early result from the session data as we have other ways to link a full check to an early ineligible check (assessment_id). Does that make sense?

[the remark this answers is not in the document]
MazOneTwoOne: Ok, sounds good. 

I need to think about this properly. I'm struggling to work out why we want to ever forget the information  in `early_result` when the user does a full check. This is a 'me' problem not the 'PR' problem. 😅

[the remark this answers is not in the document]
HettieS: No that's okay! I did um and ah about adding this in. But ultimately I'm trying to protect against weird user scenarios where they go back and forth like:
- user gets early ineligible result
- goes to results page
- goes 'back'
- decides to continue to enter all data
- gets to Check Answers
- submits

in this scenario they would see the reduced version of the results page, where capital, disposable aren't rendered. But they would have entered the info. It felt like just clearing the early result stuff on Check Answers was the simplest way to avoid weird journeys like this.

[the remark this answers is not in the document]
HettieS: (this is the problem with setting stuff in session_data - you have to clean up after yourself properly or get weird behaviour)

PR #1547, opened 2024-09-18
EL-1721 New civmeans7 form (other income fields)

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1721)

## What changed and why

Add new PDF and mapping file
update the specs
update the generate_form_key method to simplify 

## Checklist


app/lib/controlled_work_mappings/civ_means_7_new.yml
```diff
@@ -0,0 +1,457 @@
...
+      yes_value: "Yes"
+      no_value: "No"
+
+    # "The client's assets (held in sole name or jointly held) have been claimed by the opponent"
+    - name: Assets claimed by opponent
+      type: boolean_radio
+      source: smod_assets?
+      yes_value: "Yes"
+      no_value: "No"
+- section: smod_capital
```
HettieS, line 55: maybe add a space between the two sections for readability like you have for the other sections

app/lib/controlled_work_mappings/civ_means_7_new.yml
```diff
@@ -0,0 +1,457 @@
...
+    # Valuables (SMOD)
+    - name: FillText55
+      type: text
+      source: smod_valuables
+
+    # Total capital part A (SMOD)
+    - name: FillText101
+      type: text
+      source: smod_total_capital
+- section: capital
```
HettieS, line 136: space here too

app/lib/controlled_work_mappings/civ_means_7_new.yml
```diff
@@ -0,0 +1,457 @@
...
+      type: text
+      source: client_property_lodger_income
+
+    # Other Income / Client
+    - name: otherincome_client
+      type: text
+      source: client_other_income
+
+    # Benefits & Tax Credits / Partner
+    - name: benefits_partner
```
HettieS, line 293: I'm wondering if these partner fields should go in the partner income section (line 402) but to be honest I'm not sure!

spec/factories/session.rb
```diff
@@ -83,19 +83,6 @@
       national_insurance { 112 }
     end
 
-    trait :with_other_income do
```
HettieS, line 86: is there a reason why you have deleted this trait?

HettieS (approved):
approved with small comments!

spec/factories/session.rb
willc-work: I had used it then removed it, then found it was not in use, the one place it was being used in a spec it wasn't actually required for the spec. So I figured it should be removed if it wasnt being used and it didnt warrant its own ticket or PR.

app/lib/controlled_work_mappings/civ_means_7_new.yml
willc-work: I like grouping them this way as they are next to eachother on the CW form, but I understand why it might be better to move them. I will mull it over a bit.

PR #1550, opened 2024-09-20
EL-1745 Check Answers display

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1745)

## What changed and why

Added a check to see if we are responding to an EE gross income result
If we are we display only relevant sections on Check ANswers

## Checklist

### DIFF
--- a/app/services/check_answers/section_lister_service.rb
+++ b/app/services/check_answers/section_lister_service.rb
@@ -26,15 +26,22 @@ module CheckAnswers
           else
             pre = [
               Sections::Dependants.new(check),
-              Sections::ClientIncome.new(check),
-            ]
-            post = [Sections::Outgoings.new(check)]
+              # Simplecov is showing that check.under_eighteen_assets? is not being covered by a test
+              # we have tests in spec/flows/under_eighteen_flow_spec.rb that check with and without under_eighteen_assets
+              # :nocov:
+              (Sections::ClientIncome.new(check) if !check.under_eighteen? || check.aggregated_means? || check.under_eighteen_regular_income? || check.under_eighteen_assets?),
+              # :nocov:
+            ].compact
+
+            post = check.gross_early_eligibility_exit? ? [] : [Sections::Outgoings.new(check)]
             finance_sections = if check.partner?
-                                 pre + [Sections::PartnerIncome.new(check)] + post
+                                 pre + (check.gross_early_eligibility_exit? ? [] : [Sections::PartnerIncome.new(check)]) + post
                                else
                                  pre + post
                                end
-            non_finance_sections + finance_sections + assets
+            assets_sections = check.gross_early_eligibility_exit? ? [] : assets
+
+            non_finance_sections + finance_sections + assets_sections
           end
         end
       end
--- a/spec/flows/early_result_spec.rb
+++ b/spec/flows/early_result_spec.rb
@@ -33,6 +33,19 @@ RSpec.describe "Early result journey", type: :feature do
       fill_in_other_income_screen_with_friends_and_family
       fill_in_ineligible_gross_income_screen(choice: "Skip remaining questions")
       confirm_screen("check_answers")
+      expect(all(".govuk-summary-card__title").map(&:text))
+        .to eq(
+          ["Client age",
+           "Partner and passporting",
+           "Level of help",
+           "Type of matter",
+           "Type of immigration or asylum matter",
+           "Number of dependants",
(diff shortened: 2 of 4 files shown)


app/services/check_answers/section_lister_service.rb
```diff
@@ -28,13 +28,15 @@ def call(session_data)
               Sections::Dependants.new(check),
               Sections::ClientIncome.new(check),
             ]
-            post = [Sections::Outgoings.new(check)]
+            post = check.gross_early_eligibility_exit? ? [] : [Sections::Outgoings.new(check)]
             finance_sections = if check.partner?
                                  pre + [Sections::PartnerIncome.new(check)] + post
```
MazOneTwoOne, line 33: I think we also have to put in a ternary operator here, for the partner income i.e

`pre + (check.gross_early_eligibility_exit? ? [] : [Sections::PartnerIncome.new(check)]) + post`

We are showing these screens even though user did not fill those questions in.

MazOneTwoOne (commented):
Just the one query. 

Also, do we need to put in a bunch of flow tests?

app/services/check_answers/section_lister_service.rb
willc-work: thanks good spot, I think I was just expecting to see income questions.

willc-work:
@MazOneTwoOne re [this comment](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1550#pullrequestreview-2318622289) I am not keen on the idea of adding a bunch of tests to check what displays given that UATing it will confirm that for us without the ongoing overhead of running 3 or 4 tests to cover the variations needed for this bug. If the bug was more serious then tests may be required.

MazOneTwoOne:
> @MazOneTwoOne re [this comment](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1550#pullrequestreview-2318622289) I am not keen on the idea of adding a bunch of tests to check what displays given that UATing it will confirm that for us without the ongoing overhead of running 3 or 4 tests to cover the variations needed for this bug. If the bug was more serious then tests may be required.

Maybe not a bunch of tests, but at least we should update this spec `spec/flows/early_result_spec.rb` to cover the new behaviour (technically the existing behaviour we should have had)?

It may stop us adding in a similar regression.

MazOneTwoOne (approved):

app/services/check_answers/section_lister_service.rb
```diff
@@ -26,15 +26,22 @@ def call(session_data)
...
             pre = [
               Sections::Dependants.new(check),
-              Sections::ClientIncome.new(check),
-            ]
-            post = [Sections::Outgoings.new(check)]
+              # Simplecov is showing that check.under_eighteen_assets? is not being covered by a test
+              # we have tests in spec/flows/under_eighteen_flow_spec.rb that check with and without under_eighteen_assets
+              # :nocov:
+              (Sections::ClientIncome.new(check) if !check.under_eighteen? || check.aggregated_means? || check.under_eighteen_regular_income? || check.under_eighteen_assets?),
+              # :nocov:
```
MazOneTwoOne, line 33: @willc-work this feels risky 🤔

spec/flows/early_result_spec.rb
```diff
@@ -32,8 +32,47 @@
...
+    #            "Type of matter",
+    #            "Type of immigration or asylum matter",
+    #            "Number of dependants",
+    #            "Employment status",
+    #            "Client employment income 1",
+    #            "Client benefits",
+    #            "Client other income"],
+    #           )
+    # end
+
```
MazOneTwoOne, line 75: @willc-work did we mean to leave this commented out?

app/services/check_answers/section_lister_service.rb
willc-work: we have tests that cover both sides of under 18 assets and the associated views, so I think i have mitigated the risk as well as possible without spending more time understanding why simplecov isnt counting this as coverage

spec/flows/early_result_spec.rb
willc-work: removed now., this was to test if adding income values triggered the coverage, it doesnt

app/services/check_answers/section_lister_service.rb
MazOneTwoOne: Happy to leave like this. But maybe we make a separate ticket for this, so this doesn't get lost?

PR #1593, opened 2024-10-11
EL-1757: Saving Early Eligibility result to `completed_user_journey` table

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1757)

## What changed and why

- add in the values to be populated by `journey_logger_service`
- call `JourneyLoggerService` when user see banner. Also set "type" of early result, when they see the banner 
- remove when user see `set_early_result_type` from `ResultsController` as this is now set upon seeing the banner.
- tidy up naming and separate out methods 
- test that `JourneyLoggerService` puts correct things into table based on session_data
- test that a `track_completed_journey` method in my ResultsController is called based on session_data
- add suffix to assessment_id for an early result 
- add tests for code coverage

## Guidance to review

- n/a

## Checklist


app/models/check.rb
```diff
@@ -191,4 +191,8 @@ def early_ineligible_result?
   def gross_income_excess
     session_data.dig("early_result", "gross_income_excess")
   end
+
+  def is_early_result_type_gross_income?
```
willc-work, line 195: we can remove the 'is' prefix from the method name. The '?' at the end indicates this is a boolean and is the idiomatic ruby way to do things

app/services/journey_logger_service.rb
```diff
@@ -55,5 +56,12 @@ def matter_type(check)
         check.immigration_or_asylum_type
       end
     end
+
+    def early_result_type(check)
+      # Potentilly we can have more conditions here for "partner_gross" & "disposable"
```
willc-work, line 61: I agree, but we dont need this comment to be retained.

app/controllers/forms_controller.rb
```diff
@@ -41,8 +41,17 @@ def calculate_early_result
     if last_tag_in_group?(:gross_income)
       cfe_result = CfeService.result(session_data, Steps::Helper.completed_steps_for(session_data, step))
       session_data["early_result"] = { "result" => cfe_result.gross_income_result,
-                                       "gross_income_excess" => cfe_result.gross_income_excess }
-      # add analytics tracking here? completed user journeys
+                                       "gross_income_excess" => cfe_result.gross_income_excess,
+                                       "type" => "gross_income" } # I think we have to set this here so we have it in the session, when calling the `JourneyLoggerService`
+      track_completed_journey_for_early_result
```
HettieS, line 46: at the moment, this is also going to add an entry for an eligible result - we only want to track ineligible ones as a completed user journey. So we need to make sure we only save the ineligible ones.

app/controllers/forms_controller.rb
```diff
@@ -41,8 +41,17 @@ def calculate_early_result
...
     end
   end
+
+  def track_completed_journey_for_early_result
```
HettieS, line 50: have you thought about/tried passing down the CFE result stored in `cfe_result` to the tracking method? That might prevent the need to make the CFE call again. Variables are a sort of storage - once they've been set, they store that information.

So you'd do something like track_early_result_completed_journey(cfe_result) and then use/manipulate cfe_result to get what you need.

app/controllers/forms_controller.rb
```diff
@@ -41,8 +41,17 @@ def calculate_early_result
...
+    session_data["api_response"] = CfeService.call(session_data, Steps::Helper.completed_steps_for(session_data, step))
+    calculation_result = CalculationResult.new(session_data)
+    office_code = signed_in? && current_provider.present? ? current_provider.first_office_code : nil
+    assessment_id_with_early_result_suffix = "#{assessment_id}_early_result" # This is so we do not overrite newly created early result entry, if an user decides to carry on with check or change answers
```
HettieS, line 54: to be honest I'm not sold yet on using a suffix, but I can't really say exactly why, it just feels really hacky lol - have you dug into whether the original tracking method could be extended to have some conditionals/guarding in to prevent early results from updating?

app/controllers/results_controller.rb
```diff
@@ -55,15 +53,11 @@ def load_check
...
-    if signed_in? && current_provider.present?
-      JourneyLoggerService.call(assessment_id, calculation_result, @check, current_provider.first_office_code, cookies)
-    else
-      JourneyLoggerService.call(assessment_id, calculation_result, @check, nil, cookies)
-    end
+    office_code = signed_in? && current_provider.present? ? current_provider.first_office_code : nil
+
+    # We have the method `clear_early_result` on the ChecksController, so if a user goes back or ever see the check/change answers screen this will create a new entry into CompletedUserJourney
+    # The intention is that we are not counting that as a continuation an early ineligible result but as a new journey
+    JourneyLoggerService.call(assessment_id, calculation_result, @check, office_code, cookies) unless @check.early_ineligible_result?
```
HettieS, line 61: you could introduce the condition early on like 20 so that this method is never executed for an early result ie. 

track_completed_journey(@model) unless @check.early_ineligible_result?

I think this makes the condition a bit less hidden and ever so slightly saves on computing time lol.

You may need to move the @check instance variable further up the show methods if you go this way.

HettieS (commented):
Good work! I have left some comments for potential improvements and happy to pair on them!

app/controllers/forms_controller.rb
MazOneTwoOne: Yes, have tried passing down the CFE result stored in `cfe_result` to the tracking method. 

I think the issue is that they are two different method calls to the `CfeService`. I need the `.call` in `track_completed_journey_for_early_result` instead of `.result`

app/controllers/forms_controller.rb
MazOneTwoOne: I did dig into this, but very happy to pair on making this bit in the `JourneyLoggerService` guard for early results

>       CompletedUserJourney.transaction do
>         if (journey = CompletedUserJourney.find_by(assessment_id:))
>           journey.update!(attributes)
>         else
>           CompletedUserJourney.create!(attributes.merge(assessment_id:))
>         end
>       end

What I tried checking was if there was an `assessment_id:` which had `early_result_type` as not `nil`. The problem I ran into, when I tried that, was some weird thing with caching of what to put into `CompletedUserJourney`. Lets pair on this if you're happy to.

Having the suffix means that, as per the ticket, we always create an entry into `CompletedUserJourney` when there is an early ineligible result and never update it. 

I had the opposite vibe of this being simpler to understand rather then delving into more if conditionals. 🤣

MazOneTwoOne: I had this monstrosity which gave me bad vibes:

>       CompletedUserJourney.transaction do
>         early_ineligible_result_journey = CompletedUserJourney.find_by(assessment_id:) && CompletedUserJourney.where(assessment_id:).where.not(early_result_type: nil).exists?
>         if (journey = CompletedUserJourney.find_by(assessment_id:))
>           journey.update!(attributes) unless early_ineligible_result_journey
>         else
>           CompletedUserJourney.create!(attributes.merge(assessment_id:))
>         end
>       end

HettieS: Maz and I paired on doing this an alternative way without the suffix but we ran into other issues - firstly turns out the DB field has a uniqueness constraint on assessment_id, which would mean we'd have to change that if we wanted to store it potentially more than once. Also, having a more complex tracking event that is for both early tracking and end of service tracking creates issues when a user goes back and forth and triggers tracking multiple times. Suffix seems safest option.

HettieS: Ok actually we could do something like this:

```
 CompletedUserJourney.transaction do
        scope = CompletedUserJourney.where(assessment_id:)

        journey = if check.early_result_type.present?
                    scope.where(early_result_type: check.early_result_type).first
                  else
                    scope.where(early_result_type: nil).first
                  end

        if journey
          journey.update!(attributes)
        else
          CompletedUserJourney.create!(attributes.merge(assessment_id:))
        end
      end
```

MazOneTwoOne: We get 'duplicate key value violations' if we go with the above instead the suffix.

I think we're in the unavoidable loop of having an existing row in `CompletedUserJourney` and then trying to create a new one which shares the same `assessment_id`. 

It's a sensible constraint of the schema, solved by the suffix.

app/controllers/forms_controller.rb
```diff
@@ -41,8 +41,19 @@ def calculate_early_result
...
+        track_completed_journey_for_early_result
+      end
     end
   end
+
+  def track_completed_journey_for_early_result
+    session_data["api_response"] = CfeService.call(session_data, Steps::Helper.completed_steps_for(session_data, step))
+    calculation_result = CalculationResult.new(session_data)
+    office_code = signed_in? && current_provider.present? ? current_provider.first_office_code : nil
+    assessment_id_with_early_result_suffix = "#{assessment_id}_early_result" # This is so we do not overrite newly created early result entry, if an user decides to carry on with check or change answers
```
HettieS, line 56: with the addition of the new early result boolean field, we no longer plan to use the suffix to differentiate between early/full checks.

Instead we'll start allowing a check to be saved more than once i.e. take the unique constraint off the assessment_id column in the CompletedUserJourneys table.

Then the update action will change to use a combination of the assessment code and the early result boolean to know if it's updating an existing row, or creating a new one. So if it's an early check, the active record query should look to update an existing row for the assessment id and early_eligibility_result: true. If it's not an early check, it'll look for assessment id and early_eligibility_result: false. If neither of those exist it'll create the new row.

HettieS (commented):
There are now a couple extra things to do on this PR since we added the new field, I've left a comment but they are:
- db migration to remove the uniqueness constraint on assessment_id on CUJ table
- update the table correctly when updating an early result, or a full check (see my other comment)

app/controllers/forms_controller.rb
MazOneTwoOne: I'm a bit confused. Should I be working on this ticket until that migration? It still seems blocked.

HettieS: So the plan that we agreed at dev time was that the migration would be done on this ticket as it is an integral part to saving the early result. Are you happy to do that?

HettieS: If for any reason it becomes complicated and adds too much complexity to this ticket (hopefully it is straight forward but that is always famous last words) then let us know and we can look at splitting the work out again

app/services/journey_logger_service.rb
```diff
@@ -5,8 +5,13 @@ def call(assessment_id, calculation_result, check, portal_user_office_code, cook
 
       attributes = build_attributes(calculation_result, check, portal_user_office_code)
       CompletedUserJourney.transaction do
-        if (journey = CompletedUserJourney.find_by(assessment_id:))
-          journey.update!(attributes)
+        existing_journey_with_early_check = CompletedUserJourney.find_by(assessment_id:, early_eligibility_result: true)
```
HettieS, line 8: this scoping looks good, but there are a couple more things to make this work:
- we need to make sure we only look to update an existing early check if the check itself is an early check, you can use the check object to work this out 
- we need to scope the existing full checks with `early_ineligible_result: false` as well as the assessment id. find_by will just find the first instance of something, so it might find the wrong result if we don't add the additional scope.

Oherwise this looks like it is on the right track!

app/services/journey_logger_service.rb
```diff
@@ -5,8 +5,17 @@ def call(assessment_id, calculation_result, check, portal_user_office_code, cook
 
       attributes = build_attributes(calculation_result, check, portal_user_office_code)
       CompletedUserJourney.transaction do
-        if (journey = CompletedUserJourney.find_by(assessment_id:))
-          journey.update!(attributes)
+        existing_journey = CompletedUserJourney.find_by(assessment_id:)
+
+        if check.early_ineligible_result?.present?
```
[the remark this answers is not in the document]
HettieS: If I'm being completely honest - what I would probably look to do in the interest of making this more extendable is have a new service, EarlyJourneyLoggerService or something like that, which only deals with the early journey types, and either a base class or a module that shares the shared functionality. Perhaps we can chat about this tomorrow, or we could agree to implement something like that when we introduce further early check types! Be good to get @willc-work thoughts too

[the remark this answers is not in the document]
HettieS: It also might be useful to start to think about introducing some scopes into the application on the CompletedUserJourney model to make the code a little more readable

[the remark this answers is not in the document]
HettieS: Sorry I didn't mean to confuse things any further - let's forget about adding shared modules etc for now and go with modifying the existing JourneyLoggerService. I just wanted to make sure we've considered the options and make something we are happy to extend for further functionality. I think I was rubber ducking a bit last night on this thread lol.

app/controllers/results_controller.rb
```diff
@@ -18,8 +17,11 @@ def show
     @early_result_type = session_data.dig("early_result", "type")
     @early_eligibility_selection = session_data.fetch("early_eligibility_selection", nil)
     @model = CalculationResult.new(session_data)
-    # we'll need to move this tracking point or do something with it
-    track_completed_journey(@model)
+
+    # We have the method `clear_early_result` on the ChecksController, so if a user goes back or ever see the check/change answers screen this will create a new entry into CompletedUserJourney
```
willc-work, line 21: i think we can lose this comment.

app/services/journey_logger_service.rb
```diff
@@ -5,8 +5,11 @@ def call(assessment_id, calculation_result, check, portal_user_office_code, cook
 
       attributes = build_attributes(calculation_result, check, portal_user_office_code)
       CompletedUserJourney.transaction do
-        if (journey = CompletedUserJourney.find_by(assessment_id:))
-          journey.update!(attributes)
+        early_result_type = check.early_ineligible_result?
+        existing_journey = CompletedUserJourney.find_by(assessment_id:, early_eligibility_result: early_result_type || nil)
```
HettieS, line 9: This looks really nice and compact - I would probably change the variable name of early_result_type as it is not quite describing what that really returns. Also I'm not 100% sure the `|| nil` is quite right - I think it will default the value to nil if the early result type is false as well. I think I can see why you've done it to protect updating results correctly pre migration?

Maybe it would be wise to have a feature flag here so we don't change the analytics over until we switch on the new functionality, which should hopefully be quite easy to add but might affect your tests a bit!

The other thing I'd love to see is a couple more tests around the update part - in JourneyLoggerServiceSpec there are some tests at the bottom around line 163:

 ```
context "when details change" do
      let(:session_data) { { partner: false } }

      it "updates an existing record" do
        existing_record = FactoryBot.create(:completed_user_journey, assessment_id:, partner: true)
        described_class.call(assessment_id, calculation_result, check, portal_user_office_code, {})
        expect(existing_record.reload.partner).to be false
      end
    end
```

It would be great if we can add some checks here to make sure when we update an existing early result, we update the right thing successfully - and also a test when there is an early and a full result, we update the right result successfully. Does that make sense?

File: spec/features/provider_login_spec.rb [hunk 12]
```diff
@@ -42,4 +42,52 @@
...
+      expect(page).to have_content("Gross monthly income limit exceeded")
+      fill_in_outgoings_screen
+      expect(page).not_to have_content("Gross monthly income limit exceeded")
+      fill_in_forms_until(:check_answers)
+      click_on "Submit"
+      expect(page).to have_current_path(/\A\/check-result/)
+      expect(page).to have_content "Your client's key eligibility totals"
+    end
+
+    it "when I go straight to results the check" do
```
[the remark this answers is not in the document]
MazOneTwoOne: This wording is taken from the `spec/flows/early_result_spec.rb` so I'm inclined to keep it.

spec/services/journey_logger_service_spec.rb
```diff
@@ -206,5 +206,67 @@
...
+        expect(output.asylum_support).to be false
+        expect(output.partner).to be false
+        expect(output.passported).to be false
+        expect(output.outcome).to eq "ineligible"
+        expect(output.early_result_type).to eq "gross_income"
+        expect(output.early_eligibility_result).to be true
+      end
+    end
+
+    context "when details change in a full journey", :ee_banner do
```
willc-work, line 249: I dont think these 3 tests are actually testing the change that we have made, although it runs over the code, it is creating a new record but without focussing on changing the `early_eligibility_result`. I have added a test that mimics what we do in the service and shows how we are creating 2 records for the same `assessment_id`. Which I think can replace these 3 contexts.

willc-work (commented):
I have added a commit with a refactor of  `CompletedUserJourney.transaction`, it incorporates the feature flag to ensure that existing cases and the existing journey are unaffected and I think this is a much clearer way to structure the code.

spec/services/journey_logger_service_spec.rb
MazOneTwoOne: My only question on the added test:

- do we not need to add some session data, before we make the initial call, somewhere before line 234?
- do we want to check the output of what we think should be in `CompletedUserJourney`

willc-work (approved):
Approving this. We have tests covering the different journeys we expect. I have also manually tested and check the DB so that an updated EE result that remains ineligible updates the same row and continuing to the end creates a new row with the same id.

PR #1594, opened 2024-10-11
EL-1725 EE back button 

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1725)

## What changed and why

change back button behaviour from results page when they arrive there from an ee result.

## Checklist


app/views/results/show.html.slim
```diff
@@ -5,7 +5,10 @@
     = t(@check.controlled? ? ".controlled_qualifies" : ".likely_to_qualify")
 
 - content_for :back do
-  = link_to t("generic.back"), check_answers_path(assessment_code: params[:assessment_code]), class: "govuk-back-link"
+  - if FeatureFlags.enabled?(:ee_banner, @check.session_data) && @early_result_type == "gross_income"
```
HettieS, line 8: looks good - might not be necessary at the min but to future proof this for other early result type checks, you could just check for the presence of `early_result_type` (rather than tying it to gross_income specifically), as the behaviour of going back to the previous step should be the same for all early results

HettieS: and early_result_type should only exist if they never saw 'Check Answers' (the place where we clear it out from the session)

spec/flows/early_result_spec.rb
```diff
@@ -109,6 +109,18 @@
...
       end
 
+      it "when I go straight to results and use the back button it shows the banner" do
+        confirm_screen("outgoings")
+        outgoings_url = current_path
+        expect(page).to have_content("Gross monthly income limit exceeded")
+        click_on "Go to results page"
+        expect(page).to have_current_path(/\A\/check-result/)
+        expect(page).to have_content "Your client's key eligibility totals"
+        visit outgoings_url # simulates using the back button to return to outgoings
```
HettieS, line 119: I see why you've done this - I wonder if we can add a test that really checks the back button is working as expected? Perhaps an end2end test might work? But not too sure

HettieS (approved):
Approved but would love to see if we can add another test to really verify the back button goes to the right place

app/views/results/show.html.slim
HettieS: ooooh actually it's wouldn't exist for short checks e.g. passported! So maybe this is the best way for now, or maybe the condition is check for early result and check it isn't a 'short check'...will leave to you!

willc-work: I have changed it to:
`if FeatureFlags.enabled?(:ee_banner, @check.session_data) && @early_result_type`

spec/flows/early_result_spec.rb
willc-work: I did spend some time trying to get the back button to work, essentially as we move from ` def early_result_redirect` to `def show` in `results_controller.rb` we lose the ` @previous_step = params[:step].to_sym` value. This doesnt matter to the back button as the browser handles it and so does our back link.

To get the test to work we would need to add code that would only be there for the test to work, which is bad coding.

Having looked at the other uses of this pattern I think it is comething we have come up against on previous occassions and then this workaround has been adopted. So I think a lot of developer time has already been spent on trying to fix this issue and I dont think it is that important an issue. 

The current test checks that on returm to the page we still see the banner. We can manually test that it works as part of UAT and I think we can assure ourselves that the functionality is there without having to cover it with an automated test.

PR #1604, opened 2024-10-16
EL-1781 Rails update to 7.2.1

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1782)

## What changed and why

update rails version to 7.2.1
add ORM specific databasse cleaner gem because the original gem is no longer compatible with rails >=7.2
the `alias_attribute` method has been partly deprecated, unfortunately the part that was deprecated was the feature we were using. Replaced with `alias_method`

https://guides.rubyonrails.org/7_2_release_notes.html
from changelog:
Remove deprecated support to call alias_attribute with non-existent attribute names.

## Checklist


app/forms/additional_property_details_form.rb
```diff
@@ -7,7 +7,8 @@ class AdditionalPropertyDetailsForm
   ITEMS_SESSION_KEY = "additional_properties".freeze
   ITEM_MODEL = AdditionalPropertyModel
   ATTRIBUTES = %i[additional_properties].freeze
-  alias_attribute :additional_properties, :items
+  alias_method :additional_properties, :items
+  alias_method :additional_properties=, :items=
```
MazOneTwoOne, line 11: This took me quite a bit of reading to understand. 

Just leaving this here in case it comes in handy in the future. They may look to add in [alias_association](https://github.com/rails/rails/pull/49801) into a future version of rails, which might be useful.

MazOneTwoOne (approved):

PR #1625, opened 2024-10-31
EL-1833 EE data migraton 

willc-work:
add production deploy approval so staging can be tested before we do the migration in production

[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1833)

## What changed and why

added a new gem to handle data_migrations.
This is our first data_migration and although we don't expect to have many of these, they have a way of sneaking up on you as the db is modified. So it makes sense to add this now as you cant retrospectively move migrations.

The migration runs on my machine ;) 
I tested it on `early_eligibility_selection` does not exist, is `gross` and is `continue_check`

It will not do anything on UAT as the branch is new and will have no data. it will work on main UAT when it is merged. As such I have added 2 approval steps, one for STG and one for PROD. This way we can merge to main and then test each namespace before and after the migration.

The data_migration uses `:with_data` to do the data migrations after the migrations are run. I have added this where I think it needs to be, but this is the bit I am least confident about. So the pipeline will again be the test bed for this.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ references:
       name: Database Setup
       command: |
         bundle exec rake db:create db:schema:load
-        bundle exec rake db:migrate
+        bundle exec rake db:migrate:with_data
 
   setup_parallel_database: &setup_parallel_database
     run:
@@ -323,14 +323,14 @@ workflows:
 # again in the future, the requires production_deploy_approval
 # in Deploy to Production is also commented out
 
-#      - production_deploy_approval:
-#          type: approval
-#          requires:
-#            - Deploy to Staging
-#          filters:
-#            branches:
-#              only:
-#                - main
+      - production_deploy_approval:
+          type: approval
+          requires:
+            - Deploy to Staging
+          filters:
+            branches:
+              only:
+                - main
       - deploy:
           name: Deploy to Production
           environment: ccq-production
@@ -341,7 +341,7 @@ workflows:
             - linters
             - build_and_push
             - Deploy to Staging
-#            - production_deploy_approval
+            - production_deploy_approval
           post-steps:
             - jira/notify:
                 job_type: deployment
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -307,6 +307,7 @@ workflows:
             - linters
             - build_and_push
             - Deploy to UAT
+            - staging_deploy_approval
           post-steps:
             - jira/notify:
                 job_type: deployment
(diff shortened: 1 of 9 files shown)


db/data/20241031143616_populate_early_eligibility_result_in_completed_user_journeys.rb
```diff
@@ -0,0 +1,9 @@
+class PopulateEarlyEligibilityResultInCompletedUserJourneys < ActiveRecord::Migration[7.2]
+  def up
```
HettieS, line 2: This looks good - I think it might be wise to add a guard clause that scopes the batch of data to 'things without an early_eligibility_result' i.e. it is nil. This is because if this migration is applied again at any point, it won't attempt to look at all the data, just the relevant nil ones. 

My second thought was - maybe instead of a data migration, we should be doing a rake task. My thinking is:
- depending on when we apply this migration, if it doesn't sync up with switching on the new functionality (and it won't with the way we have the PRs at the minute) we'll end up continuing to collect data 'the old way' and we'd need to deal with those again after we switch the feature on
- if it is a rake task, it's more controlled. We can run it when we choose, add logging to keep track, and make sure it executes properly. We can run it more than once which helps with the above situation if we run it before we switch the functionality over. If the pod dies or it fails for some reason, we can run it again.
- it means we don't need to worry about adding the new data migration gem at this point, and we could add/investigate at our leisure

let me know what you think!

.circleci/config.yml
```diff
@@ -322,15 +323,23 @@ workflows:
 # Keeping this step commented out because we will need
 # again in the future, the requires production_deploy_approval
 # in Deploy to Production is also commented out
+      - staging_deploy_approval:
```
HettieS, line 326: if we did it as a rake task we can also move the pipeline approval steps

db/data/20241031143616_populate_early_eligibility_result_in_completed_user_journeys.rb
HettieS: If we do choose to go the migration route - I think we need to extract it from this PR and actually do it after everything is released so we don't need to update data twice

willc-work: I'm going to cogitate on this. I understand the reasons behind a rake task but I dont like the idea of having to ssh into prod to run it (and possibly having to do it more than once) as opposed to using a data migration that is run via rails standard deployments. However I agree that if thats the case we need to run the data_migration after the new eebanner is deployed to production so it is only run once.

willc-work:
closing this as the work has been split into multiple PRs

PR #1627, opened 2024-10-31
EL-1763 replace cw1&2 form

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1763)

## What changed and why

updated the CW1&2 form (english)

## Checklist


HettieS (commented):
So I ran the download locally and on the deployed UAT env and it completely broke at the download point :')

It is strangely still looking for the old form so I don't know if we've missed something....

Screenshot 2024-11-01 at 10 33 16

willc-work:
> So I ran the download locally and on the deployed UAT env and it completely broke at the download point :')
> 
> It is strangely still looking for the old form so I don't know if we've missed something....
> 
> Screenshot 2024-11-01 at 10 33 16

should be fixed now, tested locally and redeployed to UAT

HettieS (approved):
Works a charm

PR #1628, opened 2024-10-31
EL-1842 dependabot

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1842)

## What changed and why

There are 5 updates in this PR. The govuk-frontend update breaks 3 tests in `spec/system/add_another_spec.rb`
with this error:
   `http://127.0.0.1:46359/assets/application-6d3e90a9.js 16684:14 Uncaught InitError: govuk-radios: Root element (`$root`) already initialised`

which I think relates to this line:
`initAll();`
in application.js, essentially during tests we are doing something when we add another that causes the JS to reload already loaded elements whichgovuk frontend does not like.

The code changes I was making to try and get this to work involved changes to `add-another.js`, `application.js` and possibly `webpack-config.js` (I had a rethink before I started to alter a 3rd JS file).

I tested add another on bank_accounts and they work as expected and no errors arr visible in the console, so it looks like this is only something that appears in tests, hence I have elecrted to suppress this warning in `rails_helper` instead of fixing the root cause.

If someone more _au fait_ with JS knows a better way to handle this let me know.

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "private": "true",
   "dependencies": {
     "@ministryofjustice/frontend": "^3.0.1",
-    "@rails/ujs": "^7.1.402",
+    "@rails/ujs": "^7.1.400",
     "@sentry/browser": "8.35.0",
     "esbuild": "^0.24.0",
     "govuk-frontend": "^5.6.0",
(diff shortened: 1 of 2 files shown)


psweeting1 (approved):
LGTM

PR #1632, opened 2024-11-01
EL-1833 (part 2) add data_migration

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1833)

## What changed and why

added a new gem to handle data_migrations.
This is our first data_migration and although we don't expect to have many of these, they have a way of sneaking up on you as the db is modified. So it makes sense to add this now as you cant retrospectively move migrations.

The migration runs on my machine ;)
I tested it on early_eligibility_selection does not exist, is gross and is continue_check

It will not do anything on UAT as the branch is new and will have no data. it will work on main UAT when it is merged. As such I have added 2 approval steps, one for STG and one for PROD. This way we can merge to main and then test each namespace before and after the migration.

The data_migration uses :with_data to do the data migrations after the migrations are run. I have added this where I think it needs to be, but this is the bit I am least confident about. So the pipeline will again be the test bed for this.

## Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,7 +81,7 @@ references:
       name: Database Setup
       command: |
         bundle exec rake db:create db:schema:load
-        bundle exec rake db:migrate
+        bundle exec rake db:migrate:with_data
 
   setup_parallel_database: &setup_parallel_database
     run:
@@ -307,6 +307,7 @@ workflows:
             - linters
             - build_and_push
             - Deploy to UAT
+            - staging_deploy_approval
           post-steps:
             - jira/notify:
                 job_type: deployment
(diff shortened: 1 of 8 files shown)


HettieS (approved):
Approving this because I think it should be good. 

I'm also a little unsure on the use of :with_data as it isn't something I have seen before, and I think I've seen more instances of data migrations being run separately to db migrations ie with db:migrate and data:migrate. 

Reasons for that could be do to with large data processes taking a long time to execute/nuking the app BUT I'm not sure and I think for our current use case it is probably fine? Maybe one just to bear in mind. 

You don't have a rollback function in your data migration but given we are just lifting and shifting some data with minimal impact I think that's fine. We might want to think about defining a standard practice for writing these data migrations at some point if we think we'll have more (ie write a rollback method).

willc-work:
the rollback in this case would be to set the new field to null for all rows. we don't touch any of the other data in the table.

PR #1655, opened 2024-11-13
MTR Fix add bullet point styling

willc-work:

## Checklist

### DIFF
--- a/app/views/shared/_cost_of_living.html.slim
+++ b/app/views/shared/_cost_of_living.html.slim
@@ -3,6 +3,10 @@
 
 = govuk_details(summary_text:)
   p.govuk-body = t("cost_of_living_details.text")
-  p.govuk-body
-    - t("cost_of_living_details.list#{suffix}").each
-      div = _1
+  - if FeatureFlags.enabled?(:mtr_accelerated, without_session_data: true)
+    ul
+      - t("cost_of_living_details.list").each do |item|
+        li.govuk-text = item
+  - else
+    - t("cost_of_living_details.list_legacy").each do |item|
+      .govuk-text = item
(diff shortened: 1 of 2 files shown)


app/views/shared/_cost_of_living.html.slim
```diff
@@ -3,6 +3,10 @@
 
 = govuk_details(summary_text:)
   p.govuk-body = t("cost_of_living_details.text")
-  p.govuk-body
-    - t("cost_of_living_details.list#{suffix}").each
-      div = _1
+  - if FeatureFlags.enabled?(:mtr_accelerated, without_session_data: true)
+    ul
```
MazOneTwoOne, line 7: we can get rid of the `ul`, as it adds an additional space on the top which we don't want.

app/views/shared/_cost_of_living_payments.html.slim
```diff
@@ -2,5 +2,10 @@
...
 = govuk_details(summary_text:)
-  - t("cost_of_living_summary_and_details.list#{suffix}").each
-    div = _1
+  - if FeatureFlags.enabled?(:mtr_accelerated, without_session_data: true)
+    ul
+      - t("cost_of_living_summary_and_details.list").each do |item|
+        li.govuk-text = item
+  - else
+    - t("cost_of_living_summary_and_details.list_legacy").each do |item|
+      .govuk-text = item
```
MazOneTwoOne, line 11: Am I right in thinking we need to get rid of  

- `= render "shared/cost_of_living_payments"` 

and replace it with  

- `= render "shared/cost_of_living"`

That way we can get rid of this partial and the text in `en.yml`

MazOneTwoOne (approved):

app/views/shared/_cost_of_living_payments.html.slim
MazOneTwoOne: Will be picked up on another ticket.

PR #1660, opened 2024-11-14
EL-1462: MTR Accelerated Measures - LCG link updates

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1462)

## What changed and why

- update mappings based on feature flag
- add in new links
- update `500.html` with generic link

## Guidance to review

- not able to do pdf sizes will have to do in another ticket

## Checklist


app/services/external_link_service.rb
```diff
@@ -1,18 +1,122 @@
 class ExternalLinkService
   class << self
     def call(document:, sub_section: nil, page_number_only: false)
+      external_link_method = FeatureFlags.enabled?(:mtr_accelerated, without_session_data: true) ? external_links : external_links_legacy
       if page_number_only
-        external_links.dig(document, :sections).fetch(sub_section)
+        external_link_method.dig(document, :sections).fetch(sub_section)
```
willc-work, line 6: this is really minor as we will remove this `external_link_method` but I always find the use of `method` in the name of a method as a bit redundant and it doesnt help to clarify what the method is doing. so for something like this I would use _selection or _choice because it is making a decision between 2 things.

willc-work: no need to change it for this PR

willc-work (approved):

willc-work:
obviously we need to change the date and revert the circeci config

app/services/external_link_service.rb
MazOneTwoOne: I get the rationale and it makes sense and have updated.

PR #1669, opened 2024-11-21
EL-1878: `govuk-frontend` 5.7.1 bump

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1878)

## What changed and why

- update CCQ `govuk-frontend` to 5.7.1

## Guidance to review

- n/a

## Checklist


app/javascript/add-another.js
```diff
@@ -160,8 +160,10 @@ const setNumbering = (section, counter) => {
 }
 
 const setUpRadios = (section) => {
```
psweeting1, line 162:
```suggestion
const setUpRadios = (section) => {
  if (!section) return; // Early return if section is not provided or invalid

  const radioContainer = section.querySelector('[data-module="govuk-radios"]');
  if (radioContainer && !radioContainer.dataset.radiosInitialized) {
    try {
      const radios = new Radios(radioContainer); 
      radioContainer.dataset.radiosInitialized = "true"; 
    } catch (error) {
      console.error("Error initializing radios:", error);
    }
  }
};
```

You're code is great, just a little defensive programming to allow logging if initialising fails is all. 

What are you thoughts?

MazOneTwoOne: I've tried this and it too much of a guard clause and doesn't enter the try/catch. 

We have other tests that will fail on this, so will spot this in the future.

PR #1698, opened 2024-12-09
EL-1801: Investigate why specific fields shown as 0 when below £1.00

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1801)

## What changed and why

- evaluate a float instead of an integer

## Checklist


willc-work (approved):
the ticket mentions:
`However the values do not appear to be included in the eligibility calculation.`
Have we confirmed that these values are now included after this change?
Could we add a test to show that they are?

MazOneTwoOne:
> the ticket mentions: `However the values do not appear to be included in the eligibility calculation.` Have we confirmed that these values are now included after this change? Could we add a test to show that they are?

Good idea. Have added tests to the CFE payload service spec.

willc-work (approved):
double approved, thanks

PR #1707, opened 2024-12-19
EL-2005: Update CW forms on CCQ

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2005)

## What changed and why

- add new pdfs
- rename mappings
- update tests

## Guidance to review

- remove redundant feature flag, as this was no longer required and this change will not be behind a feature flag

## Checklist


app/lib/controlled_work_mappings/civ_means_7.yml
```diff
@@ -52,6 +52,7 @@
       source: smod_assets?
       yes_value: "Yes"
       no_value: "No"
+
```
willc-work, line 55: are these empty lines just to split the sections visually?

willc-work (approved):
all looks good. i was a little confused at first though. I would have expected that we would delete the .yml mappings files and then keep the ones wiith the `_new` suffix and simple renamed them after the original was deleted i.e remove the suffix. Moving the different methods seems like it may have been a lot of unnecessary effort. however it may be that you didnt manually move the methods, and it is just how git is representing the change. 

My concern is that each individual yml file was tested significantly during its initial release. With the solution that involves moving methods manually, we now have essentially a new yaml file that has not been as rigorously tested, although the component parts have. I may be being overly cautions on this, but it would be good to dicuss

app/lib/controlled_work_mappings/civ_means_7.yml
MazOneTwoOne: I think this is an artefact of changing the name i.e. removing the suffx `_new` and deleting the previous file which had the same name.

willc-work: thank you

MazOneTwoOne:
> all looks good. i was a little confused at first though. I would have expected that we would delete the .yml mappings files and then keep the ones wiith the `_new` suffix and simple renamed them after the original was deleted i.e remove the suffix. Moving the different methods seems like it may have been a lot of unnecessary effort. however it may be that you didnt manually move the methods, and it is just how git is representing the change.
> 
> My concern is that each individual yml file was tested significantly during its initial release. With the solution that involves moving methods manually, we now have essentially a new yaml file that has not been as rigorously tested, although the component parts have. I may be being overly cautions on this, but it would be good to dicuss

I did exactly that, I kept the files with `_new` suffix and removed the "legacy" mappings, then renamed the `_new` suffix files. GitHub might not represent that change easily. 

If it helps put your mind at ease, this file `spec/services/controlled_work_document_value_mapping_service_spec.rb` still captures the new mappings for the new/existing CW forms.

willc-work:
> > all looks good. i was a little confused at first though. I would have expected that we would delete the .yml mappings files and then keep the ones wiith the `_new` suffix and simple renamed them after the original was deleted i.e remove the suffix. Moving the different methods seems like it may have been a lot of unnecessary effort. however it may be that you didnt manually move the methods, and it is just how git is representing the change.
> > My concern is that each individual yml file was tested significantly during its initial release. With the solution that involves moving methods manually, we now have essentially a new yaml file that has not been as rigorously tested, although the component parts have. I may be being overly cautions on this, but it would be good to dicuss
> 
> I did exactly that, I kept the files with `_new` suffix and removed the "legacy" mappings, then renamed the `_new` suffix files. GitHub might not represent that change easily.
> 
> If it helps put your mind at ease, this file `spec/services/controlled_work_document_value_mapping_service_spec.rb` still captures the new mappings for the new/existing CW forms.

willc-work:
@MazOneTwoOne thanks for the update. I really dislike that git does that because it isn't a true representation of what happened.

But thanks for confirming your process, it means we definitely do not need the sort of extensive testing we have done for other cw form releases

PR #1775, opened 2025-03-05
EL-1947: Shared Ownership - exit page

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1947)

## What changed and why

- If applied, this change will add a `cannot-use-service` page for the shared property journey.

## Guidance to review

- n/a

## Checklist


app/forms/cannot_use_service_form.rb
```diff
@@ -0,0 +1,7 @@
+class CannotUseServiceForm
```
willc-work, line 1: is this necessary? The page is a static page with nothing that the user needs to complete. This feels like it is added just to appears the pattern in 
`app/controllers/flow/handler.rb` but without the need for a form existing.
As this is a static page could we create a static_pages controller and manage this page that way?
https://greg.molnar.io/blog/rails-tricks-issue-15/

app/lib/steps/logic.rb
```diff
@@ -88,6 +88,10 @@ def owns_property_shared_ownership?(session_data)
         !skip_capital_questions?(session_data) && session_data["property_owned"] == "shared_ownership"
       end
 
+      def landlord_is_not_the_only_joint_owner?(session_data)
```
willc-work, line 91: this scans weirdly for me, can we rename this method so we dont have the NOT, so for example, `property_has_more_than_two_owners?` or `multiple_party_ownership?`

kmahern:
@MazOneTwoOne is it please possible to rebase this on main now that EL-1616 has been merged, to make it easier to review?

app/forms/cannot_use_service_form.rb
MazOneTwoOne: @willc-work I think this is necessary and would prefer to use the `app/controllers/flow/handler.rb` logic for the following reasons:
- we use a similar pattern for the `app/forms/how_to_aggregate_form.rb` form and screen
- currently let's us tap into question flow logic but if we used a static page (as per link) might be convoluted to get it to work with `app/controllers/flow/handler.rb`, `app/lib/steps/logic.rb` & `app/controllers/change_answers_controller.rb` logic
- if we ever need to add any additional buttons etc we can add to the form model.

app/lib/steps/property_section.rb
```diff
@@ -1,6 +1,6 @@
 module Steps
   class PropertySection
-    PROPERTY_STEPS = %i[property property_landlord property_entry].freeze
+    PROPERTY_STEPS = %i[property property_landlord cannot_use_service property_entry].freeze
```
willc-work, line 3: we add `cannot_use_service` to `PROPERTY_STEPS` doesn't that mean that we will always see that page when `PROPERTY_STEPS` is called? I don't see in the file where we are determining when to show that page?

spec/forms/property_landlord_form_spec.rb
```diff
@@ -82,4 +126,24 @@
       expect(session_contents["property_landlord"]).to be true
     end
   end
+
```
kmahern, line 129: See comment above re questioning the need for a cannot_use_service_form, but if we are adding a new cannot_user_service_form I think this test should be in a separate cannot_use_service_form_spec.

MazOneTwoOne: @kmahern agreed. Can we pull this out into a separate ticket to refactor after we get the flow in?

app/forms/cannot_use_service_form.rb
willc-work: I am happy that this can be merged and we write a ticket to refactor this so we can merge this change and continue with the development of the feature, but lets keep an eye on how much work that refactor becomes. If it is a significant rework by the time we get to it then we may want to bear that in mind for future similar decisions.

app/lib/steps/property_section.rb
MazOneTwoOne: @willc-work this also really confused me. 

I think this is my understanding: `PropertySection` is only called by `all_sections` in `app/lib/steps/helper.rb`, and it needs it to know of all the sections that we have. The logic of what page to show in the flow, is dictated by `app/lib/steps/outgoings_section.rb` and the logic of the flow is there.

app/forms/cannot_use_service_form.rb
MazOneTwoOne: @kmahern happy to pair and attempt that on another refactor ticket. 🙂

willc-work: approving based on the creation of a ticket to look at refactoring this

willc-work (approved):
Approved, discussed with @MazOneTwoOne that we need a refactor ticket for this to use a static controller

PR #1776, opened 2025-03-05
EL-1579 shared home costs

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1579)

## What changed and why

added new template for shared_ownership housing costs
split housing cost form to use a base form
add shared ownership housing costs form
add tests to cover new form

TODO:
currently the shared_ownership_housing_costs page is not shown if an additional property is marked as shared_ownership
I had a look at the figma and it was not clear to me where this new page would be shown, although it would seem to make sense to show it straight after the question is asked. This need clarification and I think would warrant being in its own ticket as this one feels like it has grown legs already.

## Checklist

### DIFF
--- a/app/forms/shared_ownership_housing_costs_form.rb
+++ b/app/forms/shared_ownership_housing_costs_form.rb
@@ -1,45 +1,34 @@
 class SharedOwnershipHousingCostsForm < BaseHousingCostsForm
-  ATTRIBUTES = (BaseHousingCostsForm::ATTRIBUTES + %i[shared_ownership_mortgage mortgage_frequency rent rent_frequency]).freeze
+  ATTRIBUTES = (BaseHousingCostsForm::ATTRIBUTES + %i[shared_ownership_mortgage rent combined_frequency]).freeze
 
   attribute :shared_ownership_mortgage, :gbp
   validates :shared_ownership_mortgage,
             numericality: { greater_than_or_equal_to: 0, allow_nil: false },
             is_a_number: true
 
-  attribute :mortgage_frequency, :string
-  validates :mortgage_frequency,
-            inclusion: { in: OutgoingsForm::VALID_FREQUENCIES, allow_nil: false },
-            if: -> { shared_ownership_mortgage.to_i.positive? }
-
   attribute :rent, :gbp
   validates :rent,
             numericality: { greater_than_or_equal_to: 0, allow_nil: false },
             is_a_number: true
 
-  attribute :rent_frequency, :string
-  validates :rent_frequency,
+  attribute :combined_frequency, :string
+  validates :combined_frequency,
             inclusion: { in: OutgoingsForm::VALID_FREQUENCIES, allow_nil: false },
-            if: -> { rent.to_i.positive? }
+            if: -> { combined_rent_and_mortgage.to_i.positive? }
 
   def housing_payment_frequencies
     OutgoingsForm::VALID_FREQUENCIES.map { [_1, I18n.t("question_flow.outgoings.frequencies.#{_1}")] }
   end
 
+  def combined_rent_and_mortgage
+    (shared_ownership_mortgage.to_f + rent.to_f).round(2)
+  end
+
 private
 
   def total_annual_housing_costs
-    total = 0
+    return 0 if combined_frequency.blank?
 
-    # Only add mortgage costs if we have a frequency (amount can be zero)
-    if mortgage_frequency.present?
-      total += shared_ownership_mortgage * annual_multiplier(mortgage_frequency)
-    end
-
-    # Only add rent costs if we have a frequency (amount can be zero)
-    if rent_frequency.present?
-      total += rent * annual_multiplier(rent_frequency)
-    end
-
-    total
+    combined_rent_and_mortgage * annual_multiplier(combined_frequency)
   end
 end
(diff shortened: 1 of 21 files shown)


spec/models/shared_ownership_housing_costs_form_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+      mortgage: 0,
+      mortgage_frequency: nil,
+      rent: 0,
+      rent_frequency: nil,
+      housing_benefit_relevant: false,
+    }
+  end
+
+  context "when form is submitted and" do
+    context "when mortgage is positive" do
```
kmahern, line 17: @willc-work Can we include some tests to check that the session_data is actually updated with the correct values? When I run this locally it looks like possibly the mortgage isn't being saved to the session_data.

kmahern:
@willc-work I think the mortgage value entered in app/forms/shared_ownership_housing_costs_form.rb is being overwritten by mortgage in property_entry_form - I think it needs a different name.

app/forms/shared_ownership_housing_costs_form.rb
```diff
@@ -0,0 +1,45 @@
+class SharedOwnershipHousingCostsForm < BaseHousingCostsForm
+  ATTRIBUTES = (BaseHousingCostsForm::ATTRIBUTES + %i[mortgage mortgage_frequency rent rent_frequency]).freeze
```
kmahern, line 2: @willc-work I think you need to add the attribute shared_ownership_mortgage and remove the attribute mortgage here.

spec/models/shared_ownership_housing_costs_form_spec.rb
```diff
@@ -0,0 +1,120 @@
+require "rails_helper"
+
```
kmahern, line 2: Is it worth including some extra tests that the session_data mortgage/rent is actually updated with the correct values?

e.g.  expect(session_contents["shared_ownership_mortgage"]).to eq 500
        expect(session_contents["mortgage_frequency"]).to eq "monthly"

willc-work: added test that covers all of the new fields stored in the session

app/services/cfe/regular_transactions_payload_service.rb
```diff
@@ -5,13 +5,16 @@ def call
         relevant_form?(:other_income) ||
         relevant_form?(:mortgage_or_loan_payment) ||
         relevant_form?(:housing_costs) ||
+        relevant_form?(:shared_ownership_housing_costs) ||
```
kmahern, line 8: I wonder whether we should be making the changes to the CFE payload in a separate ticket/pull request?

app/views/shared/_money_input_no_label.html.slim
```diff
@@ -0,0 +1,13 @@
...
+  hint_text ||= ""
+  checktype = level == "controlled" ? "controlled_" : nil
+= form.govuk_text_field field,
+        width:,
+        hint: { text: hint_text },
+        label: nil,
+        prefix_text: FormsHelper::POUND,
+        value: decimal_as_money_string(form, field),
+        inputmode: "decimal",
+        "aria-label": set_aria ? t("question_flow.aria_label.#{checktype}#{field}").to_s : nil
```
kmahern, line 13: This looks like it would fail validation/couldn't be displayed. Is the comment intended?

kmahern:
Looking good.

I think the _money_input_no_label partial looks like it has a comment that needs review.

A couple of further questions/queries from me.

Combining the mortgate and rent rent frequencies doesn't seem to match the designs on the ticket. Also, it will have a knock on impact for the design of other tickets like the cya page. I think this should be discussed first.

It does also feel like this ticket has "grown legs" though and I wonder whether it would be worth separating the CFE changes into a separate pr/ticket?

app/forms/shared_ownership_housing_costs_form.rb
```diff
@@ -0,0 +1,34 @@
...
+            numericality: { greater_than_or_equal_to: 0, allow_nil: false },
+            is_a_number: true
+
+  attribute :rent, :gbp
+  validates :rent,
+            numericality: { greater_than_or_equal_to: 0, allow_nil: false },
+            is_a_number: true
+
+  attribute :combined_frequency, :string
+  validates :combined_frequency,
```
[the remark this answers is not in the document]
willc-work: we need to change the design, discussed in slack here https://mojdt.slack.com/archives/C02BKRV26KS/p1741806638512359

app/services/cfe/regular_transactions_payload_service.rb
willc-work: it could but it makes sense to go int his ticket really as the new form is created in this ticket

app/views/shared/_money_input_no_label.html.slim
willc-work: I am not clear on what this comment means. this partial is required because we dont want a label for this input field. Although this may change as we have had to revise the content of this page.

kmahern: I meant the #{checktype}#{field}").to_s : nil comment? It looks like the bracket for the translation is not closed?

kmahern: Apologies @willc-work the github formatting made it look like there was a comment but it is actually just string interpolation!

kmahern (approved):
Discussed questions in standup. Approving 👍

PR #1783, opened 2025-03-12
EL-1582: Shared Ownership, Results page (Household monthly outgoings)

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1582)

## What changed and why

- If applied, this change hides content based on feature flag & adds test coverage

## Guidance to review

- `allowed_housing_costs_exceed_cap?` is based on a constant, but is clearly labelled for future devs

## Checklist


spec/views/results_page/housing_outgoings_content_spec.rb
```diff
@@ -0,0 +1,355 @@
+require "rails_helper"
```
willc-work, line 1: So this spec can be simplified. I think you have added a complete `session_data` and `api_response` from an actual check on the service, which is a good way to ensure that you are testing the correct inputs and outputs. However now that you have this data you can work out what is actually relevant for this test and remove the excess. This will also allow you to identify what specific information is being included in this request that the current FactoryBot doesnt cover. And then we can turn that into a trait. Not that that all has to be done on this PR. I will puysh up a commit shortly that has some of the `session_data` and `api_response` removed and the tests still pass.  Can you apply this to the other test set up and also check to see if there is anything further that could be removed. One place start in `session_data` is removing anything set to `false` and in `api_response` sections with no info e.g. `cash_transactions` where all values are zero are also good options for removal.

MazOneTwoOne: @willc-work Yeah that sounds good. I had removed a bunch of stuff already, but happy to take another look and remove some more. :)

willc-work (approved):
Happy to approve but can we add a note that this test needs to be refactored and write the ticket to refactor it. From what is required to run this it looks like we need to expand the `disposable_income_summary` factory in `spec/factories/api_results.rb`.

At present it is created as an empty hash with an associated proceeding type. We need to add the relevant fields to the factory and then this test can just call factorybot with the `disposable_income_summary` trait

willc-work (approved):
happy to approve

PR #1794, opened 2025-03-18
EL-2130 add threshold flag and dependant rate

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2160)

## What changed and why

add time based feature flag and update the `dependant_monthly_upper_limit`

## Guidance to review

The ticket references partner allowance but that comes from the CFE response so is done on the CFE ticket, we will need to UAT that change on that ticket, comments to this affect added to both JIRA tickets

I had an issue with a misbehaving test, details are here: https://mojdt.slack.com/archives/C040DEGBZQQ/p1742399721684899
Hence the comment on the spec file that indicates a test will break on 7th April

## Checklist

### DIFF
--- a/app/lib/feature_flags.rb
+++ b/app/lib/feature_flags.rb
@@ -1,7 +1,7 @@
 class FeatureFlags
   ENABLED_AFTER_DATE = {
     example_2125_flag: { from: "2125-01-01", public: false },
-    thresholds_2025: { from: "2025-03-21", public: true }, # date needs to be set to 2025-04-07 before merging
+    thresholds_2025: { from: "2025-04-07", public: true },
   }.freeze
 
   # the values of some feature flags will come from the session and not the env variables.
--- a/app/lib/feature_flags.rb
+++ b/app/lib/feature_flags.rb
@@ -1,8 +1,7 @@
 class FeatureFlags
   ENABLED_AFTER_DATE = {
     example_2125_flag: { from: "2125-01-01", public: false },
-    thresholds_2025: { from: "2025-03-18", public: false },
-    # thresholds_2025: { from: "2025-04-07", public: false },
+    thresholds_2025: { from: "2025-03-21", public: true }, # date needs to be set to 2025-04-07 before merging
   }.freeze
 
   # the values of some feature flags will come from the session and not the env variables.
--- a/app/lib/feature_flags.rb
+++ b/app/lib/feature_flags.rb
@@ -1,6 +1,8 @@
 class FeatureFlags
   ENABLED_AFTER_DATE = {
     example_2125_flag: { from: "2125-01-01", public: false },
+    thresholds_2025: { from: "2025-03-18", public: false },
+    # thresholds_2025: { from: "2025-04-07", public: false },
   }.freeze
 
   # the values of some feature flags will come from the session and not the env variables.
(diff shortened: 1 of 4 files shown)


app/lib/feature_flags.rb
```diff
@@ -1,6 +1,8 @@
 class FeatureFlags
   ENABLED_AFTER_DATE = {
     example_2125_flag: { from: "2125-01-01", public: false },
+    thresholds_2025: { from: "2025-03-18", public: false },
```
willc-work, line 4: this line is added for UAT it will be removed and the line below will be uncommented priuor to merge

spec/models/dependant_income_model_spec.rb
```diff
@@ -1,7 +1,11 @@
...
 
 RSpec.describe DependantIncomeModel do
-  context "when after 8th April 2024" do
+  context "when before 7th April 2025" do
+    let(:fixed_arbitrary_date) { Date.new(2025, 3, 17) }
+
+    before { travel_to fixed_arbitrary_date }
+
     it "has new dependant limits" do
       expect(described_class.dependant_income_upper_limits).to eq({ "every_four_weeks" => 332.96,
```
kmahern, line 10: nit: should this be "it has old dependant limits" (for clarity)

app/models/dependant_income_model.rb
```diff
@@ -34,7 +34,7 @@ def income_too_high
 
   class << self
     def dependant_monthly_upper_limit
-      361.70
+      FeatureFlags.enabled?(:thresholds_2025, without_session_data: true) ? 367.87 : 361.70
```
MazOneTwoOne, line 37: @willc-work I thought the new limit was meant to be £367.87 and the olf limit was meant to be £361.70.

So should this not be
 FeatureFlags.enabled?(:thresholds_2025, without_session_data: true) ? 367.87 : 361.70

MazOneTwoOne (approved):

willc-work:
@kmahern can you just recheck this PR as I have now reverted the date to 04072025 and would like a second approval before i merge. I think just a comment to that effect will be fine and I will do the same on your CFE threshold PR

kmahern (approved):
👍 reviewed change to correct date for threshold change and lgtm.

spec/forms/dependant_details_form_spec.rb
```diff
@@ -31,11 +31,13 @@
...
 
+  # This test will fail after the 7th April when the new thresholds come into force.
+  # To fix this test, change the value `361.70` on line 40 below to `367.87`
   it "shows me the dependant allowance text" do
     expect(page).to have_content(
       "Do not include:\n"\
-      "anyone who owns any property, vehicles or other assets valued at over £8,000 in total"\
-      "anyone with £361.70 or more income every month (any income below this can be entered on the following pages and will be deducted from the dependant allowance)",
+        "anyone who owns any property, vehicles or other assets valued at over £8,000 in total"\
+        "anyone with £361.70 or more income every month (any income below this can be entered on the following pages and will be deducted from the dependant allowance)",
```
aaj-moj: Isn't this text supposed to say £367.87 now?

Edit: I see... this test is correct because the flag `thresholds_2025` isn't set, right? So we'd need a similar test that sets the flag to true and checks that this text says £367.87.

PR #1798, opened 2025-03-20
EL-2170 Lock govuk fronted version

kmahern:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2170)

## What changed and why

we are locking the version of this dependency as higher version has a breaking change. we intend to revisit this decision in 1 month to see if the issue is resolved or if there is an easier upgrade path, or failing that deal with the pain of upgrading at that stage.

## Checklist


willc-work (approved):
I think we should merge this change to get dependabot coverage back for the service. We then need a separate ticket written up to do this upgrade, which should link to this thread https://mojdt.slack.com/archives/C040DEGBZQQ/p1742470409558409?thread_ts=1742466450.575049&cid=C040DEGBZQQ 
and also to this slack channel
#moj-design-system-support 
there are some suggested ways to do the upgrade in there and they also function as a support page so we can ask for help if we have any problems. I think this way we can `kick the can down the road` whilst others are figuring it out and we can then ~~steal plagiarise borrow~~ reuse their code

MazOneTwoOne (approved):

PR #1799, opened 2025-03-20
EL-2167: Shared Ownership - send housing benefit to CFE

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2167)

## What changed and why

- If applied, this change will add housing benefits when user uses either `HousingCostsForm` or `SharedOwnershipHousingCostsForm`

## Guidance to review

- n/a

## Checklist


kmahern (approved):
lgtm - possibly we could look at adding some sort of spec to check that housing benefit is sent to cfe for shared ownership properties/? There is a spec "when client or their partner own their home shared with a housing association" in regular_transactions_payload_service_spec

willc-work (commented):
I would like to see a test that ensures both of the housing costs forms can trigger this and a non BaseHousingCostsForm such as MortgageOrLoanPaymentForm does not

MazOneTwoOne:
@willc-work @ @kmahern Have added tests to show that CFE can receive housing benefits using either `HousingCostsForm` or `SharedOwnershipHousingCostsForm`. 

I don't think adding a test to show a non `BaseHousingCostsForm` can send housing benefits, as the flow of pages prevents user seeing them and even if they got to these form pages, they have no way to enter the housing benefits (as the form doesn't have those fields).  Have opted not to include a test for this scenario.

PR #1809, opened 2025-03-27
El 2166 additional properties exit page

kmahern:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2166)

## What changed and why

Added in new Cannot Use Service screens, if the user selects shared_ownership for additional_property or partner_additional_property.

These have been added by creating a new controller/route cannot_use_service.

I believe this is a more RoR way of implementing this functionality, as these pages fall outside of the question flow service (as the user is prevented from continuing with their journey) and we should not have to create a form class in order to display the page when we are not capturing any new user input.

What do people think?

Ideally we could use something like this pattern for EL_2159.

## Checklist

### DIFF
--- a/spec/flows/cannot_use_service_flow_spec.rb
+++ b/spec/flows/cannot_use_service_flow_spec.rb
@@ -1,0 +1,47 @@
+require "rails_helper"
+
+RSpec.describe "Cannot use service flow", :ee_banner, :shared_ownership, :stub_cfe_calls_with_webmock, type: :feature do
+  context "when the client has a shared ownersship property" do
+    it "shows me a screen to indicate that I cannot use the service" do
+      start_assessment
+      fill_in_forms_until(:applicant)
+      fill_in_applicant_screen(partner: "No", passporting: "Yes")
+      fill_in_forms_until(:additional_property)
+      fill_in_additional_property_screen(choice: "Yes, through a shared ownership scheme")
+      confirm_screen(:cannot_use_service)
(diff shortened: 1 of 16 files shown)


spec/flows/cannot_use_service_flow_spec.rb
```diff
@@ -0,0 +1,47 @@
+require "rails_helper"
+
+RSpec.describe "Cannot use service flow", :ee_banner, :shared_ownership, :stub_cfe_calls_with_webmock, type: :feature do
+  context "when the client has a shared ownersship property" do
```
willc-work, line 4: minor nitpick - spelling of ownership is wrong

spec/flows/cannot_use_service_flow_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      fill_in_applicant_screen(partner: "No", passporting: "Yes")
+      fill_in_forms_until(:additional_property)
+      fill_in_additional_property_screen(choice: "Yes, through a shared ownership scheme")
+      confirm_screen(:cannot_use_service)
+      expect(page).to have_content "You cannot use this service"
+      expect(page).to have_content "You cannot use this service if your client owns a shared ownership property that they do not live in."
+    end
+  end
+
+  context "when the client's partner has a shared ownersship property" do
```
willc-work, line 17: same spelling error

willc-work (approved):
LGTM

willc-work (approved):

PR #1810, opened 2025-03-27
Bugfix/EL-2205 validate housing payments

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2205)

## What changed and why

validation order is not guaranteed so validations on HousingCostsForm may run before or after the validations on BaseHousngCostsForm. Which means `not_a_number` or `numeric` can be called after we try to undertake a calculation to see if housing benefit exceeds housing payments, the problem being that the calculation is run before these validations and this tries to use an invalid string for the calculation.

This fixes the issue by defending against these issues in the the calculation allowing the standard validation to fire instead

However I think a refactor is in order here. I initially wanted to simplify things by using a base housing costs form. And it is this split between the base form and the child forms that is causing some much pain with validation. Making it unnecessarily complex to guard against validations that are expected to run but are in a different order.

I will add a ticket to cover the refactor of this
Ticket to refactor is here:
https://dsdmoj.atlassian.net/browse/EL-2207

## Guidance to review

For some reason the test I added requires this:

.to include({ housing_payments: ["Housing payments must be a number. Enter 0 if this does not apply.", "Housing payments must be a number. Enter 0 if this does not apply."] })

I have to repeat the error message, although it only appears once on the page so I am not sure how to resolve this but for now we can UAT and perhaps resolve this with the refactor

## Checklist

### DIFF
--- a/app/forms/base_housing_costs_form.rb
+++ b/app/forms/base_housing_costs_form.rb
@@ -35,27 +35,28 @@ class BaseHousingCostsForm
 private
 
   def housing_benefit_does_not_exceed_costs
-    # Skip validation if housing benefit isn't relevant or value/frequency is missing
     return unless housing_benefit_relevant &&
-      housing_benefit_value.to_i.positive? &&
+      housing_benefit_value.respond_to?(:positive?) &&
+      housing_benefit_value.positive? &&
       housing_benefit_frequency.present?
 
-    # If housing payments are zero, any housing benefit is excessive
-    if total_annual_housing_costs.zero?
-      errors.add(:housing_benefit_value, :exceeds_costs)
-      return
-    end
+    total_costs = total_annual_housing_costs
 
-    # If we can't properly calculate housing costs due to missing frequencies,
-    # skip this validation - other validations will catch the missing inputs
-    begin
-      annual_housing_payment_value = total_annual_housing_costs
-      annual_housing_benefit_value = housing_benefit_value * annual_multiplier(housing_benefit_frequency)
+    # Return early if total_costs is not a numeric type or is zero
+    return unless total_costs.respond_to?(:zero?) && !total_costs.zero?
 
-      return if annual_housing_payment_value >= annual_housing_benefit_value
+    # Ensure housing_benefit_value is numeric and frequency is valid
+    # this line stubbornly refuses to be tested so adding a nocov as a refactor is required here
+    # :nocov:
+    return unless housing_benefit_value.respond_to?(:*) && annual_multiplier(housing_benefit_frequency)
+    # :nocov:
 
-      errors.add(:housing_benefit_value, :exceeds_costs)
-    end
+    annual_housing_payment_value = total_costs
+    annual_housing_benefit_value = housing_benefit_value * annual_multiplier(housing_benefit_frequency)
+
+    return if annual_housing_payment_value >= annual_housing_benefit_value
+
+    errors.add(:housing_benefit_value, :exceeds_costs)
   end
 
   def annual_multiplier(frequency)
--- a/app/forms/base_housing_costs_form.rb
+++ b/app/forms/base_housing_costs_form.rb
@@ -43,13 +43,11 @@ private
     total_costs = total_annual_housing_costs
 
     # Return early if total_costs is not a numeric type or is zero
-    return unless total_costs.respond_to?(:zero?) && !total_costs.zero?
+    return unless total_costs.is_a?(Numeric) && total_costs.nonzero?
+    # return unless total_costs.respond_to?(:zero?) && !total_costs.zero?
 
     # Ensure housing_benefit_value is numeric and frequency is valid
-    # this line stubbornly refuses to be tested so adding a nocov as a refactor is required here
-    # :nocov:
-    return unless housing_benefit_value.respond_to?(:*) && annual_multiplier(housing_benefit_frequency)
-    # :nocov:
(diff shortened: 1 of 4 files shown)


app/forms/base_housing_costs_form.rb
```diff
@@ -35,27 +35,28 @@ def frequencies
...
-      annual_housing_benefit_value = housing_benefit_value * annual_multiplier(housing_benefit_frequency)
+    annual_housing_payment_value = total_costs
+    annual_housing_benefit_value = housing_benefit_value * annual_multiplier(housing_benefit_frequency)
 
-      return if annual_housing_payment_value >= annual_housing_benefit_value
+    return if annual_housing_payment_value >= annual_housing_benefit_value
 
-      errors.add(:housing_benefit_value, :exceeds_costs)
-    end
+    errors.add(:housing_benefit_value, :exceeds_costs)
```
MazOneTwoOne, line 59: @willc-work I think you can do:

```
def total_annual_housing_costs
  raise NotImplementedError, "Subclasses must implement the total_annual_housing_costs method"
end
```

Which means you can get rid of `# :nocov:` on this bit

app/forms/base_housing_costs_form.rb
```diff
@@ -35,27 +35,28 @@ def frequencies
 private
 
   def housing_benefit_does_not_exceed_costs
-    # Skip validation if housing benefit isn't relevant or value/frequency is missing
     return unless housing_benefit_relevant &&
-      housing_benefit_value.to_i.positive? &&
+      housing_benefit_value.respond_to?(:positive?) &&
```
MazOneTwoOne, line 39: I'll be honest, I've never heard of `respond_to?`, but that is a useful ruby method. 

Not a blocker, but do you think we should change the checking of a number to something like this: 

`housing_benefit_value.is_a?(Numeric)`

I've just seen the use of `is_a?` more in CCQ.

app/forms/base_housing_costs_form.rb
```diff
@@ -35,27 +35,28 @@ def frequencies
...
 
-    # If housing payments are zero, any housing benefit is excessive
-    if total_annual_housing_costs.zero?
-      errors.add(:housing_benefit_value, :exceeds_costs)
-      return
-    end
+    total_costs = total_annual_housing_costs
+
+    # Return early if total_costs is not a numeric type or is zero
+    return unless total_costs.respond_to?(:zero?) && !total_costs.zero?
```
MazOneTwoOne, line 46: Could this be:

`return unless total_costs.is_a?(Numeric) && total_costs.nonzero?`

app/forms/base_housing_costs_form.rb
```diff
@@ -35,27 +35,28 @@ def frequencies
...
+
+    # Ensure housing_benefit_value is numeric and frequency is valid
+    # this line stubbornly refuses to be tested so adding a nocov as a refactor is required here
+    # :nocov:
+    return unless housing_benefit_value.respond_to?(:*) && annual_multiplier(housing_benefit_frequency)
```
MazOneTwoOne, line 51: Again, could this be:

`return unless housing_benefit_value.is_a?(Numeric) && annual_multiplier(housing_benefit_frequency)`


Now that you are checking it is numeric rather than if it can be multiplied, would it be easier to test?

MazOneTwoOne (changes requested):
I just have general comments, but please do tell me if you've already tried these.

app/forms/base_housing_costs_form.rb
willc-work: I like this solution a lot. thank you

app/forms/base_housing_costs_form.rb
willc-work: I did try `is_a?(Numeric)` and then tried respond_to to see if that helped me able to test but neither was working. I have changed it to `is_a?(Numeric)` because it is more common in our code base, but the refactor to remove the common parent form should remove the need for this complexity because we can then rely on the validates on the form to undertake these checks before we do any calcs

app/forms/base_housing_costs_form.rb
MazOneTwoOne: @willc-work did you also want to change it to `housing_benefit_value.is_a?(Numeric)` on this bit?

app/forms/base_housing_costs_form.rb
```diff
@@ -35,27 +35,28 @@ def frequencies
...
-    # If housing payments are zero, any housing benefit is excessive
-    if total_annual_housing_costs.zero?
-      errors.add(:housing_benefit_value, :exceeds_costs)
-      return
-    end
+    total_costs = total_annual_housing_costs
+
+    # Return early if total_costs is not a numeric type or is zero
+    return unless total_costs.is_a?(Numeric) && total_costs.nonzero?
+    # return unless total_costs.respond_to?(:zero?) && !total_costs.zero?
```
MazOneTwoOne, line 47: If we are going with:

`return unless total_costs.is_a?(Numeric) && total_costs.nonzero?`

we should remove the commented out code:
`# return unless total_costs.respond_to?(:zero?) && !total_costs.zero?`

app/forms/base_housing_costs_form.rb
willc-work: yes thanks I have just updated it

MazOneTwoOne (approved):

PR #1815, opened 2025-03-31
EL-2194 change answers flow for shared ownership

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2194)

## What changed and why

added in redirect on change answers when multiple parties own the property and users need to see the exit page
added in property_entry step to outgoings to account for shared-ownership in change answers flow

Issues: we have an infinite loop on property_entry, fixing this causes the change answers test of shared ownership to fail.

What is the issue?
We exit Change Answers when we reach `last_step_in_group`
`:shared_ownership_housing_costs` is the last step in the `OutgoingsSection` and `:property_entry` is the first step in PropertySection, so this works fine for the standard flow but in change answers we return to change answers without getting all required information.

Adding in a fix to send users to property_entry has so many edge cases to consider that you just end up with a complex method that may have unintended consequences elsewhere.

We have a lot of shared steps in these 2 sections:
```
  :property,
  :property_landlord,
  :cannot_use_service,
  :shared_ownership_housing_costs,
  :mortgage_or_loan_payment,
  :housing_costs,
  :property_entry
```
We may need to consider how we would refactor this to be able to complete this feature. Ideally there is a solution that we can identify that means we can refactor the service after delivery of this feature so I think that should be the first consideration

## Guidance to review
I have left in some debug steps commented out, these will be removed before merge but just in case something arises in review or UAT i want to keep these available on this branch
## Checklist

### DIFF
--- a/app/controllers/change_answers_controller.rb
+++ b/app/controllers/change_answers_controller.rb
@@ -19,7 +19,6 @@ class ChangeAnswersController < QuestionFlowController
         elsif next_step && step != :aggregated_means
           redirect_to helpers.check_step_path_from_step(next_step, assessment_code)
         elsif Steps::Helper.last_step_in_group?(session_data, step)
-          #  check answers thinks it is at the last step but it isnt in normal flow???!
           save_and_redirect_to_check_answers
         else
           # this is the mini-loop - it isn't the last step in the section, we have
--- a/app/controllers/change_answers_controller.rb
+++ b/app/controllers/change_answers_controller.rb
@@ -11,12 +11,15 @@ class ChangeAnswersController < QuestionFlowController
       session_data.merge!(@form.attributes_for_export_to_session)
       if FeatureFlags.enabled?(:ee_banner, session_data)
         next_step = step_with_inconsistent_data
-        if Steps::Helper.cannot_use_service?(session_data, step)
+        if Steps::Logic.landlord_is_not_the_only_joint_owner?(session_data)
+          redirect_to helpers.step_path_from_step(:cannot_use_service, assessment_code)
+        elsif Steps::Helper.cannot_use_service?(session_data, step)
           redirect_to cannot_use_service_path assessment_code:, step:
         # we need to check for aggregated_means so we know when to show the ":how_to_aggregate" screen when in a change loop
         elsif next_step && step != :aggregated_means
           redirect_to helpers.check_step_path_from_step(next_step, assessment_code)
         elsif Steps::Helper.last_step_in_group?(session_data, step)
+          #  check answers thinks it is at the last step but it isnt in normal flow???!
           save_and_redirect_to_check_answers
         else
           # this is the mini-loop - it isn't the last step in the section, we have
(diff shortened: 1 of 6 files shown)


app/controllers/change_answers_controller.rb
```diff
@@ -11,12 +11,15 @@ def update
       session_data.merge!(@form.attributes_for_export_to_session)
       if FeatureFlags.enabled?(:ee_banner, session_data)
         next_step = step_with_inconsistent_data
-        if Steps::Helper.cannot_use_service?(session_data, step)
+        if Steps::Logic.landlord_is_not_the_only_joint_owner?(session_data)
```
willc-work, line 14: could this be an OR statement instead of 2 different if/elsif?

app/lib/steps/helper.rb
```diff
@@ -51,6 +51,11 @@ def completed_steps_for(session_data, step)
       def last_step_in_group?(session_data, step)
         step_group = step_groups_for(session_data).detect { |group| group.steps.include?(step) }
         step == step_group.steps.last
+        # step_group = step_groups_for(session_data).detect { |group| group.steps.include?(step) }
```
willc-work, line 54: to be deleted before merge

app/lib/steps/helper.rb
```diff
@@ -100,6 +105,17 @@ def steps_list_for(session_data)
       def step_groups_for(session_data)
         all_sections.map { |section| section.grouped_steps_for(session_data) }.reduce(:+)
       end
+      # debug steps groups
```
willc-work, line 108: to be deleted before merge

app/controllers/change_answers_controller.rb
```diff
@@ -11,12 +11,15 @@ def update
...
+          redirect_to cannot_use_service_path assessment_code:, step:
```
kmahern, line 15: @willc-work the logic here seems to be getting quite complex.

I am also not sure it is correct/will work the way you intend?

The cannot_use_service_path assessment_code:, step: route is currently only for when the client is exited because they or their partner have an additional shared ownership property.

When the applicant is exited because they have a landlord who is not the sole owner this uses the cannot_use_service_path assessment_code: path (ie it doesn't use the cannot_use_service controller - the work to do that is in EL-2159)

So I think this pr will currently take the user to the wrong exit page?

kmahern:
@willc-work I am a bit concerned that the logic seems to be getting quite complicated in the change_answers_controller. 

Also I think it would currently take the user to the wrong exit page?

I am wondering whether it would have been easier to tackle this ticket once EL-2159 was complete.

happy to discuss.

app/controllers/change_answers_controller.rb
willc-work: I think we can reuse the same page to exit users, although that may impact metrics, is that your concern? The ticket just says :
`Selecting ‘No’ to 'Is the landlord the only other joint owner?' does not direct you to an exit page.`
There is no specification of what the exit page should be so I have assumed the existing exit page is to be used. I will check this with Tob

kmahern: @willc-work No there are multiple exit pages. Do you want to discuss.

PR #1818, opened 2025-04-02
EL-2222 Shared ownership Check Answers flow

willc-work:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2222)

## What changed and why

Adds a test for the existing flow which is working
this is part A of this ticket, this is the test that we must not break while trying to fix check answers loops

## Checklist

### DIFF
--- a/app/controllers/change_answers_controller.rb
+++ b/app/controllers/change_answers_controller.rb
@@ -16,7 +16,7 @@ class ChangeAnswersController < QuestionFlowController
         elsif Steps::Helper.cannot_use_service?(session_data, step)
           redirect_to cannot_use_service_path assessment_code:, step:
         # we need to check for shared_ownership_housing_costs to determine if we need to show property_entry screen when in the change loop
-        elsif step == :shared_ownership_housing_costs && !Steps::Logic.landlord_is_not_the_only_joint_owner?(session_data)
+        elsif Steps::Helper.display_property_entry_in_change_answers?(session_data, step)
           redirect_to helpers.check_step_path_from_step(:property_entry, assessment_code)
         # we need to check for aggregated_means so we know when to show the ":how_to_aggregate" screen when in a change loop
         elsif next_step && step != :aggregated_means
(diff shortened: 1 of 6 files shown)


MazOneTwoOne (dismissed):

app/controllers/change_answers_controller.rb
```diff
@@ -14,6 +14,8 @@ def update
         if Steps::Helper.cannot_use_service?(session_data, step)
           redirect_to cannot_use_service_path assessment_code:, step:
         # we need to check for aggregated_means so we know when to show the ":how_to_aggregate" screen when in a change loop
+        elsif step == :shared_ownership_housing_costs && !Steps::Logic.landlord_is_not_the_only_joint_owner?(session_data)
+          redirect_to helpers.check_step_path_from_step(:property_entry, assessment_code)
```
MazOneTwoOne, line 18: This comment is in the wrong place.

# we need to check for aggregated_means so we know when to show the ":how_to_aggregate" screen when in a change loop

MazOneTwoOne (dismissed):

app/controllers/change_answers_controller.rb
kmahern: Could we move this logic to steps::helper?

So in steps helper

```
def redisplay_property_entry_page?(session_data, step)
        step == :shared_ownership_housing_costs && !Steps::Logic.landlord_is_not_the_only_joint_owner?(session_data)
  end
```

and then in change_answers_controller

```
elsif Steps::Helper.edisplay_property_entry_page?(session_data, step)
    redirect_to helpers.check_step_path_from_step(:property_entry, assessment_code)
```

MazOneTwoOne: @willc-work this is a good point, could we try the above?

kmahern: I don't think this will work for passported applicants, as they don't get shown the shared_ownership_housing_costs page.

spec/flows/certificated_passporting_spec.rb
```diff
@@ -78,6 +78,16 @@
     end
   end
 
+  it "asks for property details if relevant", :ee_banner, :shared_ownership do
```
willc-work, line 81: All my tests were around a non-passported flow so I have added this one to cover the changes in the passported flow around shared_ownership

MazOneTwoOne (approved):

psweeting1 (approved):
Great work!

PR #1829, opened 2025-04-10
EL-2226: Bump to `moj-frontend` 5.1.0

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2226)

## What changed and why

- If applied, this bumps up `moj-frontend` to 5.1.0. 
  - It does this b  updating the load path to look for stylesheets in the current directory for local import files
  - uses `@forward` & `@use` instead of `@import`

## Guidance to review

- n/a

## Checklist


kmahern (approved):
🎖️ This lgtm but it might be worth another dev with better frontend/scss knowledge than me also taking a look.

app/assets/stylesheets/button-group-custom.scss
```diff
@@ -1,3 +1,5 @@
+@use "govuk-frontend/dist/govuk/all" as *; // This means we can use GOV.UK variables i.e. `$govuk-etc`
```
psweeting1, line 1: Ah nice work on the conversation from imports to `@use`

If you're going to assign a variable, it should probably be more descriptive than an wild card, something like https://github.com/ministryofjustice/moj-flask-govuk-skeleton/blob/main/app/static/src/scss/styles.scss would be a better use of a variable name description.

I don't think you even need to do that though, you can probably do the below if you're not utlising scss variables. 

`@use "govuk-frontend/dist/govuk/all"`

Example here in a PR https://github.com/ministryofjustice/govuk-frontend-express/blob/feature/cloud-platform-deploy/src/scss/main.scss

MazOneTwoOne: @psweeting1 I've tried both ways but I get `Error: Undefined variable.`, if I don't use a wildcard.

PR #1838, opened 2025-04-15
EL-2260: Add watermarks to english and welsh CW forms when its shared ownership

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2260)

## What changed and why

- If applied, this change will add watermarks to english and welsh CW forms pdfs, when shared ownership.

## Guidance to review

- added in `nocov` as there is apparently a scenario we are not checking in tests. However, debugging steps I've aded to `spec/forms/controlled_work_document_selection_spec.rb` &  `spec/services/controlled_work_document_value_mapping_service_spec.rb` confim session is being checked.

```
# For debugging purposes
   it "triggers shared ownership logic" do
     puts "owns_property_shared_ownership? => #{Steps::Logic.owns_property_shared_ownership?(session_data)}"
     puts "property_owned => #{session_data['property_owned']}"
     puts "skip_capital_questions? => #{Steps::Logic.skip_capital_questions?(session_data)}"
   end
```

## Checklist


spec/services/controlled_work_document_value_mapping_service_spec.rb
```diff
@@ -406,6 +406,53 @@
     end
   end
 
+  context "with main property being shared ownership" do
+    let(:session_data) do
+      FactoryBot.build(
```
willc-work, line 411: you dont need what FactoryBot is building here as you build the specific api_response below, so you can just o this:
    let(:session_data) do
      FactoryBot.build(
        :minimal_complete_session,
        api_response: FactoryBot.build(:api_result,
                                       main_home: FactoryBot.build(:shared_main_property_api_result, value: 150_000)),
      )

spec/services/controlled_work_document_value_mapping_service_spec.rb
```diff
@@ -406,6 +406,53 @@
...
+                                       main_home: FactoryBot.build(:shared_main_property_api_result, value: 150_000)),
+      )
+    end
+
+    it "can successfully populate CW1 form fields (template with CCQ header)" do
+      mappings = YAML.load_file(Rails.root.join("app/lib/controlled_work_mappings/cw1.yml")).map(&:with_indifferent_access)
+      result = described_class.call(session_data, mappings)
+      representative_sample = {
+        "Means test required" => "Yes_2", # Means test required
+        "Client in receipt of asylum support" => "No", # Asylum supported not given
```
willc-work, line 438: as this is testing shared_ownership should we have a line in the representative_sample that looks specifically at the shared//-ownership box?

willc-work (dismissed):
Couple of changes requested around the testing of this PR. On the nocov issue I agree thatt he line is being hit and `pdftk` is being called and is returning a successful response during testing but it stubbornly refuses to consider that line as being hit, so I agree with leaving that nocov and comment in the PR.

spec/forms/controlled_work_document_selection_spec.rb
```diff
@@ -107,4 +107,48 @@
       expect(page).not_to have_content "CIV Means 7 - family mediation"
     end
   end
+
+  context "when the client's main home is shared ownership" do
```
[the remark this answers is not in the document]
MazOneTwoOne: @willc-work Happy to remove this test. The intention behind it was to prove that we could still download a CW from based on shared ownership session. But if removing session does nothing, then it's not doing what was intended.

spec/services/controlled_work_document_value_mapping_service_spec.rb
willc-work: in that case the test is wrong. I can do this:
```
    let(:session_data) do
      FactoryBot.build(
        :minimal_complete_session,
        api_response: FactoryBot.build(:api_result,
                                       main_home: FactoryBot.build(:shared_main_property_api_result, value: 150_000)),
      )
    end
```
And all the tests on that page pass. So if you are testing for `shared_ownership` there is no expectation on that page that looks for it. 

I think that your `session_data` set up would be required if in this:
```
       representative_sample = {
        "Means test required" => "Yes_2", # Means test required
        "Client in receipt of asylum support" => "No", # Asylum supported not given
        "Has partner whose means are to be agrgregated" => "No_2", # No partner
      }
```
you had a line like:
`"Property Ownership is shared with Housing Association" => "Yes"`

So that we were specifically checking for shared ownership, otherwise the test is essentially saying:
`check that this page isn't broken when shared ownership is true. But if you remove the shared_ownership session data it still passes.

I guess I am not seeing the value of this test in its current form.

willc-work (approved):
Happy with the rubocop disable.

PR #1841, opened 2025-04-15
EL-2259: Remove `back` link on 'cannot_use_service` page

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2259)

## What changed and why

- Remove `back` link on 'cannot_use_service` page
- Also remove redundant view, as we now show 'cannot_use_service` page via it's own controller view 

## Guidance to review

I think this is what was happening:

- When a user changes their answer we work with a temporary copy of the session, stored in a section of the session called `pending`. This then becomes the "gospel" version when the user completes the flow of question steps.
- However, we don't don't use `app/controllers/flow/handler.rb` to send users to `cannot-use-service`, we redirect them to that page instead.
- In this bug scenario, because the users presses back button from `cannot-use-service`, CCQ doesn't know where to send them next, so sends them back to check_answers page and invokes this line:
```
elsif Steps::Helper.last_step_in_group?(session_data, step)
   save_and_redirect_to_check_answers
```
- Technically the `pending` session is then saved and becomes "gospel". You can test this out by going back through to `property-ownership` page and seeing the pages you didn't get shown.

Our options were:

1. Very complicated option: Add some logic, for when in the change answers flow and coming from the cannot-use-service , to show rest of flow from if they select yes on property_landlord page
2. Very simple option: Remove back link on cannot-use-service page, user can use  browser button and flow works and they can also start a new check. - we went with this one

## Checklist


willc-work:
is this html file no longer required: `app/views/question_flow/cannot_use_service.html.slim` ?
I am not sure why it is being deleted as I can see the back link removal on the correct page. If this is a page you found that needed to be removed can you details that in the PR.

MazOneTwoOne:
> is this html file no longer required: `app/views/question_flow/cannot_use_service.html.slim` ? I am not sure why it is being deleted as I can see the back link removal on the correct page. If this is a page you found that needed to be removed can you details that in the PR.

Yes that's correct. It was a remnant from when we used `FormsController` to render the  `cannot_use_service` page. We now use a dedicated controller to make that form. Have documented in the PR notes.

willc-work (approved):

PR #1856, opened 2025-04-29
EL-2216: Resolve failing tests, in local test suite

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2216)

## What changed and why

- If applied, this will resolve failing local tests, by using a gem to synchronize Capybara commands with client-side JavaScript.

## Guidance to review

- I did find this makes the local test suite run slower, but it might have always been like this? 

## Checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,6 +95,7 @@ end
 group :test do
   # Use system testing [https://guides.rubyonrails.org/testing.html#system-testing]
   gem "capybara"
+  gem "capybara-lockstep"
   gem "faker"
   gem "rspec"
   gem "selenium-webdriver"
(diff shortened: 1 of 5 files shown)


kmahern:
@MazOneTwoOne probably missing something, but not clear why we are making changes to views in order to make tests pass?

MazOneTwoOne:
> @MazOneTwoOne probably missing something, but not clear why we are making changes to views in order to make tests pass (locally)?

@kmahern this is [part of the initialisation](https://github.com/makandra/capybara-lockstep?tab=readme-ov-file#including-the-javascript-snippet-required) with `capybara-lockstep`. 

> capybara-lockstep requires a JavaScript snippet to be embedded by the application under test. If that snippet is missing on a screen, capybara-lockstep will not be able to synchronize with the browser. In that case the test will continue without synchronization.

PR #1895, opened 2025-06-20
El-1401: Key totals boxes on results page

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1401)

## What changed and why

- If applied, this updates the totals in the summary totals on the results page

## Checklist

### DIFF
--- a/app/models/calculation_result.rb
+++ b/app/models/calculation_result.rb
@@ -66,6 +66,10 @@ class CalculationResult
     monetise(api_response.raw_total_calculated_gross_income)
   end
 
+  def total_calculated_without_zeros_gross_income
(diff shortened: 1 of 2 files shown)


kmahern (commented):
Looks good. You could consider if you wanted adding tests to calculation_result_spec for the new methods you have added if you wanted (e.g. total_calculated_without_zeros_gross_income, total_calculated_without_zeros_disposable_income, total_calculated_without_zeros_capital) but we don't really have that pattern for the current existing methods so I don't think it is essential.

Also, I think the whole monetise_without_zeros function can be marked private (and I like Imtiaz's suggestion of extracting the precision calculation into a separate method 🙂  )

PR #1948, opened 2025-08-11
EL-2480: experiment - is a simple gem replacement sufficient?

sryalls:
This is a test branch for investigating the Rails 8 upgrade with redis gem

In this branch the redis-rails gem has been swapped out for the redis gem.


MazOneTwoOne:
Closing this as we will be going with Solid cache approach

PR #1461, opened 2024-07-22
EL-1632 delete portal users after 1 week

starswan:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1632)

## What changed and why

Delete old portal usernames for data protection purposes

## Checklist

Before you ask people to review this PR:

- Tests and rubocop should be passing
- Branch is generally up to date with main Github - definitely no conflicts
- No unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- PR description says *what* changed and *why*, with a link to the JIRA story.
- Diff has been checked for unexpected changes being included.
- Commit messages say why the change was made.

PR #1438, opened 2024-07-04
EL-1606 Logout page and signed in banner for LAA portal integration

starswan:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-1606)

## What changed and why

Simple portal integration - just IdP (portal) initiated for now

## Guidance to review

I have left the SdP (service) initiated login code from the original spike, but I am in 2 minds about it.

## Checklist

Before you ask people to review this PR:

- Tests and rubocop should be passing
- Branch is generally up to date with main Github - definitely no conflicts
- No unnecessary whitespace changes. These make diffs harder to read and conflicts more likely.
- PR description says *what* changed and *why*, with a link to the JIRA story.
- Diff has been checked for unexpected changes being included.
- Commit messages say why the change was made.

PR #1966, opened 2025-08-26
EL-2567 / EL-2569 / EL-2568: Remove portal integration

MazOneTwoOne:
[Jira ticket](https://dsdmoj.atlassian.net/browse/EL-2567)

## What changed and why

If applied, this,
- removes the Portal integration from CCQ
- removes `Providers` table with a DB migration, 
- updates what is saved into `CompletedUserJourney` table, 
- amends the privacy policy 
- updates tests and factories

## Guidance to review

This also adds in compatibility redirects from `/auth/*` → `/admins/auth/*`, preserving query params, as without it the 302 redirect was going to the wrong URI. 

This reverses the changes made in these PRs:
- [EL-1606](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1438/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7)
- [EL-1614](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1450/files#diff-0495f118c39af32429bab9896e64399b12e1b1a537431425ec4abe7677835b12)
- [EL-1632](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/1461/files#diff-67abde722d5daddd65348c082470a30d420d0604e697641b0ac090fa38f0151d)

## Checklist

[EL-1606]: https://dsdmoj.atlassian.net/browse/EL-1606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EL-1614]: https://dsdmoj.atlassian.net/browse/EL-1614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EL-1632]: https://dsdmoj.atlassian.net/browse/EL-1632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -13,7 +13,7 @@ html dir="ltr" lang="en-GB" class="govuk-template govuk-template--rebranded"
 
     = render "layouts/cookie_banner"
 
-    = govuk_header(homepage_url: "https://www.gov.uk")
+    = govuk_header(homepage_url: t("service.homepage_url"))
     = govuk_service_navigation(service_name: t("service.name"), service_url: root_path)
 
     .govuk-width-container
--- a/Gemfile
+++ b/Gemfile
@@ -55,7 +55,6 @@ gem "factory_bot_rails"
 gem "devise"
 gem "omniauth-google-oauth2"
 gem "omniauth-rails_csrf_protection", ">= 1.0.2"
-gem "omniauth-saml", "~> 2.2.4"
 
 gem "data_migrate"
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -308,9 +308,6 @@ GEM
     omniauth-rails_csrf_protection (1.0.2)
       actionpack (>= 4.2)
       omniauth (~> 2.0)
-    omniauth-saml (2.2.4)
-      omniauth (~> 2.1)
-      ruby-saml (~> 1.18)
     orm_adapter (0.5.0)
     ostruct (0.6.1)
     pagy (9.3.4)
@@ -500,9 +497,6 @@ GEM
       lint_roller (~> 1.1)
       rubocop (~> 1.72, >= 1.72.1)
     ruby-progressbar (1.13.0)
-    ruby-saml (1.18.1)
-      nokogiri (>= 1.13.10)
-      rexml
     ruby2_keywords (0.0.5)
     rubyzip (3.0.1)
     safe_shell (1.1.0)
@@ -626,7 +620,6 @@ DEPENDENCIES
   jsbundling-rails
   omniauth-google-oauth2
   omniauth-rails_csrf_protection (>= 1.0.2)
-  omniauth-saml (~> 2.2.4)
   parallel_tests
   pdf-forms
   pg (~> 1.6)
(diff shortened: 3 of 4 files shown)


app/views/layouts/application.html.slim
```diff
@@ -13,7 +13,7 @@ html dir="ltr" lang="en-GB" class="govuk-template govuk-template--rebranded"
 
     = render "layouts/cookie_banner"
 
-    = application_header
+    = govuk_header(homepage_url: "https://www.gov.uk")
```
sryalls, line 16: Can we use a helper for this rather than hardcode the URL? Or at least put it in a config?

sryalls:
Should the reference to 

`gem "omniauth-saml", "~> 2.2.4"`

in the gemfile also be removed?

app/controllers/application_controller.rb
```diff
@@ -5,13 +5,10 @@ class ApplicationController < ActionController::Base
   before_action :force_setting_of_session_cookie,
                 :specify_feedback_widget,
                 :specify_freetext_feedback_page_name,
+                :authenticate,
```
sryalls, line 8: Looking at the authenticate function it seems like we might be opening up a security vulnerability here? 

If the feature flag gets turned off (line 82) the whole auth process will get bypassed? I'd suggest either removing the reliance on the feature flag or changing the feature flag logic to require an active feature flag to skip the authentication:

```
def authenticate
  # Default to requiring authentication unless explicitly disabled
  unless FeatureFlags.enabled?(:skip_authentication, without_session_data: true)
    if cookies.signed[BASIC_AUTHENTICATION_COOKIE]
      session["user_return_to"] = nil
    else
      session["user_return_to"] = request.path
      redirect_to new_basic_authentication_session_path
    end
  end
end
```

app/views/layouts/application.html.slim
MazOneTwoOne: Have updated and put into `en.yml`

db/migrate/20250827102123_remove_providers.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveProviders < ActiveRecord::Migration[7.2]
```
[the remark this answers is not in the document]
MazOneTwoOne: Have asked our BA (Rob) and her is happy to removed the DB. We've also taken a copy of existing data of which there wasn't much.  I'll keep the way it is.

app/controllers/application_controller.rb
MazOneTwoOne: This authentication is separate from the rails admin authentication. 

The intention, [when first implementing this ](https://github.com/ministryofjustice/laa-check-client-qualifies/pull/876), was a basic authentication for Staging & UAT links, so the public couldn't accidentally interact with those URLs

sryalls (approved):

PR #2158, opened 2026-04-24
feat: ✨ Adding two different operating modes: embedded and standalone (MEM-683)

Kerl1310:
[Jira ticket](https://dsdmoj.atlassian.net/browse/MEM-683)

## What changed and why

- Introduce `CCQ_MODE=standalone|embedded`
- Add `ModeConfig` to expose configuration predicates based on the current `CCQ_MODE` value
- Use `ModeConfig` to:
  - gate initialisation of various database dependencies
  - use redis cache adapter instead of solid cache
- Add test routing for embedded mode specific specs
- Update existing specs to exercise embedded mode branches to maintain 100% coverage

## Guidance to review

We have tested this PR in standalone mode and have tested the server runs in embedded mode, we've also tested that we can run embedded mode in Docker Compose behind an nginx proxy. One thing this PR doesn't solve is the fact that CCQ's assets are served by CCQ itself and as such, asset requests need to be proxied to CCQ as well. 

This is problematic because CCQ serves some assets from `/assets` and others from `/`. Neither of these are specific enough for CCQ to be consistently routed its own asset requests on a shared domain by a reverse proxy. This issue is fixed by #2167.

## Checklist


richardcane:
From chatting, need to add more config variables that do mode conditional output in the mode config itself rather than having embedded flags distributed throughout codebase. I will pick that up while you're out next week.

richardcane:
> From chatting, need to add more config variables that do mode conditional output in the mode config itself rather than having embedded flags distributed throughout codebase. I will pick that up while you're out next week.

This ended up just being the `cache_store` for now.

richardcane:
Well that was harder than I expected. There are a number of models which are evaluated on boot by zeitwerk, which scans the app directory for files and passes them to Ruby to be evaluated. A number of these models contain macros - methods attached to classes like `ApplicationRecord` by gems like Devise. Those are never attached if Devise is prevented from initialising, which causes errors.

On local, this is obscured by two things:
- Zeitwerk only eager-loads in the CI context (config/environments/test.rb:17), which means tests only load the exact files they need on local.
- If you ever run the standalone tests, bootsnap caches the macros, which makes them accessible to the embedded tests (though given the lazy loading, I don't think this was really what was happening)

I've added a make task which disables caching, pretends to be CI and runs `rake zeitwerk:check` so that we can debug this more easily if it comes up again, though I think I got em all with the zeitwerk initializer (which is telling zeitwerk to ignore the models with standalone gem macros if running in embedded mode)

racheld-madetech (approved):

PR #2170, opened 2026-05-06
Optimise dockerfile

richardcane:
- Add docker-compose file for testing built images locally, both for embedded and standalone CCQ modes.
- Optimise docker image size. `main` currently builds at 4.4gb, this branch builds at 1.99gb, reductions largely from layer optimisation.
- One fairly important thing which is easy to overlook is Puppeteer: 
  - It's used to generate PDFs. It installs its own version of Chrome. It does not provide ARM binaries, which causes rosetta errors if you try to run the image on MacOS and then generate a PDF. The image must be built using `platform=linux/amd64`.
  - Puppeteer by default installs both Chrome for Testing and headless Chrome. We don't need both, so we tell it not to download anything at first. In the runner stage, we install Chrome manually.


docker-compose.yml
```diff
@@ -0,0 +1,102 @@
...
+    ports:
+      - 8080:8080
+    volumes:
+      - ./docker/compose/nginx/default.conf.template:/etc/nginx/templates/default.conf.template:ro
+      - ./docker/compose/nginx/proxy.conf:/etc/nginx/proxy.conf:ro
+      - ./docker/compose/nginx/maps.conf:/etc/nginx/maps.conf:ro
+    depends_on:
+      - ccq-embedded
+    environment:
+      NGINX_UPSTREAM_CCQ: ccq-embedded:3000
```
Kerl1310, line 15: Might be worth re-ordering this to be below the ccq-embedded block that starts on L17 so that it's reads a bit better with regard to the `depends_on`

Kerl1310 (approved):

docker-compose.yml
richardcane: Possibly, I've put it at the top for RCW as well because it felt like it fit my mental model of the architecture better, but I might be tunnel visioning.

app/controllers/results_controller.rb
```diff
@@ -34,7 +34,7 @@ def download
       layout: "download_application",
     })
 
-    PdfService.with_pdf_data_from_html_string(html, request.base_url) do |pdf_data|
+    PdfService.with_pdf_data_from_html_string(html, "http://localhost:3000") do |pdf_data|
```
richardcane: This needs reverting

app/controllers/results_controller.rb
```diff
@@ -62,4 +62,11 @@ def specify_feedback_widget
   def specify_satisfaction_feedback_page_name
     @satisfaction_feedback_page_name = page_name
   end
+
+  # Grover renders HTML inside Chromium. In Docker (local), Chromium needs an
+  # internal app URL to resolve relative assets. Normal deployments can use
+  # the public request base URL.
+  def pdf_display_url
+    ENV["PDF_DISPLAY_URL"].presence || request.base_url
+  end
```
richardcane, line 71: Added this, rather than hardcoding the display_url to be a localhost URL which could have unintended consequences in deployed envs

PR #2192, opened 2026-05-27
feat: ✨ Adding Host Service client (MEM-727)

Kerl1310:
[Jira ticket](https://dsdmoj.atlassian.net/browse/MEM-727)

## What changed and why

test: 🧪 Adding tests to cover the new Host Service client
feat: ✨ Adding Host Service client

## Checklist


app/services/host_service_client.rb
```diff
@@ -0,0 +1,33 @@
+class HostServiceClient
+  class ConnectionError < StandardError; end
+
+  CONNECT_TIMEOUT = 2
+  READ_TIMEOUT = 5
+
+  def load(resource_id:, cookies:)
+    post("/api/private/load", { resource_id: }, cookies:)
```
richardcane: Can we make these configurable?

app/services/host_service_client.rb
```diff
@@ -0,0 +1,33 @@
...
+private
+
+  def post(path, body, cookies:)
+    connection(cookies:).post(path, body)
+  rescue Faraday::ConnectionFailed, Faraday::TimeoutError => e
+    raise ConnectionError, e.message
+  end
+
+  def connection(cookies:)
+    Faraday.new(url: ENV.fetch("HOST_SERVICE_URL")) do |faraday|
```
richardcane, line 24: This will need to pass the value of the RCW session cookie - need to figure that bit out as well.

richardcane (approved):

app/services/host_service_client.rb
Kerl1310: Can update to be something like `post(ENV.fetch("HOST_SERVICE_LOAD_ENDPOINT"), { resource_id: }, cookies:)`?

app/services/host_service_client.rb
Kerl1310: Copilot assures me this should pass the cookie as is via the cookies argument which is put into the headers below

PR #2238, opened 2026-07-15
Configurable layouts for embedded mode

richardcane:
- Adds configuration that allows the service to be booted with a different layout for embedded mode (only).
- Adds a layout for the RCW service.


app/lib/mode_config.rb
```diff
@@ -41,6 +43,20 @@ def self.cache_store
     ]
   end
 
+  def self.embedded_layout
+    raw = ENV.fetch("CCQ_EMBEDDED_LAYOUT", DEFAULT_EMBEDDED_LAYOUT).to_s.strip
```
Kerl1310, line 47: My only question mark in this is whether we need this to be configurable via env var, or whether we could just flip between the two using `CCQ_MODE`. Are we anticipating additional layouts?

richardcane: Yeah, other teams are wanting to use embedded mode as well at some point.

jamesgrant-madetech (approved):

